PR #29, opened 2025-12-03
fix: preserve function signature in diagnostics wrapper for correct schema generation

Milofax:
## Summary
- Fixed `wrap_tool` in `diagnostics.py` to preserve the original function's signature using `functools.wraps` and explicit `__signature__` copying
- Fixed type inference in `main.py` to correctly identify generic types like `Dict[str, Any]` as `"object"` instead of `"string"`
- Regenerated `tools_manifest.json` with correct schemas

## Problem
The diagnostics wrapper was replacing the original function signature with `(*args, **kwargs)`, causing FastMCP to generate incorrect JSON schemas. Tools with `Dict[str, Any]` parameters (like `create_simple_port_forward`) were showing:

```json
{
  "inputSchema": {
    "properties": {
      "args": { "type": "string" },
      "kwargs": { "type": "string" }
    }
  }
}
```

Instead of the correct schema:

```json
{
  "inputSchema": {
    "properties": {
      "rule": { "type": "object" },
      "confirm": { "type": "boolean" }
    }
  }
}
```

## Solution
1. Use `functools.wraps(func)` decorator on the wrapper function
2. Explicitly copy `__signature__` from the original function so FastMCP can correctly introspect parameters
3. Use `typing.get_origin()` to detect generic types like `Dict[str, Any]`

## Test plan
- [x] Tested with `create_simple_port_forward` - now correctly accepts `{"rule": {...}, "confirm": false}` instead of erroring with `_wrapperArguments` validation errors
- [x] Verified schema generation via MCP `tools/list` endpoint
- [x] Regenerated manifest with correct schemas

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -82,17 +82,21 @@ def permissioned_tool(*d_args, **d_kwargs):  # acts like @server.tool
                     param_type = "string"  # default
                     if param.annotation != inspect.Parameter.empty:
                         ann = param.annotation
-                        # Basic type mapping
-                        if ann in (int, "int"):
+                        # Handle generic types like Dict[str, Any], List[str]
+                        from typing import get_origin
+
+                        origin = get_origin(ann)
+                        # Basic type mapping (check origin first for generics)
+                        if origin is dict or ann in (dict, "dict"):
+                            param_type = "object"
+                        elif origin is list or ann in (list, "list"):
+                            param_type = "array"
+                        elif ann in (int, "int"):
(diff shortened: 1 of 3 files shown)


sirkirby (changes requested):
Change looks good, thanks for finding and submitting, i missed this in my testing. Please increment the patch version in the pyproject.toml, and I'll get this approved. Please also pull in the latest code, looks like you are a couple of commits behind

PR #43, opened 2025-12-17
fix: suppress logging tracebacks when HTTP SSE runs with stdio

b3nw:
When HTTP SSE runs alongside stdio transport, the stdio transport takes over stdout for MCP JSON-RPC communication, causing any logger with a StreamHandler to fail with 'I/O operation on closed file'.

This fix patches `logging.StreamHandler.emit` at module load time to check if the stream is closed before attempting to write, preventing the '--- Logging error ---' tracebacks from appearing in container logs.

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -335,9 +335,13 @@ async def main_async():
         async def run_http():
             try:
                 logger.info(f"Starting FastMCP HTTP SSE server on {host}:{port} ...")
-                # MCP SDK >= 1.10 (pinned to 1.13.1): configure host/port via settings
                 server.settings.host = host
                 server.settings.port = port
+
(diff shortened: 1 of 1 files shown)


b3nw:
sorry, I pushed this too quickly, when I was testing it *did not* resolve the issue. you will want to revert this.

The disabled flag on a logger only affects whether logger.isEnabledFor() returns True. 

However:
The handler's emit() method is called directly
The handler still has a StreamHandler pointing to stdout
When stdout is closed, stream.write() fails


Why Our Original Patch Worked
Our StreamHandler.emit patch worked because:

It's applied at module load time - before any imports
It intercepts all StreamHandler emit calls globally
It checks stream.closed before attempting to write

PR #61, opened 2026-02-04
feat(clients): add unifi_lookup_by_ip tool for efficient IP-to-hostname lookup

alejandroed:
## Summary

Adds a new tool `unifi_lookup_by_ip` that efficiently resolves an IP address to its hostname/name/MAC without returning the full client object.

**Use case:** When integrating with LLM tools (like Claude Code MCP), querying full client lists consumes excessive tokens. This tool returns only essential fields, reducing token usage significantly.

## Changes

- **`src/managers/client_manager.py`**: Added `get_client_by_ip(ip_address)` method
  - Searches online clients first, falls back to all clients (avoids stale IP matches)
  - Uses `Client` model `.ip` property (not `.raw`)
  - Basic IPv4 format validation
- **`src/tools/clients.py`**: Added `unifi_lookup_by_ip` tool with `site` in response
- **`tests/unit/test_client_lookup.py`**: 5 unit tests covering found/not-found/invalid/online-first
- **`src/tools_manifest.json`**: Regenerated (84 tools)

**No changes to existing tools** — `block_client`, `unblock_client`, `force_reconnect_client`, and `get_client_details` remain untouched.

## Example Response

\`\`\`json
{
  "success": true,
  "site": "default",
  "ip": "192.168.1.100",
  "hostname": "PC-OFFICE",
  "name": "John's Computer",
  "mac": "aa:bb:cc:dd:ee:ff"
}
\`\`\`

## Review feedback addressed

All 6 items from the initial review:
1. Regenerated tool manifest
2. Added unit tests (5 tests)
3. Uses Client model \`.ip\` property
4. Includes \`site\` in response
5. Searches online clients first, falls back to all
6. Basic IP format validation

## Test plan

- [x] 152/152 unit tests pass
- [x] Manifest sync tests pass
- [x] No changes to existing tool signatures

### DIFF
--- a/README.md
+++ b/README.md
@@ -525,6 +525,7 @@ The server merges settings from **environment variables**, an optional `.env` fi
 | `UNIFI_ENABLED_CATEGORIES` | Comma-separated list of tool categories to load (eager mode). See table below |
 | `UNIFI_ENABLED_TOOLS` | Comma-separated list of specific tool names to register (eager mode) |
 | `UNIFI_MCP_ALLOWED_HOSTS` | Comma-separated list of allowed hostnames for reverse proxy support. Required when running behind Nginx/Cloudflare/etc. Default `localhost,127.0.0.1` |
+| `UNIFI_MCP_ENABLE_DNS_REBINDING_PROTECTION` | Enable/disable DNS rebinding protection. Set to `false` for Kubernetes/proxy deployments where `UNIFI_MCP_ALLOWED_HOSTS` is insufficient. Default `true` |
 
 ### Tool Categories (for UNIFI_ENABLED_CATEGORIES)
 
@@ -746,6 +747,8 @@ These tools will give any LLM or agent configured to use them full access to you
 
 The server includes a comprehensive permission system with **safe defaults**:
 
+> **Permissions control tool visibility.** Tools with disabled permissions are **not registered** with the MCP server and will not appear in your client's tool list. If you're missing expected tools, check that the relevant permissions are enabled. All tools remain discoverable via `unifi_tool_index` regardless of permission settings — but disabled tools cannot be called. See [docs/permissions.md](docs/permissions.md) for full details.
+
 **Disabled by Default (High-Risk):**
 - Network creation/modification (`unifi_create_network`, `unifi_update_network`)
 - Wireless configuration (`unifi_create_wlan`, `unifi_update_wlan`)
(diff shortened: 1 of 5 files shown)


sirkirby (changes requested):
Thanks for the thorough follow-up — all 6 of the original review items are addressed well. The tests are solid and the `get_client_by_ip` improvements (online-first search, IP validation, Client model properties) are exactly what we asked for.

However, the second commit introduces several changes beyond what was requested that I'd like us to handle separately.

### Concerns

**1. Breaking parameter renames**

`block_client`, `unblock_client`, and `force_reconnect_client` have had their `mac_address` parameter renamed to `client`. Any existing automation calling these tools with `mac_address=` will break. This is a semver-breaking change to the public API.

**2. Partial name matching on mutating operations is risky**

`find_client` uses substring matching (`query_lower in name`), which means `block_client(client="PC")` would block the *first* client whose name or hostname contains "PC". For destructive operations this is dangerous — an ambiguous match should not silently pick the first result.

**3. Premature deprecation**

`unifi_lookup_by_ip` doesn't exist in `main` yet, so marking it `[DEPRECATED]` in the same PR that introduces it sends a confusing signal. Similarly, deprecating `unifi_get_client_details` (which serves a different purpose — full raw output for debugging) should be a separate decision.

**4. Inconsistent refactor**

`rename_client`, `authorize_guest`, `unauthorize_guest`, and `set_client_ip_settings` still use `mac_address`. A partial migration creates an inconsistent API surface.

### Suggested path forward

Could you split this into two PRs?

**This PR (merge now):** Revert to just the originally scoped changes:
- `get_client_by_ip()` manager method (with your online-first search, IP validation, Client model properties — all great)
- `unifi_lookup_by_ip` tool (with `site` in response, no deprecation label)
- The unit tests for `get_client_by_ip`
- Regenerated manifest
- No changes to existing tools (`block_client`, `unblock_client`, `force_reconnect_client`, `get_client_details`)

**Follow-up PR (separate discussion):** The broader `find_client` refactor:
- `find_client()` manager method + `unifi_find_client` tool
- Parameter renames on existing tools
- Deprecation strategy
- This deserves its own design discussion around: exact vs partial matching for mutating ops, migration strategy for the parameter rename, and consistency across all client tools

The `find_client` concept is genuinely useful — I just want to get it right rather than rush it in alongside the IP lookup feature. Happy to discuss the design in a separate issue!

Issue #62, opened 2026-02-06
Unknown tool errors when querying with Claude Desktop


mojo333:
I have created a local Unifi-Network MCP server as per Option 1 in https://github.com/sirkirby/unifi-network-mcp?tab=readme-ov-file#using-with-claude-desktop-1 where Claude invokes the local package.

Login seems to work fine, but I get errors such as the following when running any query:

```
ERROR:unifi-network-mcp:Error executing tool 'list_networkconf': Unknown tool: list_networkconf
Traceback (most recent call last):
  File "C:\Users\ibret\AppData\Local\uv\cache\archive-v0\2MCgb3HftHS2VjGmUD9M-\Lib\site-packages\src\utils\meta_tools.py", line 90, in execute_handler
    result = await server.call_tool(tool, arguments)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\ibret\AppData\Local\uv\cache\archive-v0\2MCgb3HftHS2VjGmUD9M-\Lib\site-packages\mcp\server\fastmcp\server.py", line 346, in call_tool
    return await self._tool_manager.call_tool(name, arguments, context=context, convert_result=True)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\ibret\AppData\Local\uv\cache\archive-v0\2MCgb3HftHS2VjGmUD9M-\Lib\site-packages\mcp\server\fastmcp\tools\tool_manager.py", line 91, in call_tool
    raise ToolError(f"Unknown tool: {name}")
```
These are the log entries just above this (controller URL anonymised). 
```
INFO:unifi-network-mcp:Initializing global Unifi connection from main_async...
INFO:unifi-network-mcp:Attempting to connect to Unifi controller at unifi.xyz.net...
INFO:unifi-network-mcp:Pre-login auto-detected controller type: UniFi OS (proxy)
WARNING:unifi-network-mcp:Auto-detection failed - both endpoints unsuccessful
WARNING:unifi-network-mcp:Auto-detection failed - both endpoints unsuccessful
INFO:unifi-network-mcp:Successfully connected to Unifi controller at unifi.xyz.net for site 'default'
INFO:unifi-network-mcp:Global Unifi connection initialized successfully from main_async.
[...]

PR #64, opened 2026-02-06
fix: enable on-demand tool loading in meta_only mode

### LINKED ISSUE #63: fix: enable on-demand tool loading in meta_only mode
## Context

Reported in #62 — when running in `meta_only` mode, `unifi_execute` and `unifi_batch` fail with `Unknown tool` errors because the lazy loading interceptor is not wired up.

In `lazy` mode, `setup_lazy_loading()` monkey-patches `server.call_tool` to auto-import tool modules on first use. In `meta_only` mode this interceptor was missing, so `server.call_tool()` only knows about the 4 registered meta-tools and throws for anything else.

## Fix

Call `setup_lazy_loading()` in `meta_only` mode so that `unifi_execute`/`unifi_batch` can load tools on demand, the same way `lazy` mode does. The distinction remains: `lazy` also registers `unifi_load_tools` for clients that support `tool_list_changed` notifications, while `meta_only` does not.

## Parent issue

Closes #62

sirkirby:
## Summary

- Wire up `setup_lazy_loading()` in `meta_only` mode so `unifi_execute` and `unifi_batch` can load tools on demand
- Previously, `meta_only` mode only registered 4 meta-tools and had no lazy interceptor, causing all `unifi_execute` calls to fail with `Unknown tool`
- The distinction between `meta_only` and `lazy` is preserved: `lazy` additionally registers `unifi_load_tools` for clients supporting `tool_list_changed` notifications

Fixes #63
Relates to #62

## Test plan

- [x] All 153 unit tests pass
- [x] Ruff linter clean
- [x] Manual: start server with `UNIFI_TOOL_REGISTRATION_MODE=meta_only`, call `unifi_execute` with a tool name — should auto-load and succeed
- [x] Manual: verify `lazy` and `eager` modes still work as before

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -253,6 +253,13 @@ async def main_async():
         logger.info("   Meta-tools: unifi_tool_index, unifi_execute, unifi_batch, unifi_batch_status")
         logger.info("   Use unifi_execute to run any tool discovered via unifi_tool_index")
         logger.info("   To load all tools directly: set UNIFI_TOOL_REGISTRATION_MODE=eager")
+
+        # Setup lazy loading interceptor so unifi_execute/unifi_batch can load tools on demand
+        setup_lazy_loading(server, _original_tool_decorator)
+
+        from src.utils.lazy_tool_loader import TOOL_MODULE_MAP
+
+        logger.info(f"   On-demand loader ready - {len(TOOL_MODULE_MAP)} tools available via unifi_execute")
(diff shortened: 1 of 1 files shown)


sirkirby (commented):
manually verified with a local smoke test that meta-only behaves as expected and lazy load automaticlaly preventing the execute tool from hallucinating tool names

PR #76, opened 2026-03-13
fix: handle list responses in SystemManager stat endpoints

boinger:
## Summary

`ConnectionManager.request()` unwraps the UniFi API envelope `{"meta": {...}, "data": [...]}` and returns the inner `data` array. Three `SystemManager` methods assumed dict responses from `/stat/` endpoints but received lists, silently falling through to empty dicts:

- `get_system_info()` — `/stat/sysinfo`
- `get_controller_status()` — `/stat/status`
- `check_firmware_updates()` — `/stat/fwupdate/latest-version`

Now extracts the first element when the response is a non-empty list, preserving dict fallback for forward compatibility.

Note: other managers in the codebase already handle this correctly (e.g. `get_settings()` checks `isinstance(response, list)`, `get_network_health()` accepts both via `isinstance(response, (list, dict))`). Only these three singleton-expecting methods had the bug.

## Test plan

- [x] `make pre-commit` passes (203 tests, lint clean)
- [x] Called `unifi_get_system_info` via MCP and verified non-empty response with controller version, uptime, etc.


sirkirby (approved):
Looks good. Thanks!

PR #135, opened 2026-04-06
fix(acl): add source_macs/destination_macs params to prevent silent ANY→ANY rule creation

level99:
## Summary

Fixes a security bug where `create_acl_rule` silently dropped MAC addresses
passed as `source_macs`/`destination_macs`, creating ANY→ANY ALLOW rules
instead of MAC-specific rules. Also adds flattened-to-nested field translation
in `update_acl_rule` and clarifies the same class of list-vs-create field
shape mismatch in OON and content filtering tools.

### Bug details

`list_acl_rules` outputs flattened fields (`source_macs`, `destination_macs`),
but `create_acl_rule` expected nested dicts (`traffic_source:
{specific_mac_addresses: [...]}`). When an LLM or caller passed the field
names from the list output, they were silently ignored by the MCP framework
(not in the function signature), and the tool defaulted to empty MAC lists =
match any device. An admin intending to allow one specific MAC got an ALLOW
ANY→ANY rule with no error.

Discovered during live use: an attempt to create a single-MAC ALLOW rule on
a default-deny network silently punched a hole allowing all device-to-device
traffic. The rule was caught and deleted within seconds, but only because the
caller knew to verify the created rule's MAC lists.

### Why OON and content filtering are in this PR

After discovering the ACL bug, we audited all tools for the same pattern
(list output uses flattened field names, create/update expects nested dicts,
LLM naturally uses the list field names and they get silently dropped or
ignored). Found two more instances:

- **OON**: `list_oon_policies` outputs `qos_enabled`/`route_enabled`
  (flattened booleans), but `create_oon_policy` expects `qos`/`route`
  (full nested dicts). Lower severity than ACL — an LLM passing
  `qos_enabled=True` would get an error (not a valid param), not a silent
  misconfiguration. Fix: clarified the create description to explicitly
  warn against using the flattened list fields.

- **Content filtering**: `list_content_filters` outputs `schedule_mode`
  (flattened from nested `schedule.mode`), but the update tool doesn't
  accept schedule changes at all. No risk — read-only summary field.
  Fix: added code comment documenting it as read-only.

Bundled here because they're the same class of bug found in the same audit.
The OON and content filtering changes are description/comment-only — no
behavioral changes outside of ACL.

### Fix

- **`create_acl_rule`**: Added `source_macs` and `destination_macs` as
  top-level convenience params (same field names as `list_acl_rules` output).
  These build the nested `traffic_source`/`traffic_destination` dicts
  internally. Original nested dict params retained for backward compatibility
  but marked as "(Advanced)" — convenience params take precedence when both
  are provided.
- **`update_acl_rule`**: Translates `source_macs`/`destination_macs` keys in
  `rule_data` to nested structure before schema validation, so callers can
  use the same field names from list output.

## Files changed

| File | Change |
|---|---|
| `tools/acl.py` | **Modified** — added `source_macs`/`destination_macs` convenience params on create, flattened-to-nested translation on update, entry logging, updated descriptions |
| `tools/oon.py` | **Modified** — clarified create description re: nested vs flattened fields |
| `tools/content_filtering.py` | **Modified** — added read-only comment on `schedule_mode` |
| `tests/unit/test_acl_tools.py` | **New** — 7 tests covering convenience params, precedence, backward compat, defaults, preview, update translation |
| `tools_manifest.json` | Regenerated (166 tools) |

## Live testing

| Environment | Details |
|---|---|
| **UniFi OS** | 5.0.16 |
| **MCP Client** | Claude Code |
| **MCP Transport** | stdio |
| **Shell** | Git Bash 5.2.37 (MINGW64) |

| Test | Result |
|---|---|
| `create_acl_rule` with `source_macs` + `destination_macs` — MACs persisted in controller response | ✅ Pass |
| Verified `traffic_source.specific_mac_addresses` and `traffic_destination.specific_mac_addresses` populated correctly | ✅ Pass |
| Test rule deleted after verification | ✅ Pass |

## Unit tests

7 new tests in `test_acl_tools.py`:
- `source_macs` builds correct nested dict
- `source_macs` takes precedence over `traffic_source` dict
- Backward compat: `traffic_source` dict still works alone
- Omitting both defaults to ANY (empty list)
- Preview mode shows resolved MACs
- `update_acl_rule` translates `source_macs` in rule_data
- `update_acl_rule` translates `destination_macs` in rule_data

```
Full suite — 578 passed, 0 failed
```

### DIFF
--- a/apps/network/src/unifi_network_mcp/tools/acl.py
+++ b/apps/network/src/unifi_network_mcp/tools/acl.py
@@ -8,7 +8,7 @@ Application with Policy Engine support.
 
 import json
 import logging
-from typing import Annotated, Any, Dict, Optional
+from typing import Annotated, Any, Dict, List, Optional
 
 from mcp.types import ToolAnnotations
 from pydantic import Field
@@ -115,8 +115,8 @@ async def get_acl_rule_details(
 @server.tool(
     name="unifi_create_acl_rule",
     description="Create a new MAC ACL rule for Layer 2 access control within a VLAN. "
-    "Requires confirmation. IMPORTANT: traffic_source and traffic_destination type MUST be 'CLIENT_MAC' "
-    "(not 'ANY'). Use an empty specific_mac_addresses list to match any device.",
+    "Use source_macs/destination_macs to specify MAC addresses (same field names as unifi_list_acl_rules output). "
+    "Empty list = match any device. Requires confirmation.",
     permission_category="acl_rules",
     permission_action="create",
     annotations=ToolAnnotations(readOnlyHint=False, destructiveHint=False, idempotentHint=False, openWorldHint=False),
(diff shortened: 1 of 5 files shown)


sirkirby (approved):
Good to merge. follow-up issue #136  pending feedback for larger fix

Issue #138, opened 2026-04-08
Transport-layer unknown-kwarg rejection middleware


sirkirby:
## Summary

Add a JSON-RPC transport-layer middleware that intercepts incoming `tools/call` requests, diffs the `arguments` dict keys against the tool's declared input schema in `tools_manifest.json`, and rejects unknown keys with a structured error before the request reaches FastMCP's pydantic layer. This is the local tactical mitigation for the silent-drop mechanism tracked in #136 — it catches typos, version skew, and shape drift across all 166+ tools in a single implementation.

## Why this shape

The root cause in #136 is that FastMCP's `ArgModelBase` is configured with pydantic's default `extra="ignore"`, which silently drops unknown kwargs. The upstream fix is blocked on spec-level deliberation (tracked in #136). We don't need to wait — we can intercept at a layer we control.

The cleanest intercept point is **before FastMCP sees the call**, at the JSON-RPC dispatch layer. Implementation options in rough order of preference:

1. **Subclass `FastMCP` / `MCPServer`** with a custom dispatch method that runs kwarg validation against `tools_manifest.json` before delegating to super. Cleanest — composition, not patching.
2. **FastMCP middleware hook** if one exists in the SDK's extension API (needs investigation).
3. **Decorator wrapper** around `@server.tool(...)` that captures the function signature's parameter names and checks incoming calls. Works for top-level kwargs but can't catch the dict-param case.

Option 1 is probably the target. Specifics land in the implementing PR.

## Benefits

- **One implementation covers every tool.** No per-tool maintenance burden.
- **Zero FastMCP patching.** We subclass, not monkey-patch. Safe across SDK upgrades.
- **Uses existing source of truth.** `tools_manifest.json` already enumerates every tool's input schema — no new registry.
- **Self-retiring.** Once upstream lands `extra="forbid"` on `ArgModelBase`, this middleware becomes a no-op and can be removed cleanly.
- **Complementary to #137.** The field symmetry rule eliminates the most common silent-drop case (caller round-tripping list output into create calls with correct field names). This middleware catches everything else: typos, version skew, misaligned calls, and any tool pair not yet migrated.

## Error contract

When the middleware rejects a call, the response should be structured enough for an LLM to self-correct (per SEP-1303's intent):

```json
{
  "error": {
    "code": -32602,
    "message": "Invalid params: unknown arguments {source_mac}",
    "data": {
      "tool": "unifi_create_acl_rule",
[...]

PR #140, opened 2026-04-08
feat(acl): pilot shared field model for ACL domain (#139)

level99:
## Summary

Pilots the shared field model pattern for the ACL domain: defines all ACL
rule fields once in a pydantic model (`AclRule`), derives list/get output
and create/update input from the same source, and retires the per-tool
convenience-param shim from #135 and the JSON Schema from `schemas.py`.

This is the structural fix for the field symmetry rule in #137 — field
drift between list and create/update tools becomes impossible without
editing the model. Once landed, this becomes the anchor file for the new
AGENTS.md golden path and the migration template for other domains.

### What changed

- **New `models/acl.py`** — `AclRule` pydantic model with mutable/read-only
  field metadata, plus three translation helpers (`from_controller`,
  `to_controller_create`, `to_controller_update`) co-located with the model.
- **Refactored all 4 ACL tools** — `list_acl_rules` and `get_acl_rule_details`
  use `from_controller().model_dump()` for consistent output. `create_acl_rule`
  builds an `AclRule` from params then translates to controller shape.
  `update_acl_rule` validates field names against `MUTABLE_FIELDS` and
  translates via `to_controller_update()`.
- **Retired `traffic_source`/`traffic_destination` params** — the nested
  controller-dialect dicts are now an internal translation concern, not
  caller-facing. The #135 convenience-param shim is no longer needed.
- **Retired `ACL_RULE_UPDATE_SCHEMA`** from `schemas.py` and
  `validator_registry.py` — the pydantic model replaces the JSON Schema.
- **Renamed `mac_acl_network_id` → `network_id`** on `create_acl_rule` —
  matches the list output field name.
- **Added `enabled` param** to `create_acl_rule` — caught by the field
  symmetry test (was a mutable field on the model but missing from the
  create signature).
- **Unknown field rejection** on `update_acl_rule` — fields not in
  `MUTABLE_FIELDS` are rejected with a clear error listing allowed fields.
  This is the tool-layer equivalent of `additionalProperties: false`.
- **AGENTS.md** — new golden path section "Add or migrate a domain to
  shared field models" with ACL as the anchor.

### Design decisions (per #139 discussion)

| Question | Decision | Rationale |
|---|---|---|
| Translation location | Co-located with model | One file to update when controller API changes; manager stays dumb, tools stay thin |
| Replace or coexist with JSON Schema | Replace | Pilot proves the pydantic-only path end-to-end; two sources of truth defeats the purpose |
| Read-only field marking | `json_schema_extra={"mutable": False}` | Co-located, introspectable by CI, no external allowlist |
| Retire traffic_source/traffic_destination | Yes | Never the public interface; list always output flat field names |

## Files changed

| File | Change |
|---|---|
| `models/__init__.py` | **New** — empty package init |
| `models/acl.py` | **New** — `AclRule` model, translation helpers, `MUTABLE_FIELDS`/`READ_ONLY_FIELDS`/`UPDATE_FIELD_MAP`/`MAC_TRANSLATED_FIELDS` |
| `tools/acl.py` | **Modified** — all 4 tools refactored to derive I/O from model; `traffic_source`/`traffic_destination` retired; `UniFiValidatorRegistry` removed; `network_id` renamed |
| `schemas.py` | **Modified** — `ACL_RULE_UPDATE_SCHEMA` removed |
| `validator_registry.py` | **Modified** — `acl_rule_update` registration removed |
| `tests/unit/test_acl_tools.py` | **Rewritten** — 19 tests: 6 model translation, 4 create, 6 update, 3 list/symmetry |
| `tools_manifest.json` | Regenerated (166 tools) |
| `AGENTS.md` | **Modified** — new golden path for shared field models |

## Live testing

| Environment | Details |
|---|---|
| **UniFi OS** | 5.0.16 |
| **MCP Client** | Claude Code |
| **MCP Transport** | stdio |
| **Shell** | Git Bash 5.2.37 (MINGW64) |

| Test | Result |
|---|---|
| `list_acl_rules` — returns model shape (flat `source_macs`, `network_id`) | ✅ Pass |
| `create_acl_rule` — `network_id` + `source_macs` + `destination_macs` persisted, response in model shape | ✅ Pass |
| `update_acl_rule` — `source_macs` + `name` in model field names translated correctly | ✅ Pass |
| `update_acl_rule` — `traffic_source` rejected as unknown with clear error listing allowed fields | ✅ Pass |
| `get_acl_rule_details` — returns model shape, confirms update persisted | ✅ Pass |
| `delete_acl_rule` — cleanup successful | ✅ Pass |

## Zero-context LLM validation

To verify the tools are self-documenting for LLM callers, we ran a
separate agent with zero prior knowledge of the codebase, schemas, or
field names. It was given only a goal ("list ACL rules, create one,
update it, verify it, delete it") and access to the MCP tools.

**Results:** The agent completed the full CRUD round-trip successfully.
All data-level field names (`source_macs`, `destination_macs`,
`network_id`, `name`, `acl_index`, `action`) from list output were
reused directly in create and update calls — no translation needed,
no errors, no silent drops.

**Two tool-level friction points** (not data-level):
- `id` in list output vs `rule_id` as the get/update/delete param name
  — resolved on first attempt via clear error message
- `rule_data` dict wrapper on update — discovered via error, self-corrected

Neither is the silent-drop class of bug. Both produced loud errors with
enough context for the agent to self-correct immediately. The core
promise — list output field names round-trip into create/update without
translation — held completely.

## Unit tests

19 tests in `test_acl_tools.py`:

**Model translation (6):** from_controller flattening, to_controller_create
nesting, round-trip preservation, partial update, network_id mapping,
mutable/read-only field metadata.

**Create tool (4):** MACs flow through, no-macs defaults to ANY, preview
includes MACs, response in model shape.

**Update tool (6):** source_macs translated, empty list clears, sibling
fields preserved, unknown fields rejected, read-only fields rejected,
network_id accepted and mapped.

**List + symmetry (3):** model shape output, create-path field symmetry
(every mutable field is a create param), update-path field coverage
(every mutable field handled by to_controller_update).

```
Full suite — 590 passed, 0 failed
```

## Specialist audit

An independent MCP/security specialist audited the implementation against
the upstream spec discussion (#1898), the FastMCP root cause (#136), the
field symmetry rule (#137), and the pilot issue (#139). All checks passed.
Two findings were addressed before submission:

1. `MUTABLE_FIELDS` frozenset comprehension rewritten from
   `not ... is False` to `is not False` for PEP 8 compliance and
   readability (this will be copy-pasted into every domain model).
2. Update-path symmetry test added — asserts every mutable field is
   handled by `to_controller_update` (either in `UPDATE_FIELD_MAP` or
   `MAC_TRANSLATED_FIELDS`), preventing silent drops if a new field is
   added to the model but not the translation layer.

## Review fixups (bundled, commit `a5dea2b`)

During review this PR picked up two fixes that belonged in the pilot
rather than a follow-up (this file is the anchor for future migrations):

1. **Type validation on the update path.** Removing `ACL_RULE_UPDATE_SCHEMA`
   left `update_acl_rule` validating field names but not values. Added
   `validate_update_fields()` in `models/acl.py` using `TypeAdapter` in
   strict mode, so `{"action": "DROP"}`, `{"acl_index": "5"}`, and
   `{"enabled": "yes"}` are rejected at the tool layer with field-named
   error messages. Strict mode is load-bearing — lax mode would coerce
   `"5"` to int and lose the enum guarantees the old JSON Schema had.
2. **`get_acl_rule_details` test coverage.** The tool was refactored to
   use `from_controller().model_dump()` but had no tests in the rewritten
   file. Added `TestGetAclRuleDetails` (4 tests): happy path, empty
   rule_id validation, fallback to list search, and not-found.

Plus 3 update-path type validation tests. Test suite: 590 → 608.

## Follow-up work

- Open issues for migrating other domains (OON, firewall, content
  filtering, port profiles) to the shared model pattern
- Runtime detection shim (#138) as interim protection for unmigrated domains

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -154,6 +154,30 @@ Update tools MUST use the fetch-merge-put pattern. The manager fetches current s
 6. Run `make manifest`
 7. Add tests covering: partial merge preserves unmentioned fields, not-found returns False, empty update is a no-op
 
+### Add or migrate a domain to shared field models
+
+When a tool domain has list/create/update tools, define a shared pydantic model as the single source of truth for field names, types, and mutability. This ensures list output field names are always accepted by create/update tools — preventing silent data loss when callers round-trip fields from list output into create/update calls.
+
+1. Create model in `apps/<server>/src/<pkg>/models/<domain>.py`
+   - One `BaseModel` class with all fields (mutable + read-only)
+   - Read-only fields marked with `json_schema_extra={"mutable": False}`
+   - Export `MUTABLE_FIELDS` and `READ_ONLY_FIELDS` frozensets
+   - Co-locate translation helpers: `from_controller(raw)`, `to_controller_create(model)`, `to_controller_update(fields)`
+   - **Anchor:** `apps/network/src/unifi_network_mcp/models/acl.py`
+2. Refactor tool functions to derive I/O from the model
+   - List/get tools: `from_controller(raw).model_dump()`
+   - Create tool: build model from params → `to_controller_create()` → manager
+   - Update tool: validate keys against `MUTABLE_FIELDS`, translate via `to_controller_update()` → manager
+   - **Anchor:** `apps/network/src/unifi_network_mcp/tools/acl.py`
+3. Retire the domain's JSON Schema from `schemas.py` and `validator_registry.py`
+   - The pydantic model replaces the JSON Schema for validation
+   - Leave a comment noting the migration for other contributors
+4. Manager layer is unchanged — continues to speak the controller API dialect
+5. Add a field symmetry test asserting every mutable field is a create param
+   - **Anchor:** `apps/network/tests/unit/test_acl_tools.py:TestListAclRules.test_list_and_create_field_symmetry`
+6. Run `make manifest`
+7. Commit model + refactored tools + retired schema + tests together
+
 ### Modify the permission system
 
 1. Shared logic: `packages/unifi-mcp-shared/src/unifi_mcp_shared/policy_gate.py`
--- a/apps/network/src/unifi_network_mcp/models/acl.py
+++ b/apps/network/src/unifi_network_mcp/models/acl.py
@@ -11,9 +11,9 @@ described in AGENTS.md. Other domains should follow this pattern.
 
 from __future__ import annotations
 
-from typing import Any, Dict, List, Literal, Optional
+from typing import Any, Dict, List, Literal, Optional, Tuple
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, TypeAdapter, ValidationError
 
 
 class AclRule(BaseModel):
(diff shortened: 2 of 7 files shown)


level99:
Pushed a small rider fix: updated `rule_id` param descriptions on get/update/delete from `"Unique identifier (_id)"` to `"The id field from unifi_list_acl_rules output"` — directly addresses friction point #1 from the zero-context LLM test.

Friction point #2 (update takes a `rule_data` dict wrapper vs create taking individual params) is a project-wide convention across all update tools — the fetch-merge-put golden path uses `*_data: dict` for partial updates. Changing it for ACL alone would create inconsistency; changing it project-wide is a separate design decision. Leaving as-is since the error message is clear enough for self-correction.

sirkirby (approved):
Thanks for building this. Everything looks good. I did a minor update to add in the validation and just some additional testing.

PR #142, opened 2026-04-11
feat(tool-index): add category/search/names_only filters to fix token overflow

riichard:
## Problem

The full tool index (166 tools with complete JSON schemas) returns ~185K characters, which exceeds MCP client token limits. This makes `unifi_tool_index` unusable in clients like Claude Code.

## Solution

Add three optional filter parameters to `unifi_tool_index`:

| Parameter | Type | Default | Description |
|-----------|------|---------|-------------|
| `names_only` | boolean | `true` | Return only tool names — ~4K chars vs 185K |
| `category` | string | — | Filter to one area: `clients`, `firewall`, `devices`, `network`, `stats`, `switch`, `vpn`, `dns`, `routing`, etc. |
| `search` | string | — | Case-insensitive substring match on tool name |

The response always includes a `categories` array listing all valid category values.

## Usage

```
# Compact overview (default) — ~4K chars
unifi_tool_index {}

# Full schemas for one area
unifi_tool_index {"category": "clients", "names_only": false}

# Find tools by keyword
unifi_tool_index {"search": "firewall"}
```

## Bug fix

`apps/network/src/unifi_network_mcp/tool_index.py` had its own `tool_index_handler` that silently dropped the `args` parameter, so filters passed by the MCP client were never applied. Fixed in this PR.

## Test plan

- [ ] `unifi_tool_index {}` returns ~4K chars with all 166 tool names
- [ ] `unifi_tool_index {"category": "clients"}` returns only the 11 client tools
- [ ] `unifi_tool_index {"category": "clients", "names_only": false}` returns full schemas for client tools
- [ ] `unifi_tool_index {"search": "firewall"}` returns matching tools
- [ ] Response always includes `categories` array

### DIFF
--- a/apps/network/src/unifi_network_mcp/tool_index.py
+++ b/apps/network/src/unifi_network_mcp/tool_index.py
@@ -12,8 +12,8 @@ from unifi_mcp_shared.tool_index import (
 )
 
 
-def get_tool_index():
-    """Get the complete tool index, using network server's registration mode and manifest."""
+def get_tool_index(category=None, search=None, names_only=True):
+    """Get the tool index, using network server's registration mode and manifest."""
     try:
         from unifi_network_mcp.bootstrap import UNIFI_TOOL_REGISTRATION_MODE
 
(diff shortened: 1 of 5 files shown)


sirkirby (commented):
Thanks for flagging this, Richard — the tool index size is a real problem worth solving, and you're right that the full ~127K char response exceeds Claude Code's default `MAX_MCP_OUTPUT_TOKENS` cap (~25K tokens). Good catch on the `tool_index_handler` silently dropping `args` too.

That said, I'm going to close this in favor of a tracked issue for a few reasons:

**Scope** — This bundles three separate features (tool index filtering, structured error responses, `include_fields` on detail tools) that should each be evaluated independently.

**Framing** — There's no "MCP client token limit" in the protocol spec. The constraint is Claude Code's soft `MAX_MCP_OUTPUT_TOKENS` default, which is configurable via env var and overridable per-tool via the `anthropic/maxResultSizeChars` annotation. Understanding the actual constraint matters for choosing the right fix.

**Implementation concerns:**
- `names_only` strips descriptions despite the schema saying "name + description" — bare names are too sparse for tool discovery
- `names_only=True` as the default is a breaking change that forces two round-trips (names → filter → schemas) instead of one
- Categories derived from internal module paths couples public API to file structure
- Inconsistent defaults between `get_tool_index()` (False) and the handler (True)
- No tests, no `make manifest` run

I'm opening #143 to track the actual problem and evaluate approaches (names+descriptions default, `maxResultSizeChars` annotation, pagination, etc.). Appreciate you surfacing this — it's going into the roadmap.

sirkirby:
Closing in favor of #143 (will create momentarily). The problem is real — we'll track the solution there.

Issue #113, opened 2026-03-28
Improve LLM usability of update tools with nested dict parameters


level99:
## Problem

Several update tools accept a nested `dict` parameter (e.g., `update_data`, `rule_data`, `profile_data`) that LLM callers consistently get wrong on first attempt. The LLM passes fields flat instead of nested, gets a Pydantic validation error, then self-corrects — wasting a tool call every time.

Example failure (`update_network`):
```json
// LLM tries this (flat):
{"network_id": "abc123", "dhcpd_enabled": false, "confirm": true}
// Error: update_data Field required

// Then corrects to (nested):
{"network_id": "abc123", "update_data": {"dhcpd_enabled": false}, "confirm": true}
```

Additionally, some PUT-style update tools require the caller to send the complete object but don't auto-merge with current state, causing silent failures when only changed fields are sent.

## Affected tools (10)

**Nested dict params (nesting confusion):**
- `unifi_update_acl_rule` — `rule_data: dict`
- `unifi_update_network` — `update_data: Dict[str, Any]`
- `unifi_update_wlan` — `update_data: Dict[str, Any]`
- `unifi_update_port_profile` — `profile_data: dict`
- `unifi_update_client_group` — `group_data: dict`
- `unifi_update_content_filter` — `filter_data: dict`
- `unifi_update_oon_policy` — `policy_data: dict`
- `unifi_update_port_forward` — `update_data: Dict[str, Any]`
- `unifi_update_qos_rule` — `update_data: Dict[str, Any]`
- `unifi_create_oon_policy` — `policy_data: dict`

**Full object required, no auto-merge (silent failure on partial):**
- `update_acl_rule`, `update_client_group`, `update_content_filter`, `update_oon_policy`, `update_port_profile`

**Tools that already handle this well (for reference):**
- `update_network`, `update_qos_rule` — accept partial updates with validation
- `update_device_radio`, `update_route`, `set_client_ip_settings` — flat individual params
- `create_port_profile` — refactored from dict to flat params in #111

### Option A: Fetch-merge-put in the manager layer

Tool accepts partial dict from caller. Manager auto-fetches current object, merges caller's changes on top, PUTs the complete merged object.

| | |
|---|---|
| Pros | LLMs can send just `{"enabled": false}` and it works. No nesting confusion (still accepts a dict, but partial is fine). No silent failures. Proven pattern — `update_device_config` in #111 does this. Matches how the UI works. |
[...]

PR #146, opened 2026-04-12
fix(network): fix zone-based firewall policy creation for firmware 5.x

MizterB:
## Summary

Fixes `unifi_create_firewall_policy` for zone-based (v2 API) policies on UDM Pro firmware 5.x / Network 10.x. Previously, `confirm: true` silently failed with a generic error; the controller's actual rejection reason was never surfaced.

### Root causes found (in order of discovery)

1. **Error swallowed in manager** — `create_firewall_policy` returned `None` on both API errors and unexpected responses, so the tool layer hit a "Check server logs" branch with no detail. Fixed by raising exceptions so the controller's `errorCode`/`message` propagate to the MCP client.

2. **Missing required fields** — The UDM Pro v2 API rejects requests where `schedule` or `ip_version` are absent (Spring `@NotNull`). The schema already had a default for `ip_version` (`"BOTH"`) but `schedule` had none, and neither default was actually applied to the outbound payload.

3. **Wrong `create_allow_respond` default** — Schema defaulted to `True`, but the API returns `api.err.FirewallPolicyCreateRespondTrafficPolicyNotAllowed` for BLOCK/REJECT policies. Changed to `False`.

### Changes

- **`firewall_manager.py`**: raise exceptions instead of `return None` so API errors surface with full detail; add INFO logging of outbound payload
- **`schemas.py`**: add `"default": {"mode": "ALWAYS"}` to `schedule` in `FIREWALL_POLICY_V2_CREATE_SCHEMA`; change `create_allow_respond` default to `False`
- **`validators.py`** (shared): `ResourceValidator.validate()` now applies schema-defined `default` values to any top-level property absent from the input — making schema defaults authoritative across all resource types

## Test plan

- [x] `unifi_create_firewall_policy` with `confirm: false` returns correct preview
- [x] `unifi_create_firewall_policy` with `confirm: true` successfully creates a zone-based BLOCK policy (`matching_target: ANY` on both sides) on UDM Pro firmware 5.0.16 / Network 10.x
- [x] `unifi_create_firewall_policy` with `confirm: true` successfully creates a zone-based ALLOW policy
- [x] 15 zone-to-zone policies (BLOCK and ALLOW) created in batch without errors
- [ ] Verify existing legacy (ruleset-based) policy creation still works

### DIFF
--- a/apps/network/src/unifi_network_mcp/managers/firewall_manager.py
+++ b/apps/network/src/unifi_network_mcp/managers/firewall_manager.py
@@ -619,7 +619,7 @@ class FirewallManager:
         try:
             policy_name = policy_data.get("name", "Unnamed Policy")
             logger.info("Attempting to create firewall policy '%s' via V2 endpoint.", policy_name)
-            logger.info("Firewall policy create payload: %s", json.dumps(policy_data, indent=2))
+            logger.debug("Firewall policy create payload: %s", json.dumps(policy_data, indent=2))
 
             api_request = ApiRequestV2(method="post", path="/firewall-policies", data=policy_data)
 
--- a/apps/network/src/unifi_network_mcp/managers/firewall_manager.py
+++ b/apps/network/src/unifi_network_mcp/managers/firewall_manager.py
@@ -619,8 +619,7 @@ class FirewallManager:
         try:
             policy_name = policy_data.get("name", "Unnamed Policy")
             logger.info("Attempting to create firewall policy '%s' via V2 endpoint.", policy_name)
-            # Log the payload for debugging, ensuring sensitive data isn't exposed if necessary
-            # logger.debug("Firewall policy create payload: %s", json.dumps(policy_data, indent=2))
+            logger.info("Firewall policy create payload: %s", json.dumps(policy_data, indent=2))
 
             api_request = ApiRequestV2(method="post", path="/firewall-policies", data=policy_data)
 
@@ -644,31 +643,18 @@ class FirewallManager:
                 logger.error(
                     "Failed to create firewall policy '%s'. Unexpected V2 response format: %s", policy_name, response
                 )
-                return None
+                raise RuntimeError(
+                    "Unexpected response from controller (no _id in response). Raw: %s" % json.dumps(response, default=str)
+                )
 
         except Exception as e:
-            # Attempt to extract a more specific error message if possible
-            api_error_message = str(e)
-            if hasattr(e, "args") and e.args:
-                try:
-                    error_details = e.args[0]
-                    if isinstance(error_details, dict) and "message" in error_details:
-                        api_error_message = error_details["message"]
-                    elif isinstance(error_details, str):
-                        api_error_message = error_details
-                except Exception as parse_exc:
-                    logger.warning(
-                        "Could not parse specific API error from exception args: %s. Parse error: %s", e.args, parse_exc
-                    )
-
             logger.error(
                 "Error creating firewall policy '%s' via V2: %s",
                 policy_data.get("name", "Unnamed Policy"),
-                api_error_message,
+                e,
                 exc_info=True,
             )
-            # Optionally re-raise or return a custom error object instead of None
-            return None
+            raise
 
     async def delete_firewall_policy(self, policy_id: str) -> bool:
         """Delete a firewall policy by ID.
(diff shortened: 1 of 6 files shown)


sirkirby (dismissed):
Thanks for the excellent debugging work here, MizterB — the root cause analysis is thorough and the real-hardware testing on UDM Pro 5.x is exactly what we need. The schema fixes (schedule default, create_allow_respond flip) are correct.

Two things need fixing before we can merge:

### 1. `ResourceValidator.validate()` default injection — blast radius too wide

This change lives in `packages/unifi-mcp-shared` and affects every validator across all apps (network, protect, access). The problem: **update tools** intentionally omit fields to mean "don't change this." With this change, calling any update tool with a partial payload would silently inject schema defaults, overwriting existing settings — the exact class of bug we track in #113.

**Suggested fix:** Move the default injection out of the shared validator and into the create tool's code path in `firewall.py`, right after validation succeeds. That keeps the fix scoped to where it's needed without affecting every other tool.

### 2. `migrate_firewall_rule_to_policy` broken by raise-instead-of-None

`firewall.py:713` calls `create_firewall_policy` and checks `if created is None:`, but the manager now raises instead of returning None. This caller has no try/except — API errors will propagate as unhandled exceptions. The other caller (line 393) is fine since it's already wrapped.

**Fix:** Add a try/except around the call at line 713, matching the pattern at line 392-414.

### Minor

- `logger.info` for the full payload dump (line 622) should be `logger.debug` — too noisy for production and could log sensitive data.
- Your own test plan has the legacy policy check unchecked — worth confirming before we merge.

Happy to help if you have questions on any of this.

sirkirby:
Hi @MizterB — I pushed a follow-up commit (fb495d7) to address the two merge blockers and the two minor items from the review. Wanted to land it for you rather than sit on the PR, hope that's OK. Your authorship and the original commit are preserved; this is stacked on top as a single maintainer edit.

**What changed:**

1. **`validators.py` — blast-radius fix.** Reverted `ResourceValidator.validate()` to its pre-PR behaviour (never injects defaults), and added `validate_and_apply_defaults()` as an opt-in helper with identical semantics to your original logic. The concern was that every validator in the codebase — network, protect, access update tools — would have silently filled missing fields with schema defaults, which for update tools means "overwrite existing state with whatever the schema said the default was." Same class of bug as #113. The opt-in split keeps your zone-based fix working while leaving all other tools unaffected.

2. **`tools/firewall.py` — zone path uses the opt-in helper.** `create_firewall_policy` now calls `validate_and_apply_defaults` on the V2/zone-based schema branch and plain `validate` on the legacy branch. End result for the V2 path is identical to what your PR produced — `schedule`, `create_allow_respond`, and `ip_version` all get filled from the schema.

3. **`tools/firewall.py` — migrate caller guarded.** Wrapped the `firewall_manager.create_firewall_policy` call in `migrate_firewall_rule_to_policy` with try/except matching the pattern at lines 392–414. Since the manager now raises on API errors, the previous bare call would have propagated `AiounifiException` as an unhandled exception on any migration failure.

4. **`firewall_manager.py` — payload log demoted.** `logger.info` → `logger.debug` for the payload dump.

5. **`test_validators.py` — regressions added.** One test confirming plain `validate()` never injects defaults (locks in the blast-radius guard), four tests covering the new opt-in helper.

**Verified end-to-end on my UDM Pro (firmware 5.x / UniFi Network 10.2.105):**

- Plain `validate()` on a sparse V2 payload returns params untouched.
- `validate_and_apply_defaults()` fills `schedule: {mode: ALWAYS}`, `create_allow_respond: false`, `ip_version: BOTH`.
- Sparse zone-based BLOCK policy (`matching_target: ANY` on both sides) accepted by the controller and created successfully via the tool path — which was your original bug. Test policy cleaned up afterwards.
- Manager raises `AiounifiException` on bad input instead of silently returning None.
- 585 unit tests pass across the network app and shared validators package.

The unchecked box on your test plan (legacy policy creation) I couldn't fully verify on this controller — it routes all firewall creates through the V2 endpoint now, so a legacy-shaped payload gets rejected at the controller level regardless of validator. That behaviour is unrelated to this PR and was the same before your changes. Your schema-level validation for the legacy path still works.

sirkirby (approved):
G2G

MizterB:
@sirkirby No problem at all. I’ve been traveling and unable to implement the feedback yet. I appreciate your help in making this work better for all. Thank you!

PR #155, opened 2026-04-26
fix(network): list_firewall_zones — correct V2 path for Network 10.2+, surface errors, trim O(N²) payload, align response shape

### LINKED ISSUE #154: unifi_list_firewall_zones returns empty success on Network 10.2.x — wrong V2 API path + error is silently swallowed
## Summary
`unifi_list_firewall_zones` returns `{"success": true, "count": 0, "zones": []}` on UniFi Network application **10.2.105**, even when multiple zones exist on the controller. There are actually two bugs stacked:

1. The V2 API path `/firewall/zones` returns **404 Not Found** on Network 10.2.x. The endpoint was apparently moved/renamed in firmware 5.0+ / Network 10.2+.
2. `FirewallManager.get_firewall_zones()` catches the resulting exception and returns an empty list. The tool wrapper then reports `success: true` to the client, making the failure invisible. This is the same class of bug fixed for `create_firewall_policy` in #146.

## Environment
- `unifi-network-mcp` **0.14.9** (`ghcr.io/sirkirby/unifi-network-mcp:0.14.9`)
- UDM SE running UniFi Network **10.2.105** (build `atag_10.2.105_33556`)
- Zone-based firewall is fully migrated/enabled on the controller (`site-feature-migration` shows `ZONE_BASED_FIREWALL` migrated)

## Reproduction
```
unifi_list_firewall_zones
→ {"success": true, "count": 0, "zones": []}
```

## Evidence the zones exist
Calling `unifi_list_firewall_policies` returns policies referencing zone IDs beyond the standard five — including newly-created zones with later ObjectId timestamps:

```
675ddd33aaad975be1cd8431  (stock)
675ddd33aaad975be1cd8432  (stock)
675ddd33aaad975be1cd8433  (stock)
675ddd33aaad975be1cd8434  (stock)
675ddd33aaad975be1cd8435  (stock)
675ddd33aaad975be1cd8436  (additional)
69ee319021611449c08b76cb  (user-created — different ObjectId timestamp)
```

## Server log shows the underlying 404
```
ERROR - API request error: GET /firewall/zones - Call https://<udm>:443/proxy/network/v2/api/site/default/firewall/zones received 404 Not Found
INFO  - TOOL {"tool": "unifi_list_firewall_zones", "duration_ms": 34, ..., "result": {"success": true, "count": 0, "zones": []}}
```

## Direct API confirmation
Hitting the path the MCP uses, against the same UDM, with valid CSRF + cookie:
```
GET /proxy/network/v2/api/site/default/firewall/zones → 404
```
While `firewall-policies` works fine on the same V2 prefix:
```
GET /proxy/network/v2/api/site/default/firewall-policies → 200
```
Probed several alternate paths (`firewall-zone`, `firewall-zones`, `firewall-zone-matrix`, `zone-policy`, `firewall-policies/zones`, `rest/firewallzone`, etc.) — none returned 200. The actual path used by the Network 10.2 UI is unknown to me; happy to capture it from the browser's network tab if helpful.

## Code location

`apps/network/src/unifi_network_mcp/managers/firewall_manager.py` (current `main` and tag `network/v0.14.9`):

```python
async def get_firewall_zones(self) -> List[Dict[str, Any]]:
    """Return list of firewall zones via V2 API."""
    ...
    try:
        api_request = ApiRequestV2(method="get", path="/firewall/zones")
        resp = await self._connection.request(api_request)
        data = resp if isinstance(resp, list) else resp.get("data", []) if isinstance(resp, dict) else []
        ...
        return data
    except Exception as e:
        logger.error("Error fetching firewall zones: %s", e)
        return []
```

## Suggested fix
1. Identify and use the correct V2 zones path for Network 10.2+ (or fall back across known paths).
2. Stop swallowing the exception — let it propagate so the tool wrapper surfaces the real error to the client (matching the pattern adopted in #146 for policy creation).

nalditopr:
Closes #154.

Four stacked fixes for `unifi_list_firewall_zones`, verified end-to-end against UDM SE running UniFi Network 10.2.105 (12 zones: 6 stock + 6 user-created).

## 1. Use the correct V2 API path

The endpoint moved on Network 10.2+:

| Network app | Path |
|---|---|
| ≤ 10.1 | `/firewall/zones` |
| **≥ 10.2** | **`/firewall/zone-matrix`** |

The MCP previously hardcoded `/firewall/zones`, which 404s on current firmware. The new path is also the one the official Network 10.x web UI calls (verified via browser DevTools).

The fix tries `/firewall/zone-matrix` first and falls back to `/firewall/zones` so older firmware keeps working.

## 2. Surface API errors instead of swallowing them

Same pattern adopted for `create_firewall_policy` in #146:

- **Manager** (`firewall_manager.py::get_firewall_zones`): re-raise the exception (with `exc_info=True`) so the controller's actual error reaches the caller.
- **Tool wrapper** (`tools/firewall.py::list_firewall_zones`): catch and return `{"success": False, "error": "..."}`.

Previously, a 404 (or any other API failure) was caught and converted into an empty list, which the tool wrapper then reported as `success: true` — making path-mismatch and other failures completely invisible.

## 3. Trim the response payload

The `/firewall/zone-matrix` response embeds an inter-zone matrix (`data` array) on every zone — O(N²) entries, not zone metadata. For a "list zones" call we only return zone metadata; the matrix would be a separate dedicated tool if there is demand.

Concretely, on a controller with 12 zones the response went from ~11 KB to ~1.5 KB.

## 4. Align response shape with neighboring list tools

Match the shape used by `list_firewall_policies` / `list_firewall_groups` in the same file:

- top-level `site` field
- per-entry `id` (not raw `_id`)
- explicit projection (`id`, `name`, `zone_key`) so callers do not depend on whatever extra fields the V2 endpoint happens to return

## End-to-end output (Network 10.2.105, 12 zones)

```json
{
  "success": true,
  "site": "default",
  "count": 12,
  "zones": [
    {"id": "675ddd33aaad975be1cd8431", "name": "Internal",      "zone_key": "internal"},
    {"id": "675ddd33aaad975be1cd8432", "name": "External",      "zone_key": "external"},
    {"id": "675ddd33aaad975be1cd8433", "name": "Gateway",       "zone_key": "gateway"},
    {"id": "675ddd33aaad975be1cd8434", "name": "Vpn",           "zone_key": "vpn"},
    {"id": "675ddd33aaad975be1cd8435", "name": "Hotspot",       "zone_key": "hotspot"},
    {"id": "675ddd33aaad975be1cd8436", "name": "Dmz",           "zone_key": "dmz"},
    {"id": "69ee319021611449c08b76cb", "name": "<custom>",      "zone_key": ""},
    {"id": "69ee319d21611449c08b7766", "name": "<custom>",      "zone_key": ""},
    ...
  ]
}
```

On a 404 (path mismatch) the response is now:

```json
{
  "success": false,
  "error": "Failed to list firewall zones: Call .../firewall/zones received 404 Not Found"
}
```

…instead of the misleading `success: true, count: 0, zones: []`.

## Test plan

- [x] Network 10.2.105: returns all 12 expected zones with name + zone_key in the canonical shape.
- [x] Fallback path is exercised when the primary returns 404.
- [x] Both paths failing returns a structured `success: false` with the controller's actual error (validated by manually pointing the manager at a bad path).
- [ ] No automated tests existed for `list_firewall_zones` / `get_firewall_zones`; happy to add some if you'd like.

## Commits

1. `surface errors from list_firewall_zones instead of swallowing` — error visibility
2. `use /firewall/zone-matrix on Network 10.2+ for list_firewall_zones` — correct path with fallback
3. `drop zone-matrix payload from list_firewall_zones response` — payload trim
4. `match list_firewall_zones response shape to other list tools` — shape alignment with `list_firewall_policies` / `list_firewall_groups`

### DIFF
--- a/apps/network/src/unifi_network_mcp/managers/firewall_manager.py
+++ b/apps/network/src/unifi_network_mcp/managers/firewall_manager.py
@@ -703,7 +703,7 @@ class FirewallManager:
             raise ConnectionError("Not connected to controller")
         try:
             # Network 10.2+ exposes zones at /firewall/zone-matrix (returns
-            # zone metadata plus inter-zone policy counts).
+            # zone metadata plus an inter-zone policy-count matrix).
             # Older firmware exposed a flat list at /firewall/zones; try that
             # as a fallback so this works across versions.
             try:
@@ -717,6 +717,15 @@ class FirewallManager:
                 api_request = ApiRequestV2(method="get", path="/firewall/zones")
                 resp = await self._connection.request(api_request)
             data = resp if isinstance(resp, list) else resp.get("data", []) if isinstance(resp, dict) else []
+            # The zone-matrix endpoint includes a `data` field per zone that
+            # contains the policy-count matrix to every other zone (O(N^2)
+            # payload). For a zones listing we only want the zone metadata,
+            # so drop the matrix field if present. The matrix is still
+            # available via a dedicated tool if needed.
+            data = [
+                {k: v for k, v in zone.items() if k != "data"} if isinstance(zone, dict) else zone
+                for zone in data
+            ]
             self._connection._update_cache(cache_key, data)
             return data
         except Exception as e:
--- a/apps/network/src/unifi_network_mcp/managers/firewall_manager.py
+++ b/apps/network/src/unifi_network_mcp/managers/firewall_manager.py
@@ -708,7 +708,7 @@ class FirewallManager:
             self._connection._update_cache(cache_key, data)
             return data
         except Exception as e:
-            logger.error("Error fetching firewall zones: %s", e)
+            logger.error("Error fetching firewall zones: %s", e, exc_info=True)
             raise
 
     # ---- Firewall Groups (v1 REST: address-group, port-group) ----
--- a/apps/network/src/unifi_network_mcp/managers/firewall_manager.py
+++ b/apps/network/src/unifi_network_mcp/managers/firewall_manager.py
@@ -702,8 +702,20 @@ class FirewallManager:
         if not await self._connection.ensure_connected():
             raise ConnectionError("Not connected to controller")
         try:
-            api_request = ApiRequestV2(method="get", path="/firewall/zones")
-            resp = await self._connection.request(api_request)
+            # Network 10.2+ exposes zones at /firewall/zone-matrix (returns
+            # zone metadata plus inter-zone policy counts).
+            # Older firmware exposed a flat list at /firewall/zones; try that
+            # as a fallback so this works across versions.
+            try:
+                api_request = ApiRequestV2(method="get", path="/firewall/zone-matrix")
+                resp = await self._connection.request(api_request)
+            except Exception as primary_exc:
+                logger.debug(
+                    "Primary /firewall/zone-matrix failed (%s), falling back to /firewall/zones",
+                    primary_exc,
+                )
+                api_request = ApiRequestV2(method="get", path="/firewall/zones")
+                resp = await self._connection.request(api_request)
             data = resp if isinstance(resp, list) else resp.get("data", []) if isinstance(resp, dict) else []
             self._connection._update_cache(cache_key, data)
             return data
--- a/apps/network/src/unifi_network_mcp/tools/firewall.py
+++ b/apps/network/src/unifi_network_mcp/tools/firewall.py
@@ -747,10 +747,23 @@ async def create_simple_firewall_policy(
 async def list_firewall_zones() -> Dict[str, Any]:
     try:
         zones = await firewall_manager.get_firewall_zones()
+        formatted = [
+            {
+                "id": z.get("_id"),
+                "name": z.get("name"),
+                "zone_key": z.get("zone_key", ""),
+            }
+            for z in zones
+        ]
+        return {
+            "success": True,
+            "site": firewall_manager._connection.site,
+            "count": len(formatted),
+            "zones": formatted,
+        }
     except Exception as exc:
         logger.error("Error listing firewall zones: %s", exc, exc_info=True)
         return {"success": False, "error": f"Failed to list firewall zones: {exc}"}
-    return {"success": True, "count": len(zones), "zones": zones}
 
 
 # ---- Firewall Groups (address-group, port-group) ----
--- a/apps/network/src/unifi_network_mcp/tools/firewall.py
+++ b/apps/network/src/unifi_network_mcp/tools/firewall.py
@@ -745,7 +745,11 @@ async def create_simple_firewall_policy(
     annotations=ToolAnnotations(readOnlyHint=True, openWorldHint=False),
 )
 async def list_firewall_zones() -> Dict[str, Any]:
-    zones = await firewall_manager.get_firewall_zones()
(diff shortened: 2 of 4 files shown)


sirkirby (changes requested):
Thanks for the detailed PR — I validated this end-to-end against my own UDM SE on Network 10.2.x and everything checks out:

- `/firewall/zones` → 404 (reproduced)
- `/firewall/zone-matrix` → 200, returns zones with `_id`, `name`, `zone_key`, `data`
- Stripped payload + tool projection produces the clean shape you documented
- Existing 50 firewall tests still pass on the branch

The implementation is also a clean mirror of the PR #146 pattern (manager re-raises with `exc_info=True`, tool wrapper returns `{"success": False, "error": ...}`), which is exactly what we want here.

## One ask before merge: tests

You called this out yourself — there's no automated coverage for `get_firewall_zones` / `list_firewall_zones`, and this PR adds enough branching that I'd like to lock it in. Three cases worth covering:

1. **Primary path success** — `/firewall/zone-matrix` returns zones; verify the matrix `data` field is stripped and the tool output projects `id`/`name`/`zone_key`.
2. **Fallback path** — primary raises, fallback `/firewall/zones` succeeds; verify the fallback path is taken and zones are returned.
3. **Both fail** — exception propagates from the manager and the tool wrapper returns `{"success": False, "error": ...}` (not the old empty-list-as-success).

`apps/network/tests/unit/test_firewall_manager.py` is the right place for the manager-level tests; `test_firewall_tools.py` for the tool wrapper. Both files have the mocking pattern (`AsyncMock` on `_connection.request`) you'll need.

## Follow-up — separate scope, not for this PR

Reviewing this surfaced a broader pattern: the `if not await self._connection.ensure_connected(): return []` guard at the top of these manager methods can still produce `success: True, count: 0` when the controller isn't reachable. That's the same class of silent failure you fixed for the API-error path here, just at a different layer. I'm opening a separate issue to track that across the manager files — explicitly out of scope for this PR.

Once tests land, this is good to merge. Thanks again for the thorough write-up and the live verification.

sirkirby (changes requested):
Tests to be added

sirkirby:
Heads up — #157 just merged on `main` and includes changes to `firewall_manager.py::get_firewall_zones` (the broad `except Exception: return []` block was converted to bare `raise` as part of the silent-failure fix in #156). When you rebase to add the tests, you'll hit a small conflict on lines 731-735 of `firewall_manager.py`. Resolution: keep your new fallback logic, but the outer `except Exception` in your version should end with `raise` (matching the new pattern) instead of `return []`. Happy to help if anything's unclear.

sirkirby:
Pushed test coverage to your branch (`eb96fb1`) — three manager tests + two tool-wrapper tests covering primary `/firewall/zone-matrix`, fallback to `/firewall/zones`, both-fail propagation, the `_id`→`id` projection, and the structured-error regression guard. All 55 firewall tests pass locally on Python 3.13.

Once CI is green I'll dismiss the prior changes-requested review and merge. Thanks for sticking with this — implementation has been solid throughout.

sirkirby (approved):
shipit

PR #175, opened 2026-04-30
PR3: Access — UniFiNotFoundError standardization

sirkirby:
## Summary

This is **PR3** of the manager-owned-existence-checks refactor (siblings: merged #172 network, merged #173 protect).

Audit of `apps/access/` flagged 8 lookup-then-act candidates, all of which turned out to be preview/execute pattern (`apply_X` + `X`) — same shape as protect. Those are intentional two-method splits, not lookup-then-act, so they need PR4 `DISPATCH_OVERRIDES` rather than refactor.

This PR lands the smaller, useful piece: **standardize the existence-check exception type** across access managers from `ValueError` to `UniFiNotFoundError`. Mirrors PR #173 for protect; aligns access with network from PR #172.

## Changes

**Manager helpers** (5 files):
- `door_manager.get_door`, `door_manager.get_door_status` (2 sites)
- `device_manager.get_device` (1 site)
- `event_manager.get_event` (2 sites including bare-except fallback)
- `credential_manager.get_credential` 404 path (1 site)
- `visitor_manager.get_visitor` 404 path (1 site)

All now raise `UniFiNotFoundError("<resource>", id)` instead of `ValueError`.

**Tool exception handlers** (6 files): `except ValueError as e` → `except (UniFiNotFoundError, ValueError) as e` so existing tools surface the new exception type without losing the validation-bound `ValueError` paths (empty IDs, missing required fields).

**Tests** (3 files): `pytest.raises(ValueError, match="X not found")` → `pytest.raises(UniFiNotFoundError)` for the 3 not-found assertions.

## Test deltas

| Package | Before | After | Δ |
|---|---|---|---|
| unifi-core | 69 | 69 | 0 |
| unifi-mcp-shared | 205 | 205 | 0 |
| unifi-network-mcp | 630 | 630 | 0 |
| unifi-protect-mcp | 336 | 336 | 0 |
| unifi-access-mcp | 157 | 157 | 0 |
| unifi-api | 336 | 336 | 0 |

## Deferred to PR4 DISPATCH_OVERRIDES (preview/execute split, 8 tools)

- `credentials.py`: `access_create_credential`, `access_revoke_credential`
- `devices.py`: `access_reboot_device`
- `doors.py`: `access_lock_door`, `access_unlock_door`
- `policies.py`: `access_update_policy`
- `visitors.py`: `access_create_visitor`, `access_delete_visitor`

## Test plan

- [x] All 6 package suites pass
[...]

PR #173, opened 2026-04-30
PR2: Protect — UniFiNotFoundError standardization

sirkirby:
## Summary

This is **PR2** of the manager-owned-existence-checks refactor (sibling to merged #172). 

Audit of `apps/protect/` flagged 11 lookup-then-act candidates, but on inspection **nearly all are preview/execute pattern** — separate `preview_X` and `apply_X` methods that the AST walker captures as two awaits. That's a deliberate two-method split, not lookup-then-act, so those tools need PR4 `DISPATCH_OVERRIDES` rather than refactor.

This PR lands the smaller, useful piece: **standardize the existence-check exception type** across protect managers from `ValueError` to the new `UniFiNotFoundError` introduced in PR1 (#172). Aligns protect with network and gives the rich-API dispatch layer a typed exception to catch in PR4.

## Changes

**Manager helpers** (5 files): `camera_manager._get_camera`, `chime_manager._get_chime`, `light_manager._get_light`, `liveview_manager._get_liveview`, `recording_manager._get_camera`, plus `event_manager.get_event` / `get_event_thumbnail` / `acknowledge_event` / `apply_acknowledge_event` — all now raise `UniFiNotFoundError("<resource>", id)` instead of `ValueError(f"X not found: {id}")`.

**Tool exception handlers** (6 files): `except ValueError as e` → `except (UniFiNotFoundError, ValueError) as e` so existing tools surface the new exception type without losing PTZ-not-supported / validation-bound `ValueError` paths.

**Tests** (5 files): `pytest.raises(ValueError, match="X not found")` → `pytest.raises(UniFiNotFoundError)` for the 15 not-found assertions across cameras / devices / events / liveviews / recordings. Other `ValueError` raises (validation, PTZ-capability) stay unchanged.

## Test deltas

| Package | Before | After | Δ |
|---|---|---|---|
| unifi-core | 69 | 69 | 0 |
| unifi-mcp-shared | 205 | 205 | 0 |
| unifi-network-mcp | 630 | 630 | 0 |
| unifi-protect-mcp | 336 | 336 | 0 |
| unifi-access-mcp | 157 | 157 | 0 |
| unifi-api | 336 | 336 | 0 |

## Deferred to PR4 DISPATCH_OVERRIDES (preview/execute split, ~11 tools)

These tools have legitimate two-method bodies because preview state is a distinct manager call from the mutation. The dispatch override pattern from spec §4.4 is the right fix:

[...]

PR #176, opened 2026-04-30
PR4: dispatch overrides + stats serializer re-registration (finale)

sirkirby:
## Summary

**Finale of the manager-owned-existence-checks refactor.** Siblings: #172 (network managers), #173 (protect), #175 (access).

After PR1–PR3, manager existence checks raise `UniFiNotFoundError` consistently, and ~19 simple lookup-then-act tools resolved through pure refactor. The remaining ~45 compose-case tools have legitimate two-method bodies and need the **dispatch override pattern** from spec §4.4 — that's what this PR lands.

### 1. New `DISPATCH_OVERRIDES` static table

`apps/api/src/unifi_api/services/dispatch_overrides.py` — maps 45 tool names to `(manager_attr, method)` tuples. Covers two structural patterns:

- **Lookup-then-act with state-dependent preview** (network/clients): tool body genuinely needs current resource state to render a useful preview (current device name before reboot, old WLAN config before update, toggle's current enabled flag). Refactoring would strip a real UX feature.
- **Preview/execute split** (protect/access): managers expose `X(id)` for preview state and `apply_X(id)` for execution; both awaited conditionally on `confirm`.

### 2. Wires overrides into `build_dispatch_table`

`apps/api/src/unifi_api/services/actions.py` — AST walk runs first, overrides applied after, so override entries always win.

### 3. Re-registers `unifi_get_device_stats` and `unifi_get_client_stats` as TIMESERIES

`apps/api/src/unifi_api/serializers/network/stats.py` — pre-PR4 these were DETAIL because AST captured `get_X_details` (a single dict). PR4's override redirects dispatch at `stats_manager.get_X_stats` (list of points), so TIMESERIES is now correct.

- `unifi_get_top_clients` stays LIST (per-client aggregates, not per-point series).
- `unifi_get_dpi_stats` stays DETAIL (`{applications: [], categories: []}` shape).

### 4. Updates resource routes for the new shape

`apps/api/src/unifi_api/routes/resources/network/stats.py` — `/v1/sites/{site}/stats/devices/{mac}` and `/v1/sites/{site}/stats/clients/{mac}` now call `stats_manager.get_X_stats` directly instead of `*_details`, so route output matches the new TIMESERIES serializer kind.

### 5. Tests

- `test_dispatch_overrides_redirect_compose_tools_to_mutation` — every override entry resolves correctly (asserts on the full DISPATCH_OVERRIDES table).
- `test_dispatch_overrides_specific_targets` — spot-check representative samples across all three override categories.
- Updated stats serializer/route tests for the TIMESERIES shape (was DETAIL).

## Override breakdown (45 tools)

- **Network (26)**: 8 client mutations (`block`/`unblock`/`rename`/`force_reconnect`/`authorize_guest`/`unauthorize_guest`/`set_ip_settings`/`forget`), `list_clients`, 6 network/wlan/ap-group mutations, 3 firewall mutations (incl. `create_simple_firewall_policy` multi-manager compose), 4 toggle tools (`port_forward`/`qos_rule_enabled`/`oon_policy`/`traffic_route`), `update_device_radio`, 3 stats compose tools.
- **Protect (11)**: `alarm_arm`/`alarm_disarm`, 6 camera tools (`ptz_*`/`reboot`/`toggle_recording`/`update_camera_settings`), 2 device tools (`update_chime`/`update_light`), `acknowledge_event`.
- **Access (8)**: 2 credential mutations, `reboot_device`, 2 door mutations, `update_policy`, 2 visitor mutations.

## Test deltas

| Package | Before | After | Δ |
|---|---|---|---|
| unifi-core | 69 | 69 | 0 |
| unifi-mcp-shared | 205 | 205 | 0 |
| unifi-network-mcp | 630 | 630 | 0 |
| unifi-protect-mcp | 336 | 336 | 0 |
| unifi-access-mcp | 157 | 157 | 0 |
| unifi-api | 388 | **390** | +2 (override coverage tests) |

## After this PR merges

The Phase 2 AST-based dispatch table now resolves correctly for **every** mutation tool across network/protect/access. The Phase 5A "AST introspection fix" item from the original Phase 5A scope is no longer needed — the static-override pattern is the architecturally cleaner solution.

## Test plan

- [x] All 6 package suites pass
- [x] New dispatch_overrides coverage test asserts all 45 entries resolve
- [x] Existing `test_serializer_coverage` Phase 4A gate stays green
- [x] Live smoke recommendable but not strictly required (no behavior change at the controller layer; only dispatch routing within unifi-api)

### DIFF
--- a/apps/api/src/unifi_api/routes/resources/network/stats.py
+++ b/apps/api/src/unifi_api/routes/resources/network/stats.py
@@ -1,20 +1,16 @@
 """GET /v1/sites/{site_id}/stats/* — network stats endpoints.
 
-Phase 5A PR2 Cluster 6. Stats endpoints span TIMESERIES and DETAIL kinds.
+Phase 5A PR2 Cluster 6, amended PR4 of mgr-existence-refactor.
 
-Phase 4A registered 5 stats tools as TIMESERIES (dashboard, network, gateway,
-site_dpi_traffic, client_dpi_traffic) and 3 as DETAIL (device_stats,
-client_stats, dpi_stats) because Phase 2's AST captured the lookup method,
-not the stats fetch method. The parallel manager-refactor will eventually
-re-register the 3 DETAIL ones as TIMESERIES; the dual-kind handler below
-makes Phase 5A robust to whichever state is current at request time.
-
-For the 3 currently-DETAIL stats tools, we call the manager method that the
-AST-derived dispatch table currently routes to (verified at module import
-time): device_manager.get_device_details, client_manager.get_client_details,
-stats_manager.get_dpi_stats. When the parallel refactor lands, both the
-serializer kind and the dispatch table will update together — the route
-inherits both via the registry/dispatch.
+Stats endpoints span TIMESERIES and DETAIL kinds. After PR4:
+- TIMESERIES (7): dashboard, network, gateway, site_dpi_traffic,
+  client_dpi_traffic, **device_stats** (newly), **client_stats** (newly).
+  ``device_stats`` and ``client_stats`` now call ``stats_manager.get_X_stats``
+  directly (was ``device_manager.get_device_details`` /
+  ``client_manager.get_client_details`` pre-PR4); the dispatch override and
+  serializer kind both flipped together.
+- DETAIL (1): dpi_stats — shape is ``{applications: [], categories: []}``,
+  not a per-point timeseries.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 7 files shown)


sirkirby:
## Live smoke validation — passed (refactor-clean, dispatch override firing)

Ran \`scripts/live_smoke.py --phase api-actions\` against the production controller. This phase spins up unifi-api locally with the PR4 changes, registers all 3 real controllers, and exercises the dispatch path through \`POST /v1/actions/{tool_name}\`.

**Result: 6/6 passed, 0 regressions, 0 preexisting failures.**

| Tool | HTTP | Result | Live count | Latency |
|---|---|---|---|---|
| \`unifi_list_clients\` | 200 | success=true | **173** clients | 682ms |
| \`unifi_list_devices\` | 200 | success=true | 41 devices | 440ms |
| \`protect_list_cameras\` | 200 | success=true | 13 cameras | 733ms |
| \`protect_list_lights\` | 200 | success=true | 1 light | 5ms |
| \`access_list_doors\` | 200 | success=true | 1 door | 907ms |
| \`access_list_users\` | 200 | success=true | 5 users | 89ms |

### Override-firing evidence

\`unifi_list_clients\` returning **173** clients is the smoking gun that the override is actually being applied at runtime (and not just registered):

- Pre-PR4 the AST captured \`client_manager.get_all_clients\` (the \`include_offline=True\` branch — would return ~434 entries on this controller).
- Post-PR4 the \`DISPATCH_OVERRIDES\` table pins \`unifi_list_clients\` to \`client_manager.get_clients\` (online-only).
- Live count of 173 matches the online-clients-only path.

The PR1 smoke run for this same controller showed \`live_count=434\` for \`unifi_list_clients\` — confirming the override changed the dispatch target end-to-end through the action endpoint.

### What this validates

1. Module imports cleanly; \`dispatch_overrides.py\` doesn't break unifi-api startup.
2. \`build_dispatch_table()\` runs successfully with the overrides applied (server stays up, 6/6 calls succeed).
3. Dispatch through \`/v1/actions/\` routes to the override-pinned manager method, not the AST-captured one.
4. All 3 controllers (network/protect/access) register and respond after the cross-cutting \`unifi-core\` exception changes from PR1–PR3 land alongside.

### Why this isn't a full mutation-tool exercise

The api-actions harness phase exercises a fixed baseline of 6 read-only tools. Mutation tools (the bulk of the 45 overrides) would need an extended manual smoke session against a non-production controller. The unit-test \`test_dispatch_overrides_redirect_compose_tools_to_mutation\` asserts every override entry resolves correctly in the table; combined with the live evidence that the table is actually consulted by the dispatch path, mutation-tool dispatch correctness is high-confidence.

Reports archived at \`/tmp/smoke-pr4-api/\`.

Issue #200, opened 2026-05-06
MCP tools not registered in Claude Code session when installed via plugin


patrick-motard:
## Problem

When the `unifi-network` plugin is installed via the Claude Code plugin system, the MCP tools (`unifi_execute`, `unifi_tool_index`, `unifi_batch`, etc.) are not available as callable tools in the session. The skill loads successfully and references these tools, but they are never registered with Claude Code's tool system.

## Steps to reproduce

1. Install the `unifi-network` plugin via the Claude Code plugin manager
2. Start a new Claude Code session
3. Invoke the `unifi-network:unifi-network` skill (e.g. ask "list connected devices in unifi")
4. The skill loads and instructs the agent to call `unifi_execute` or `unifi_tool_index`
5. Those tools don't exist - attempting to call them fails with `InputValidationError` or no tool found

## Expected behavior

After installing the plugin, the MCP server should be automatically registered in Claude Code's MCP server config, and the UniFi tools should be available in the session.

## Actual behavior

The plugin installs the skill files under `~/.claude/plugins/cache/unifi-plugins/unifi-network/` but does NOT add the MCP server to `~/.claude/settings.json` or `~/.claude/settings.local.json`. The `mcpServers` block in `plugin.json` is never applied.

The user is left with skills that reference tools that don't exist, with no indication of why or how to fix it.

## Environment

- Claude Code plugin cache: `~/.claude/plugins/cache/unifi-plugins/unifi-network/0.14.15/`
- Plugin `mcpServers` config present in `.claude-plugin/plugin.json` but not applied to settings
- `uvx unifi-network-mcp==0.14.15` is the expected MCP server command

## Suggested fix

The plugin installation process should automatically register the `mcpServers` block from `plugin.json` into the user's Claude Code settings, or the setup skill (`/unifi-network:setup`) should do it automatically with no manual steps required.

The `/unifi-network:setup` skill exists but it's not clear it needs to be run, and it's not triggered automatically after install.

patrick-motard:
Additional context: I ran `/unifi-network:setup` in the session prior to hitting this bug. The setup completed, wrote credentials to `~/.claude/settings.local.json`, and said to restart Claude Code. After restarting, the MCP tools still weren't available.

The setup script itself also hit a bug - `set-env.sh` uses `declare -A` (bash associative arrays) which requires bash 4+, but macOS ships with bash 3.5. The script fails with:

```
declare: -A: invalid option
[...]

PR #202, opened 2026-05-06
fix(transport): HTTP failure no longer cancels stdio (#200)

sirkirby:
## Summary

When \`UNIFI_MCP_HTTP_FORCE=true\` and the HTTP bind fails (e.g. something else is on port 3000), the entire MCP server becomes unreachable — including the stdio transport that Claude Code actually talks over. Caught by @patrick-motard with a clean test matrix on #200:

| HTTP_ENABLED | HTTP_FORCE | PORT | Result |
|---|---|---|---|
| false | false | 3000 | works |
| true | false | 3000 | works |
| true | true | 3000 | **fails (the bug)** |
| true | true | 3001 | works |

**Root cause**: \`run_transports\` used \`asyncio.wait(FIRST_COMPLETED)\`. When uvicorn's bind fails it raises \`SystemExit\`; \`run_http\` catches it and returns normally. From asyncio.wait's perspective HTTP "finished first", so stdio was cancelled as the pending task — taking down the only transport Claude Code uses.

### 1. plugin.json default (immediate user fix)

Flip \`UNIFI_MCP_HTTP_FORCE\` from \`"true"\` → \`"false"\` in all three plugin manifests. Claude Code drives the server over stdio (uvx-launched, not PID 1), so HTTP is a side channel at best — FORCE=true was buying users a port-conflict footgun. With FORCE=false, the existing PID-1 check in \`resolve_http_config\` disables HTTP cleanly in plugin contexts; nothing else changes.

### 2. shared/transport.py refactor (defensive fix)

Restructure \`run_transports\` so stdio is the primary control flow and HTTP runs as a cancellable background task:

- HTTP failure is contained: stdio keeps serving when HTTP dies fast.
- Stdio exit cancels HTTP via the \`finally\` clause, preserving the original "no orphan listener" behaviour.
- Adds two regression tests pinning both behaviours (\`test_http_systemexit_does_not_cancel_stdio\`, \`test_stdio_exit_cancels_http\`).

This protects anyone who deliberately sets \`FORCE=true\` (e.g. for local debugging) from the same crash class.

## Test plan

- [x] \`uv run pytest packages/unifi-mcp-shared/tests/test_transport.py\` — 9/9 (7 existing + 2 new regression tests)
- [x] \`uv run pytest packages/unifi-mcp-shared\` — 207/207
- [x] \`uv run pytest apps/network/tests\` — 651/651
- [x] \`uv run pytest apps/protect/tests\` — 336/336
- [x] \`uv run pytest apps/access/tests\` — 157/157
- [x] \`claude plugin validate\` passes for all three plugin manifests
- [x] \`/bin/bash scripts/smoke-plugin-setup.sh\` — 25/25
- [ ] **Reporter re-test:** ask @patrick-motard to re-verify with the merged fix

Refs #200. Will require a release wave once merged: \`shared\` (real code change), then \`network\`, \`protect\`, \`access\` (plugin manifests + transitive shared bump).

### DIFF
--- a/packages/unifi-mcp-shared/src/unifi_mcp_shared/transport.py
+++ b/packages/unifi-mcp-shared/src/unifi_mcp_shared/transport.py
@@ -80,15 +80,21 @@ async def run_transports(
     logger: logging.Logger,
     protocol_version: str = "v1",
 ) -> None:
-    """Run stdio (always) and optionally HTTP, with coupled lifecycles.
+    """Run stdio (always) and optionally HTTP, with stdio as the primary.
 
-    When both transports are active, uses ``asyncio.wait(FIRST_COMPLETED)``
-    so that when either transport exits the other is cancelled.  This ensures
-    the process terminates cleanly when the stdio client disconnects instead
-    of being kept alive by the HTTP server.
+    Stdio is the primary transport — Claude Code (and any other stdio MCP
+    client) talks to the server over it.  HTTP, when enabled, is a side
+    channel that runs as a background task: if HTTP fails (port conflict,
+    bind error, etc.) stdio keeps serving.  When stdio exits, HTTP is
+    cancelled so the process terminates cleanly.
+
+    The previous implementation used ``asyncio.wait(FIRST_COMPLETED)``, which
+    caused HTTP-died-fast to look like "transport finished" and cancelled
+    stdio — taking down the only transport Claude Code actually uses.  See
+    issue #200.
 
     ``SystemExit`` raised by uvicorn on port-bind failures is caught inside
-    ``run_http`` so it does not propagate and kill the stdio transport.
+    ``run_http`` so it does not propagate.
     """
     # Future: when MCP SDK v2 changes the transport API, branch here:
     # if protocol_version == "v2":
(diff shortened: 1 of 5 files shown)


sirkirby:
Currently running as much live smoke testing as I can against my own infra, will merge once it passes.

sirkirby:
Verification done before this is merge-ready:

**Unit tests** — 9/9 in `test_transport.py` (7 existing + 2 new regression tests pinning both behaviours: HTTP failure must not cancel stdio; stdio exit must cancel HTTP). Full suite green: 1351/1351 across shared + network + protect + access.

**Live smoke** against a real controller (readonly phases): network 38/38, protect all-ok with one streaming tool skipped by design, access mostly ok with a couple of unrelated controller-side permission errors that propagated cleanly through stdio (which itself confirms the transport refactor isn't dropping responses).

**End-to-end port-conflict repro** — built wheels for `unifi-mcp-shared` and `unifi-network-mcp` from this branch, bound port 3000 with a local TCP listener to simulate the Docker conflict from #200, launched the wheel via `uvx` with the exact bug-triggering env (`HTTP_ENABLED=true`, `HTTP_FORCE=true`, `PORT=3000`), and sent an MCP `initialize` over stdin. Result: HTTP fails to bind cleanly (`address already in use`), HTTP task exits, stdio keeps serving, `initialize` gets a valid response — and crucially, no 'Cancelling stdio transport' log line, which was the racy code path that took stdio down.

For contrast, ran the same probe with the published shared package from PyPI (the old code) and confirmed that 'Cancelling stdio transport' line still fires — whether stdio responds before it gets cancelled is timing-dependent, which is exactly why the bug appeared intermittent / environment-specific. The fix removes the race entirely rather than papering over it.

PR #288, opened 2026-05-19
fix(firewall): allow IP+OBJECT validation with ip_group_id (empty ips)

rmaher001:
## Summary

The `_validate_zone_targeting` validator unconditionally requires `ips` to be non-empty when `matching_target == "IP"`, but the V2 firewall API legitimately supports policies that reference a reusable IP group object via `ip_group_id` with an empty `ips` array:

```json
{
  "matching_target": "IP",
  "matching_target_type": "OBJECT",
  "ip_group_id": "<group-object-id>",
  "ips": []
}
```

This is the standard shape for any IP-group-based policy on modern firmware (UDM-SE 8.4.x+), but the validator rejected it at create-time, blocking `unifi_create_firewall_policy` for any rule that references an IP group object — even though equivalent policies exist on the controller already.

## Fix

Under `matching_target == "IP"`:
- If `matching_target_type == "OBJECT"`: require `ip_group_id` (allow empty `ips`)
- Otherwise: require non-empty `ips`

Also widened the missing-`matching_target_type` error hint to include "SPECIFIC" or "OBJECT".

## Tests

Two new cases in `apps/network/tests/unit/test_firewall_tools.py::TestCreateZoneTargetingValidation`:
- `test_ip_object_targeting_with_ip_group_id_passes_validation` — verifies the OBJECT + `ip_group_id` shape passes the validator and creates the policy
- `test_ip_object_targeting_without_ip_group_id_fails` — verifies the validator still rejects OBJECT without `ip_group_id`

### DIFF
--- a/apps/network/src/unifi_network_mcp/tools/firewall.py
+++ b/apps/network/src/unifi_network_mcp/tools/firewall.py
@@ -281,14 +281,18 @@ def _validate_zone_targeting(validated_data: Dict[str, Any]) -> str | None:
             continue
         target = ep.get("matching_target")
         if target in ("IP", "NETWORK") and not ep.get("matching_target_type"):
-            expected = "SPECIFIC" if target == "IP" else "OBJECT"
-            return "%s.matching_target_type is required when matching_target is '%s'. Use '%s'." % (
+            expected = "'SPECIFIC' or 'OBJECT'" if target == "IP" else "'OBJECT'"
+            return "%s.matching_target_type is required when matching_target is '%s'. Use %s." % (
                 direction,
                 target,
                 expected,
             )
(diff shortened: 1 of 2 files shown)


sirkirby (approved):
Thanks @rmaher001 — clean, well-scoped fix and the two test cases (positive + negative for OBJECT/ip_group_id) are exactly right. The V2 firewall API does support `matching_target=IP` with `matching_target_type=OBJECT` and `ip_group_id` (empty `ips`) for IP-group-referenced policies, and the validator should not block that shape.

Pushed a single `ruff format` line-wrap to your branch to clear the format check (no behavior change). Merging.

PR #295, opened 2026-05-20
feat(protect): server-side camera filter + opt-in event metadata passthrough

rmaher001:
## Summary

Two related, universally-beneficial fixes to the Protect MCP's event-listing path. Both are pure-additive and backwards-compatible.

### 1. Server-side camera filter

`EventManager.list_events()` and `list_smart_detections()` previously fetched all-camera events from UniFi and filtered by `camera_id` client-side. When the caller asked for a single camera, this wasted roughly half of the per-response event budget on events from other cameras.

This PR adds a code path that, when `camera_id` is set, calls
`client.api_request_list("events", params={..., "cameras": [camera_id]})`
directly. UniFi accepts the `cameras[]=<id>` query parameter and returns only matching events — empirically roughly doubling the useful event count per call on a multi-camera site.

### 2. Opt-in event metadata passthrough

uiprotect's `Event` model silently drops several fields UniFi returns, including `linesStatus`, `linesSettings`, `zonesStatus`, `weather`, `detectedAreas`, `loiterStatus`, and `tamperStatus`. These are useful for line-crossing direction analysis, zone-status checks, and other downstream uses.

This PR adds a `metadata_fields: list[str]` parameter to all four event-listing MCP tools:

- `protect_list_events`
- `protect_list_smart_detections`
- `protect_get_event`
- `protect_recent_events`

Default empty list = no metadata in the response (preserves existing token budget for current callers). Pass top-level key names to opt in to specific fields, or `["*"]` to include the full metadata dict.

The selector shape is forward-compatible with future dotted-path nested selection — purely an internal parser upgrade, no API change.

### Implementation notes

- A new `EventManager._raw_event_to_dict()` helper mirrors `_event_to_dict()`'s output shape but builds from a raw event dict (from `api_request_list`/`api_request_obj`) instead of a parsed `uiprotect.Event`. This is what enables surfacing the dropped metadata fields.
- `EventManager.get_event()` gains the same parameter and, when set, calls `api_request_obj(f"events/{event_id}")` — the same single-event endpoint uiprotect's own `get_event()` uses under the hood (`get_device → get_device_raw → api_request_obj("events/<id>")`).
- The websocket-buffer-backed `protect_recent_events` filters metadata in the tool layer (shallow copy of each buffer event so the buffer is not mutated).
- The `Event` and `SmartDetection` pydantic models gain an optional `metadata: Optional[dict[str, Any]]` field that `from_controller()` passes through.

## Tests

- 7 manager tests in `packages/unifi-core/tests/protect/managers/test_event_manager_metadata.py` — raw-API routing, server-side camera filter, metadata filter semantics (`[]` default, named keys, `["*"]`, unknown key omission, metadata_fields-only routing without camera_id).
- 4 model tests in `packages/unifi-core/tests/protect/models/test_event_metadata_passthrough.py` — `from_controller`/`smart_detection_from_controller` passthrough (present + absent for both).
- 9 tool-level tests in `apps/protect/tests/unit/test_event_tools_metadata.py` — parameter pass-through for all four tools + recent_events buffer-filter semantics (no-metadata default, named-fields, `["*"]`, no buffer mutation).
- 2 pre-existing camera_id tests in `apps/protect/tests/unit/test_event_manager.py` updated to mock the new raw-API path.

Full Protect test suite: 642 passing (263 unifi-core + 379 apps/protect).


sirkirby (approved):
looks good. I also ran a live smoke test on my infra, everything check out.

PR #301, opened 2026-05-25
Add firewall policy ordering tools

bmoor:
## Summary

Add explicit firewall policy ordering support for UniFi zone-based firewall rules and wire the ordering read surface into the API.

### Changes

- Add `unifi_get_firewall_policy_ordering`
- Add `unifi_reorder_firewall_policies`
- Use the UniFi Network integration API endpoint:
  - `GET /v1/sites/{siteId}/firewall/policies/ordering`
  - `PUT /v1/sites/{siteId}/firewall/policies/ordering`
- Require API-key auth for ordering calls via `X-API-Key`
- Translate normal V2 firewall zone IDs from `unifi_list_firewall_zones` to integration API zone UUIDs
- Add REST and GraphQL API read surfaces for firewall policy ordering
- Require explicit confirm for API action dispatch of firewall reorder
- Harden reorder payload validation:
  - require `beforeSystemDefined` and `afterSystemDefined` arrays
  - reject non-string or empty policy IDs
  - reject duplicate policy IDs
  - reject payloads that add/drop policy IDs instead of only moving existing IDs
- Validate reorder payloads against a fresh current controller ordering before PUT, not cached ordering
- Invalidate firewall ordering and policy caches after reorder
- Harden integration API error reporting for non-JSON or empty error bodies
- Fix V2 firewall create defaults required by the controller:
  - default `schedule` to `{"mode":"ALWAYS"}`
  - default `create_allow_respond=false` for BLOCK/REJECT
- Document the new ordering behavior in the firewall skill reference

## Why

Zone-based firewall rule order is user-defined per source/destination zone pair. The controller may expose an `index` on policies, but actual reorder operations use the dedicated ordering endpoint and the `orderedFirewallPolicyIds` payload.

This avoids brittle `index` writes and matches the official UniFi Network API surface. The confirmed reorder path now re-fetches current ordering immediately before mutation so a stale cache cannot accidentally validate a payload that would drop or add policy IDs.

## Testing

- `ruff check .`
- `pytest apps/network/tests -q`
- `pytest packages/unifi-core/tests -q`
- `pytest apps/api/tests/test_actions_dispatcher.py -q`
- WSL: `make pre-commit`

Additional live smoke against a UniFi controller before the API-surface follow-up:

- discovered `unifi_get_firewall_policy_ordering`
- listed firewall zones
- read ordering for `Internal -> Internal`
- verified reorder preview shows both current and proposed ordering without executing a mutation

Notes:

- WSL `make pre-commit` passed.
- Offline MCP metadata smoke checks log expected local connection failures against dummy controller credentials; they completed successfully.
- Existing Worker `npm audit` warning remains: 6 vulnerabilities in Worker dev tooling. Not touched by this firewall PR.

### DIFF
--- a/.agents/skills/add-integration-api-tool/SKILL.md
+++ b/.agents/skills/add-integration-api-tool/SKILL.md
@@ -1,0 +1,205 @@
+---
+name: add-integration-api-tool
+description: >
+  Use this skill when adding any tool that talks to the UniFi public Integration API
+  (X-API-Key auth, /proxy/network/integration/...). The integration API uses a
+  different ID namespace and field schema from the V2 controller API, so these tools
+  form their own "families" with scoped IDs. Covers when to start a new family vs
+  join an existing one, how to scope IDs in tool descriptions, the narrow conditions
+  under which silent ID translation is allowed (rare), how to wire auth, and how to
+  bridge between ID spaces (explicit named bridging tools only). Applies whether
+  you are adding a read tool, a mutation tool, or a paired family of both.
+user-invocable: true
+allowed-tools: Read, Edit, Write, Bash, Grep, Glob
+---
+
+# Adding a Tool That Uses the UniFi Integration API
+
+The UniFi controller exposes two API surfaces. They are not interchangeable.
+Skipping this distinction creates silent footguns where IDs from one API are
+accepted by tools backed by the other and silently fail — or worse, succeed
+against the wrong resource.
+
+## The two surfaces
+
+| | V2 controller API | Integration API |
+|---|---|---|
+| Auth | session cookie | `X-API-Key` header |
+| Path prefix | `/proxy/network/api/...`, `/proxy/network/v2/api/...` | `/proxy/network/integration/v1/...` |
+| IDs | Mongo ObjectIDs (`674f0...`) | UUIDs (`0998be5a-...`) |
+| Field naming | `snake_case` | `camelCase` |
+| Action shape | `"action": "ALLOW"` + flat fields | nested `{"action": {"type": "ALLOW", "allowReturnTraffic": true}}` |
+| `ip_version` | `"BOTH"` | `"ipProtocolScope": {"ipVersion": "IPV4_AND_IPV6"}` |
+| Coverage | everything the web UI uses | curated, expanding subset |
+| Membership | full controller state | filtered (system policies often excluded) |
+
+These differences are not cosmetic. The same logical resource appears in V2
+with a Mongo ID and in Integration with a UUID, with no native bridge field
+connecting them.
+
+## Step 1: Identify the family
+
+A *tool family* is a set of tools whose returned IDs and accepted IDs are
+interchangeable.
+
+- Reading + writing to the same API endpoint group? Same family.
+- Reading from V2 and writing to V2? Same family (the existing firewall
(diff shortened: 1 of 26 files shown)


sirkirby (changes requested):
Make sure to run `make pre-commit`, which should help you get the CI checks passing.

sirkirby:
thanks for getting the updates. I'm in the process of reviewing, but your pull request brings in a bigger question that I have to really consider how we deal with the v1 public API versus the private APIs, which most of the tools support. It turns out the data not only uses different IDs but are often different shapes, so I'm just really thinking this through before getting back to you with a concrete solution here. Either way, we will support the ordering. I'm experimenting with the right pattern to move us forward as these API's contiue to diverge...keeping the maintenance buren to a minimum.

bmoor:
Thanks for the thoughtful context. That makes sense to me — this PR definitely surfaced the public-v1-vs-private-API shape/ID split more clearly than I expected.

I’m happy to adapt this PR to whatever pattern you decide is best for the project.

sirkirby:
Thanks for this PR @bmoor — pushed a follow-up commit (`c5724c9`) to your branch that:

1. **Tightens the tool descriptions** for `unifi_get_firewall_policy_ordering` and `unifi_reorder_firewall_policies` to explicitly scope the integration-API UUIDs to the ordering tool family. Same scoping clause added to the GraphQL type/query and REST route descriptions, and regenerated `openapi.json` + the reference markdown.
2. **Codifies the dual-API pattern.** During review we verified against a live controller that the V2 controller API and the public integration API have disjoint ID namespaces (Mongo vs UUID), different field shapes (snake_case vs camelCase, flat action vs nested action object), and disjoint coverage (180 V2 policies vs 25 integration policies). To prevent agents from mixing IDs across the two surfaces in future tools, the project now has:
   - A new `### API Surface Boundaries` section in `AGENTS.md` defining the tool-family pattern
   - A new `.agents/skills/add-integration-api-tool/` skill documenting the full procedure
   - Your ordering tools serve as the canonical reference implementation
3. **Adds two tests** previously flagged as coverage gaps:
   - REJECT action auto-injecting `create_allow_respond=false` (BLOCK was tested; REJECT wasn't)
   - Payload missing `afterSystemDefined` entirely is rejected before the API call

No behavior change to your underlying logic — solid work on the actual feature. The architecture pieces are the project's responsibility to formalize, not yours to carry; we wanted to land them with your PR since it's the first integration-API surface in the codebase and the right reference point.

Approving and dropping the previous "request changes."

sirkirby (approved):
Approving after the architectural follow-up commit (c5724c9) bundling the AGENTS.md rule, the new skill, and the description scoping. Quality gates verified: make lint clean, all 800 pre-commit tests pass, live smoke confirms list_firewall_policies + get_firewall_policy_ordering + validation guards work against the controller (Internal->External returns 23 ordered policy UUIDs, validation correctly rejects duplicates and dropped-id payloads with helpful error messages).

bmoor:
Thanks a lot for all the work you put into this project.

I really appreciate the time and care you put into the architecture, documentation, reviews, and maintenance. The project has been genuinely useful for me, and it is clear that a lot of thoughtful work has gone into it.

Thanks for building and maintaining it.

PR #308, opened 2026-05-27
feat(network): add search/limit/field & section selection to read-only tools to reduce response sizes

rmaher001:
## Summary

The network read tools (`list_clients`, `list_devices`/`get_device_details`, `list_networks`/`get_network_details`, `list_wlans`, `list_firewall_policies`) can return very large controller payloads — full radio/port/network tables, every client, all raw fields. For MCP/LLM callers that's a lot of tokens for data they often don't need.

This PR adds **opt-in** controls to trim those responses, using one consistent pattern across the tools:

- **`search`** — case-insensitive substring filter (name/hostname/IP/MAC, or SSID/VLAN as appropriate).
- **`limit`** — cap rows, with `total_count` / `returned_count` in the response so truncation is explicit.
- **`fields`** — comma-separated subset of fields per row (where applicable).
- **`include`** — comma-separated section selector on the detail tools (`get_device_details`, `get_network_details`, `get_client_details`), with `summary=false` as a full-raw-data escape hatch. For `get_client_details`, `summary` defaults to `false`, so its default response is unchanged from #300 (the full raw object); trimming is strictly opt-in.
- **Compressed list summaries** for `list_devices` (`include_details=true`): AP radio/vap counts, switch port up/PoE counts, gateway WAN/stat flattening — instead of the full nested tables.

It also improves `get_device_details` port data with **LLDP neighbor resolution**: each `port_summary` entry distinguishes `last_seen_mac` (last MAC seen on the port — may be wireless client traffic, unreliable) from `lldp_neighbor` (the actual directly-connected infrastructure device, from the LLDP table).

These tools have been running in my deployment for ~a month; this PR ports them onto current `main` and adds the unit-test coverage they previously lacked.

## Points worth your attention (please weigh in)

1. **Could be split.** It's one cohesive theme, but if you'd rather review in pieces, the `get_device_details` LLDP-accuracy improvement is cleanly separable from the filtering work.

## Note on the list serializers

`list_networks` and `list_firewall_policies` still go through `network_from_controller` / `fw_from_controller`; they source values from the typed model and then narrow to a curated lean subset (the full model dump is the token bloat this PR targets), with `list_networks` also backing the optional `fields` selector. One small consequence: values reflect the model's coercion (e.g. `Network.vlan` is typed `str`, so the listed `vlan` is a string).

## Backward compatibility

Fully additive — all new parameters are optional and default to the existing behavior. `list_clients` keeps its default `limit` of 100; `list_devices` returns all devices unless `limit` is passed; `get_client_details` defaults to the full raw object. The one shape change is that list responses replace the `count` field with `total_count`/`returned_count`/`limit` (plus echoed filter values).

## Testing

- New: `test_clients_filtering.py`, `test_devices_filtering.py`, `test_network_filtering.py`; updated `test_firewall_tools.py` for the new response shape.
- Full `apps/network` suite passes (689 tests); `ruff format --check` and `ruff check` clean.
- `tools_manifest.json` regenerated.

### DIFF
--- a/apps/network/src/unifi_network_mcp/tools/clients.py
+++ b/apps/network/src/unifi_network_mcp/tools/clients.py
@@ -32,6 +32,28 @@ from unifi_network_mcp.runtime import client_manager, server
 
 logger = logging.getLogger(__name__)
 
+_ACTIVE_CONNECTION_KEYS: tuple[str, ...] = (
+    "_uptime_by_uap",
+    "_uptime_by_usw",
+    "_uptime_by_ugw",
+    "uptime",
+)
+
+
+def _is_online(raw: dict) -> bool:
+    """Mirror of unifi_core.network.models.clients._is_online — derives
+    online status from a controller record, falling back to
+    active-connection indicators when ``is_online`` is omitted by the
+    controller firmware.
+    """
+    if raw.get("is_online") is True:
+        return True
+    for key in _ACTIVE_CONNECTION_KEYS:
+        v = raw.get(key)
+        if isinstance(v, (int, float)) and v > 0:
+            return True
+    return False
+
 
 @server.tool(
     name="unifi_lookup_by_ip",
@@ -71,7 +93,8 @@ async def lookup_by_ip(
         "Returns connected clients with mac, name, hostname, ip, status (online/offline), "
         "connection type (wired/wireless), and for wireless clients: ssid, signal_dbm, "
         "channel, radio. Filter by filter_type (all/wired/wireless), set "
-        "include_offline=true for historical clients. For a single client's full raw "
+        "include_offline=true for historical clients. Use search to filter by "
+        "name/hostname/IP/MAC (case-insensitive). For a single client's full raw "
         "object, use unifi_get_client_details. For IP-to-client lookup, use "
         "unifi_lookup_by_ip."
     ),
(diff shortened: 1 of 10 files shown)


sirkirby (changes requested):
Thanks Richard — this is a meaty PR and the direction is exactly right. Token bloat on these read tools has been a real pain point. A few things to work through before merge:

## Hard blocker

**`get_device_details` default should preserve the full raw object.** You nailed the symmetry with `get_client_details` (defaulting `summary=False` to preserve the #300 contract) — `get_device_details` needs the same treatment. Its current default (`summary=True, include="basic,ports"`) silently narrows the response for every existing caller, and the pre-PR tool description literally promises "the full raw device object." Flip the default to `summary=False` and we're good.

## Design questions

The "fully additive" framing in the PR body is mostly accurate, but two spots quietly narrow existing response shapes:

1. **`list_devices` with `include_details=true`** — drops `radio_table`, `vap_table`, `port_table`, `network_table`, `system_stats`, `speedtest_status`, `poe_info` in favor of the compressed summaries. Love the compressed shape as a *default*, but want to keep an escape hatch for the full tables. Could the compression go behind a `summary=True` (default) with `summary=False` returning the legacy include_details payload?

2. **`list_firewall_policies`** — narrowed from full `fw_from_controller().model_dump()` to a 6-key curated entry; drops `protocol`, `schedule`, `logging`, `ip_version`, `index`. Same question — gate behind `summary`?

3. **`count` → `total_count`/`returned_count`/`limit`** — you flagged this in the PR body, so we're aligned, but it's a one-line back-compat win if we also emit `"count": returned_count`. Up to you whether it's worth it.

## Should fix

- **`_is_online` in `tools/clients.py:35–55` is byte-identical to `unifi_core.network.models.clients._is_online`.** Let's import rather than mirror — easy drift target otherwise.
- **Unknown `fields`/`include` tokens silently drop.** A typo like `include="basix,ports"` runs as if no section requested. Surfacing unknown tokens in the response (or just an `unknown_sections` key) would help LLM callers self-correct.

## Test coverage gaps

Nothing dramatic, but worth pinning while we're here:
- `include="all"` selector — documented across the section-based tools, not exercised
- `list_firewall_policies` filter composition (`action` + `enabled_only` + `search` together)
- Zero-controller-items case on the list tools (`total_count=0, returned_count=0`)
- Explicit `summary=False` raw-passthrough test for `get_client_details` (devices/network have the equivalent)
- LLDP: a port with `last_seen_mac` set but no LLDP entry (the "wireless client traffic" cell)

## Nit

`test_*_filtering.py` uses `patch.start()` / `try-finally`; the rest of the suite uses `with patch(...) as m:` — small consistency thing.

LLDP work in `get_device_details` is great, by the way — the `last_seen_mac` vs `lldp_neighbor` split is exactly the right call.

sirkirby (approved):
Nicely done — every item addressed, plus the symmetric `summary=False` flip on `get_network_details` and the `unknown_fields`/`unknown_sections` echo are exactly the kinds of touches that make this PR easy to live with. Merging.

PR #310, opened 2026-05-28
feat(protect): add alarm rule CRUD (list/get/create/update/delete) and GraphQL exposure

rmaher001:
## Summary

Extend the Protect Alarm Manager wrapper with read + write coverage of the private `/proxy/protect/api/automations` endpoint. The existing arm/disarm tools (Protect 6.1+ Alarm Manager) get five new siblings that round out programmatic control of alarm rules (automations):

| Tool | Endpoint | Notes |
|---|---|---|
| `protect_alarm_list_rules` | `GET automations` | Flat list of all rules |
| `protect_alarm_get_rule` | `GET automations` + filter | Per-rule fetch (see below) |
| `protect_alarm_create_rule` | `POST automations` | Full-body create |
| `protect_alarm_update_rule` | `PATCH automations/{id}` | Full-body PATCH (read-modify-write) |
| `protect_alarm_delete_rule` | `DELETE automations/{id}` | Idempotent delete |

The five tools are also exposed through GraphQL / the REST action surface (see "GraphQL / REST exposure" below).

## Motivation

The arm/disarm pair is useful but only exposes the system-wide state. Real automation needs to inspect and tune the rule definitions themselves -- the conditions that trigger an alarm, the actions taken, and the cooldown -- via API. With these five tools the full alarm-rule lifecycle is reachable from MCP.

## Endpoint behavior verified against a live Protect 7.x controller

A few quirks shaped the implementation:

- **`GET automations/{id}` is not routed.** The controller returns 404 for per-rule path requests. The per-rule fetch is `GET automations` then filter by id. `get_rule` does that and raises `UniFiNotFoundError` when no row matches.
- **`DELETE automations/{id}` returns an empty body on success.** Using the default `api_request` triggers a spurious "Could not decode JSON" error. `delete_rule` uses `api_request_raw` (no JSON decode) to avoid it.
- **`POST automations` is strict camelCase.** The controller's read response is camelCase (`isCreatedBySystem`, `historyConditions`, action metadata `useThumbnail`), and POST rejects snake_case with `400 Failed to parse 'request-body'`. Since this codebase normalizes read responses to snake_case via the existing `rule_from_controller` pattern (consistent with the rest of the protect models), `rule_to_controller` is the inverse: it translates the small set of case-mismatched fields back to camelCase before POST/PATCH and passes every other key through untouched. Both casings are accepted; if the caller provides both, the camelCase value wins. This means the natural read-modify-write clone flow (`get_rule` → mutate → `create_rule`) works directly.
- **Empty `actions` list bricks the Protect UI.** The controller accepts `actions: []` (or no `actions` key) and returns a normal-looking rule on POST. But the resulting rule cannot be opened in the Protect UI -- it shows the generic "We're Unable to Complete Your Request" error and the rule becomes only deletable via API. The input validators on `AlarmCreateRuleInput` and `AlarmUpdateRuleInput` reject this client-side before any controller call.

## Implementation

- **Lossy on deeply-nested condition bodies.** `AlarmRuleCondition.condition` is a free-form `Dict[str, Any]` rather than enumerating every Protect condition source/type combination. Protect adds new detection sources over time (`license_plate_known`, `smartDetectLine`, `Vehicle Description` specifics, etc.) and a strict schema would block passthrough of an unknown condition type. The list of valid sources is also documented in the create/update tool descriptions.
- **`rule_to_controller` is intentionally asymmetric with `rule_from_controller`.** The "from" path deeply normalizes all controller fields (id, timestamps, etc.) because the read tools need a stable snake_case surface. The "to" path only translates the small set of fields that actually cause a 400 (`is_created_by_system`, `history_conditions`, action metadata `use_thumbnail`). The `_RULE_TOP_LEVEL_RENAMES` dict makes the gap explicit and auditable.
- **Update is a full-body PATCH.** Protect rejects partial PATCH bodies. The expected flow is documented on the `update_rule` tool: (1) call `get_rule`, (2) mutate the returned dict, (3) pass back as `body`. The tool layer translates to camelCase once and shares the translated body between the preview and apply paths so the proposed-changes preview matches what actually gets PATCHed.
- **All write tools take `confirm: bool`.** Default false returns a preview (current vs proposed for update; proposed-only for create; warning + current state for delete). Matches the pattern used by the existing `apply_*` action tools.

## GraphQL / REST exposure

- `AlarmRule` plus nested `AlarmRuleSource` / `AlarmRuleCondition` / `AlarmRuleAction` / `AlarmRuleCooldown` are registered as Strawberry types in `apps/api/src/unifi_api/graphql/types/protect/alarms.py`.
- Read tools (`protect_alarm_list_rules`, `protect_alarm_get_rule`) are registered as Phase 6 type-migrated tools via `type_registry_init.py`, so they dispatch through the shared type registry for both REST and GraphQL.
- Mutation tools (`protect_alarm_create_rule` / `update_rule` / `delete_rule`) extend the existing `AlarmMutationAckSerializer` (which already covers arm/disarm).
- `dispatch_overrides.py` registers `update_rule` / `delete_rule` so the action endpoint targets the mutation method instead of the preview that the AST walker would otherwise capture (`create_rule` doesn't need an override -- its tool body's first manager await is `create_rule` itself, not a preview).
- `schema.graphql` and `graphql-reference.md` updated to match.

## Test coverage

Unit tests at three layers:

- **Manager (`test_alarm_manager_rules.py`)** -- 13 tests: list shape variants, get filter behavior + not-found, update happy path + non-dict response coercion + whitespace id, create happy path + non-dict body, delete happy path + empty body handling. All written against an `AsyncMock` of `connection_manager.client.api_request` / `api_request_raw` with raw camelCase payloads matching what the live controller returns.
- **Models (`test_alarms.py`)** -- 10 tests for `rule_to_controller` covering happy path, both-casings-provided (camelCase wins), unknown-fields passthrough, nested action metadata translation, non-dict / non-list edge cases, and the full `from_controller` → mutate → `to_controller` round-trip.
- **Input validators (`test_actions.py`)** -- 9 tests across `TestAlarmCreateRuleInput` / `TestAlarmUpdateRuleInput` covering valid input + empty-actions reject + missing-actions reject + non-list-actions reject.

Total new tests: 32. Full unifi-core protect suite: 324/324 pass. Full apps/protect suite: 379/379 pass. `ruff format --check` and `ruff check` both clean (pinned version per uv.lock).

## Verified live

All five MCP tools exercised end-to-end against a live Protect 7.x controller:
- `list_rules` -- returns the full rule set with the same shape as the test fixtures.
- `get_rule` -- per-rule fetch returns the row from the list. Not-found path returns a clean `UniFiNotFoundError`.
- `create_rule` -- exercised with both snake_case input (translated via `rule_to_controller`) and native camelCase. Controller returns 200 with the server-assigned id and the rule renders correctly in the Protect UI under the expected `Vehicle Description` / `Specific` binding for a `license_plate_known` condition.
- `update_rule` -- read-modify-write flow validated: `get_rule` (snake_case) → mutate (rename + cooldown bump) → `update_rule`. Re-read confirms persistence.
- `delete_rule` -- returns `{"deleted": true, "rule_id": ...}` and the rule is gone on the next `list_rules`.

Empty-actions rejection was also verified live: a body with `actions: []` is rejected client-side with the documented error, never reaching the controller (so the rule that would brick the UI never gets created).

The GraphQL/REST surface was exercised end-to-end against the same live controller through a running `apps/api` instance:
- GraphQL `alarmRules(controller)` -- returns the AlarmRuleList wrapper with the full set of automations from the controller.
- GraphQL `alarmRule(controller, id)` -- single-rule fetch returns the rule; not-found path propagates `UniFiNotFoundError` to a GraphQL error correctly.
- REST `POST /v1/actions/protect_alarm_create_rule` -- create-via-action with a snake_case body returned 200 with the server-assigned id (the `dispatch_overrides.py` routing and the snake/camel translation both engaged).
- REST `POST /v1/actions/protect_alarm_update_rule` -- rename + cooldown bump persisted; verified by re-reading via `alarmRule(id)` and by direct MCP read against the controller.
- REST `POST /v1/actions/protect_alarm_delete_rule` -- returns `{deleted: true, rule_id: ...}`; verified gone by `alarmRule(id)` (expected not-found) and by direct MCP read against the controller.

## Scope note

This is a larger-than-typical PR (~2.5k lines) because foundation (manager + models + their tests), the MCP tool layer + input models + manifest, AND the GraphQL/REST API exposure are interdependent: the `apps/api` `SerializerRegistry` enforces that every MCP tool in the manifest has a projection (Strawberry type or serializer), so adding the tools without the API surface fails CI. Splitting along those layer boundaries would create artificial dependency chains rather than independently-mergeable PRs.

## Out of scope (separate PRs)

- **License-plate identity lookup tool.** A companion `protect_list_known_license_plates` tool is in a sibling PR so an MCP caller can discover the `id` values needed for a `license_plate_known` condition. The two PRs are independent at the file level; this one only references the condition source by name in its tool description.
- **Arm-profile CRUD.** Out of scope here -- this PR is rule (automation) CRUD only. Arm profiles are the preset bundles (e.g. "Arm Night", "Arm All") that group automations together with an activation delay and schedule; arming with `PATCH arm + armProfileId` activates a profile's set. The existing `protect_alarm_list_profiles` already reads them, but creating / editing / deleting the profile definitions themselves is not covered here -- a reasonable follow-up PR.

## Closes

(none -- contributes to filling out the Alarm Manager surface)

### DIFF
--- a/apps/api/docs/graphql-reference.md
+++ b/apps/api/docs/graphql-reference.md
@@ -190,6 +190,28 @@ type AlarmProfileList {
   count: Int
 }
 
+"""
+A UniFi Protect alarm rule (Alarm Manager automation) — flat passthrough.
+"""
+type AlarmRule {
+  id: ID
+  name: String
+  enable: Boolean
+  isCreatedBySystem: Boolean
+  sources: JSON
+  conditions: JSON
+  historyConditions: JSON
+  schedules: JSON
+  actions: JSON
+  cooldown: JSON
+}
+
+"""Wrapper for protect_alarm_list_rules — {rules, count}."""
+type AlarmRuleList {
+  rules: JSON
+  count: Int
+}
+
 """UniFi Protect alarm system arm-state snapshot."""
 type AlarmStatus {
   armed: Boolean
@@ -1349,6 +1371,14 @@ type ProtectQuery {
   """List configured alarm profiles ({profiles, count})."""
   alarmProfiles(controller: ID!): AlarmProfileList
 
+  """
+  List configured alarm rules / Alarm Manager automations ({rules, count}).
+  """
+  alarmRules(controller: ID!): AlarmRuleList
+
+  """Fetch a single alarm rule (automation) by id."""
+  alarmRule(controller: ID!, id: ID!): AlarmRule
+
   """List Protect events (paginated, most recent first)."""
   events(controller: ID!, limit: Int! = 50, cursor: String = null, eventType: String = null, cameraId: String = null): EventPage!
 
@@ -1989,6 +2019,8 @@ Read-only access to UniFi Protect resources.
 **Fields:**
 
 - `alarmProfiles: AlarmProfileList`  — List configured alarm profiles ({profiles, count}).
+- `alarmRule: AlarmRule`  — Fetch a single alarm rule (automation) by id.
+- `alarmRules: AlarmRuleList`  — List configured alarm rules / Alarm Manager automations ({rules, count}).
 - `alarmStatus: AlarmStatus`  — Get the alarm system arm-state snapshot.
 - `camera: Camera`  — Look up a single Protect camera by id.
 - `cameraAnalytics: CameraAnalytics`  — Get analytics summary for a Protect camera.
--- a/apps/api/src/unifi_api/graphql/resolvers/protect.py
+++ b/apps/api/src/unifi_api/graphql/resolvers/protect.py
@@ -33,7 +33,12 @@ from strawberry.types import Info
 
 from unifi_api.graphql.context import GraphQLContext
 from unifi_api.graphql.permissions import IsRead
-from unifi_api.graphql.types.protect.alarms import AlarmProfileList, AlarmStatus
+from unifi_api.graphql.types.protect.alarms import (
+    AlarmProfileList,
+    AlarmRule,
+    AlarmRuleList,
+    AlarmStatus,
+)
 from unifi_api.graphql.types.protect.cameras import (
     Camera,
     CameraAnalytics,
@@ -415,6 +420,48 @@ async def _fetch_alarm_profiles(ctx: GraphQLContext, controller: str) -> Any:
                 "protect",
             )
             return await mgr.list_arm_profiles()
+
+    return await ctx.cache.get_or_fetch(key, _do)
+
+
+async def _fetch_alarm_rules(ctx: GraphQLContext, controller: str) -> Any:
+    key = f"protect/alarm-rules/{controller}"
+
+    async def _do() -> Any:
+        async with ctx.sessionmaker() as session:
+            mgr = await ctx.manager_factory.get_domain_manager(
+                session,
+                controller,
+                "protect",
+                "alarm_manager",
+            )
+            await ctx.manager_factory.get_connection_manager(
+                session,
+                controller,
+                "protect",
+            )
+            return await mgr.list_rules()
+
+    return await ctx.cache.get_or_fetch(key, _do)
+
+
+async def _fetch_alarm_rule(ctx: GraphQLContext, controller: str, rule_id: str) -> Any:
+    key = f"protect/alarm-rules/{controller}/{rule_id}"
+
+    async def _do() -> Any:
+        async with ctx.sessionmaker() as session:
+            mgr = await ctx.manager_factory.get_domain_manager(
+                session,
+                controller,
+                "protect",
+                "alarm_manager",
+            )
+            await ctx.manager_factory.get_connection_manager(
+                session,
+                controller,
+                "protect",
+            )
+            return await mgr.get_rule(rule_id)
 
     return await ctx.cache.get_or_fetch(key, _do)
 
@@ -1056,6 +1103,45 @@ class ProtectQuery:
         if raw is None:
             return None
         return AlarmProfileList.from_manager_output(raw)
+
+    @strawberry.field(
+        permission_classes=[IsRead],
+        description="List configured alarm rules / Alarm Manager automations ({rules, count}).",
+    )
+    async def alarm_rules(
+        self,
+        info: Info,
+        controller: strawberry.ID,
+    ) -> AlarmRuleList | None:
+        ctx: GraphQLContext = info.context
+        raw = await _fetch_alarm_rules(ctx, controller)
+        if raw is None:
+            return None
+        wrapped = (
+            raw
+            if isinstance(raw, dict) and "rules" in raw
+            else {
+                "rules": list(raw) if isinstance(raw, list) else [],
+                "count": len(raw) if isinstance(raw, list) else 0,
+            }
+        )
+        return AlarmRuleList.from_manager_output(wrapped)
+
+    @strawberry.field(
+        permission_classes=[IsRead],
+        description="Fetch a single alarm rule (automation) by id.",
+    )
+    async def alarm_rule(
+        self,
+        info: Info,
+        controller: strawberry.ID,
+        id: strawberry.ID,
+    ) -> AlarmRule | None:
+        ctx: GraphQLContext = info.context
+        raw = await _fetch_alarm_rule(ctx, controller, str(id))
+        if raw is None:
+            return None
+        return AlarmRule.from_manager_output(raw)
 
     # ---- Events ----------------------------------------------------------
 
--- a/apps/api/src/unifi_api/graphql/schema.graphql
+++ b/apps/api/src/unifi_api/graphql/schema.graphql
@@ -179,6 +179,28 @@ type AlarmProfileList {
   count: Int
 }
 
+"""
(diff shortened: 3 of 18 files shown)


sirkirby (changes requested):
Thanks Richard — this is a meaty contribution that genuinely fills out the Alarm Manager surface. The PR body documenting the controller quirks (empty actions bricking the UI, `GET automations/{id}` 404ing, the DELETE empty-body decode issue) is exactly the kind of context we want preserved. Live-verified against 7.x is great.

A few items before merge:

## Hard blocker

**Nested action metadata violates the "camelCase wins" rule.** Copilot caught this and it's real — `_action_to_controller` in `models/alarms.py:447-455` is a single-pass comprehension, so a body with both `useThumbnail` and `use_thumbnail` in action metadata will have the snake_case clobber the camelCase depending on dict order. Top-level `rule_to_controller` handles this correctly via two passes; the nested helper needs to mirror that. Worth a parallel test in `test_alarms.py` (the `test_both_snake_and_camel_*` style you have for top-level) to lock the behavior down.

## CI failures — we'll handle on our end

The `apps/api` serializer-projection / dispatch-override wiring is a downstream surface you wouldn't have hit on the protect side. I'll wire the 5 new tools into `apps/api/src/unifi_api/serializers/protect/alarms.py` and `apps/api/src/unifi_api/services/dispatch_overrides.py` and run `scripts/generate_skill_references.py` on our end once the substantive items below are in. No action needed from you on those gates.

## Should fix

- Module docstring at `alarm_manager.py:21` lists `GET automations/{id}` as a supported endpoint, but the implementation correctly notes it 404s. Drop that line or annotate it.
- The `update_rule` tool description should warn more loudly that this is a full-body PATCH — if the LLM caller omits a field, the controller treats it as a delete. Right now the "expected flow" framing reads as advisory rather than mandatory.

## Nit

- `get_rule(" rule-001 ")` would silently miss-match since `_require_rule_id` only rejects whitespace-only — consider stripping in the helper.

Great work on the snake/camel translation strategy overall, and on shipping the empty-actions client-side reject — that's the kind of UX footgun worth catching before it hits the controller.

sirkirby (changes requested):
GraphQL exposure tacked on is a nice bonus — saved a fork-edit pass.

Couple things still:

**Nested camelCase-wins is order-dependent.** The two-pass pattern didn't make it into `_action_to_controller` — the inner comprehension still depends on metadata insertion order:

| Order | `useThumbnail` returned |
|---|---|
| `useThumbnail: False`, then `use_thumbnail: True` | `True` (snake clobbers camel) |
| `use_thumbnail: True`, then `useThumbnail: False` | `False` (correct) |

Mirror the top-level two-pass inside the metadata branch, and add the parallel of `test_both_snake_and_camel_prefers_camel` for nested metadata (that test would have caught this):

```python
def test_nested_both_snake_and_camel_prefers_camel(self) -> None:
    body = {"actions": [{"metadata": {"use_thumbnail": True, "useThumbnail": False}}]}
    out = rule_to_controller(body)
    meta = out["actions"][0]["metadata"]
    assert meta["useThumbnail"] is False
    assert "use_thumbnail" not in meta
```

**`alarm_manager.py:21` still misleads.** The endpoint table lists `GET automations/{id} -- fetch a single rule` but that route 404s and `get_rule` filters the list instead. Drop the line or annotate it.

sirkirby (approved):
Rebased onto main (resolved the manifest/skill-ref conflicts from the recognition + lockfile-gate merges) and pushed the remaining review items straight to your branch rather than bounce another round:

- **Hard blocker — nested camelCase.** `_action_to_controller` metadata is now a two-pass mirror of the top-level `rule_to_controller`, so an explicit camelCase key always wins over its snake_case sibling regardless of dict order. Added `test_nested_both_snake_and_camel_prefers_camel` (both orderings) + a snake-only case.
- **Docstring.** Annotated the `GET automations/{id}` line as 404ing (no per-rule GET; `get_rule` filters the list).
- **Nit — id strip.** `_require_rule_id` now returns the stripped id and callers rebind it, so a padded `" rule-001 "` matches instead of silently missing. Added `test_get_rule_strips_padded_id`.

The `update_rule` full-body warning and the whole `apps/api` surface (GraphQL types/resolvers/schema, serializers, dispatch overrides, type registry) you'd already handled — nicely done, and registering `list_rules` as a `detail` wrapper (mirroring `AlarmProfileList`) sidesteps the dict-vs-list action-serializer issue cleanly.

Verification:
- unifi-core protect 327, apps/protect 385, apps/api 808 — all pass; ruff + format clean; generated artifacts in sync.
- Live (Protect 7.x): `list_rules`/`get_rule` MCP-direct; the action endpoint and GraphQL `alarmRules`/`alarmRule` all return real data; and `rule_to_controller` round-tripped over every live rule with zero snake/camel collisions.

Clearing my change request — good to merge once CI re-confirms green.

PR #311, opened 2026-05-28
feat(protect): list known license plates via recognition/vehicle/groups, plus GraphQL/REST exposure

rmaher001:
## Summary

Add `protect_list_known_license_plates`, a read-only MCP tool that lists Protect's license-plate identities (vehicle recognition groups) via the private `recognition/vehicle/groups` endpoint. The new tool mirrors the existing `protect_list_known_faces` exactly -- same parameter signature, same pagination/ordering/group-type filters, same response shape modulo the rename `faces` → `license_plates`.

## Motivation

The Protect alarm-rule schema supports a `license_plate_known` condition that references a specific recognized-vehicle identity by id. Without a discovery tool, an MCP caller has no way to know what id values exist on the controller -- they would have to inspect existing rules in the UI and copy id strings out by hand. This tool closes that gap so an LLM (or any caller) can:

1. List the registered plates with `protect_list_known_license_plates`
2. Pick the id of the vehicle to match
3. Create an alarm rule with a `license_plate_known` condition whose `value` is that id

This mirrors the same workflow already supported for Known Faces via `protect_list_known_faces` (the existing tool's description language is preserved here for consistency). The shape of the underlying `recognition/vehicle/groups` endpoint is identical to `recognition/face/groups` -- both keyed by a stable group id, both returning the recognized text in `matched_name`, both surfacing controller image reference paths only (no image bytes fetched). The tool is `readOnlyHint=True, openWorldHint=False`.

## Implementation

- **`packages/unifi-core/src/unifi_core/protect/models/recognition.py`** -- new `KnownLicensePlate` Pydantic model, mirror of `KnownFace`. `matched_name` description clarifies it carries the recognized plate text (e.g. `"ABC1234"`). `metadata` typically carries vehicle attributes (`color`, `vehicleType` sub-dicts) for plated vehicles -- noted in the field description.
- **`packages/unifi-core/src/unifi_core/protect/models/recognition.py`** -- new `license_plate_from_controller(raw)` helper, parallel to `from_controller`, normalizing camelCase Protect fields into snake_case.
- **`packages/unifi-core/src/unifi_core/protect/managers/recognition_manager.py`** -- new `list_known_license_plates` method on `RecognitionManager`. Reuses the existing `_build_group_labels` helper (shared validation of `group_types` / `include_interest`) and the same `_VALID_ORDER_BY` / `_VALID_ORDER_DIRECTION` sets used by the faces method. Calls `recognition/vehicle/groups` with the same param shape (`labels`, `minConfidence`, `orderBy`, `orderDirection`, `pageSize`, optional `page`).
- **`apps/protect/src/unifi_protect_mcp/tools/recognition.py`** -- new `protect_list_known_license_plates` MCP tool registration. Same parameter signature as `protect_list_known_faces`; description explicitly cross-references that each entry's `id` is the value to use in a `license_plate_known` alarm-rule condition.

## Note on existing faces tool

The new tool intentionally does NOT double-serialize the way the existing `protect_list_known_faces` tool currently does (the existing tool calls `from_controller(face).model_dump(...)` again on already-dumped snake_case dicts from the manager -- safe but redundant). The new tool does a pass-through (`raw.get("license_plates", [])`) since the manager already serialized via `license_plate_from_controller + model_dump`. The pre-existing redundancy in the faces tool is not touched (out of scope); a small follow-up could trim it for consistency.

## Test coverage

- **`apps/protect/tests/unit/test_recognition.py`** -- new `TestRecognitionManagerLicensePlates` class:
  - `test_list_known_license_plates_success` -- mocks the group response, asserts the manager calls `recognition/vehicle/groups` with `labels="groupType:known,groupType:interest"`, validates the returned shape.
  - `test_list_known_license_plates_known_only` -- asserts `include_interest=False` builds `labels="groupType:known"`.
  - `test_list_known_license_plates_unknown_group_type` -- asserts `group_types=["unknown"]` builds `labels="groupType:unknown"` (override path).

Full apps/protect suite: 382/382 pass. Full unifi-core protect suite: 304/304 pass. `ruff format --check` and `ruff check` both clean (pinned version per uv.lock). Manifest regenerated via `make manifest`; `make check-generated` exit 0.

## Verified live

Verified end-to-end against a live Protect 7.x controller through three surfaces:
- **MCP tool** `protect_list_known_license_plates` -- returns the registered plates with the expected schema and pagination links.
- **GraphQL** `protect { knownLicensePlates(controller, limit, ...) { items { id name matchedName detectionsCount metadata } } }` -- returns the same plates with camelCase field names, full metadata (including `color` and `vehicleType` sub-dicts when Protect provides them).
- **REST** `GET /v1/sites/{site_id}/known-license-plates` -- returns the same plates with snake_case field names per the REST convention.

Two id formats observed in real Protect data: UUID (`697f9f5900c63903e419cb66`-style for older registered vehicles) and `vehicle_LPR-<PLATE>` (for newer registrations) -- both are accepted as the `value` in a `license_plate_known` condition. The id-format note is captured in the `KnownLicensePlate.id` field description.

## Out of scope

- **Vehicle attribute filtering (color / vehicleType).** Protect surfaces these as recognition metadata on the group object, but querying by them is a separate concern best served via smart detection event metadata rather than the recognition-group endpoint. Not added here.
- **Create / update / delete of license-plate identities.** Read-only intentionally for this PR -- mutating the recognition-group set is a richer workflow that deserves its own PR.

## Closes

(none -- complements the alarm-rule CRUD additions in the companion PR)

### DIFF
--- a/apps/api/docs/graphql-reference.md
+++ b/apps/api/docs/graphql-reference.md
@@ -759,6 +759,31 @@ type KnownFacePage {
   nextCursor: String
 }
 
+"""A UniFi Protect license-plate identity (vehicle recognition group)."""
+type KnownLicensePlate {
+  id: ID
+  name: String
+  matchedName: String
+  type: String
+  imagePath: String
+  enhancedPath: String
+  detectionsCount: Int
+  firstDetectedAt: String
+  lastDetectedAt: String
+  isNotificationEnabled: Boolean
+  isDegraded: Boolean
+  tags: JSON
+  description: String
+  createdAt: String
+  metadata: JSON
+}
+
+"""Paginated page of UniFi Protect license-plate identities."""
+type KnownLicensePlatePage {
+  items: [KnownLicensePlate!]!
+  nextCursor: String
+}
+
 """A UniFi Protect light (PIR-triggered floodlight)."""
 type Light {
   id: ID
@@ -1309,6 +1334,11 @@ type ProtectQuery {
 
   """List assigned Protect Known Faces / named face recognition groups."""
   knownFaces(controller: ID!, limit: Int! = 50, cursor: String = null, minConfidence: Int! = 30, includeInterest: Boolean! = true, groupTypes: [String!] = null, orderBy: String! = "name", orderDirection: String! = "asc"): KnownFacePage!
+
+  """
+  List UniFi Protect license-plate identities (vehicle recognition groups). Each entry's id is the value to use in a `license_plate_known` alarm-rule condition.
+  """
+  knownLicensePlates(controller: ID!, limit: Int! = 50, cursor: String = null, minConfidence: Int! = 30, includeInterest: Boolean! = true, groupTypes: [String!] = null, orderBy: String! = "name", orderDirection: String! = "asc"): KnownLicensePlatePage!
 
   """List paired Protect chimes (paginated)."""
   chimes(controller: ID!, limit: Int! = 50, cursor: String = null): ChimePage!
@@ -1971,6 +2001,7 @@ Read-only access to UniFi Protect resources.
 - `firmwareStatus: FirmwareStatus`  — Get firmware status for the NVR plus its devices.
 - `health: ProtectHealth`  — Get the NVR health snapshot (cpu / memory / storage).
 - `knownFaces: KnownFacePage!`  — List assigned Protect Known Faces / named face recognition groups.
+- `knownLicensePlates: KnownLicensePlatePage!`  — List UniFi Protect license-plate identities (vehicle recognition groups). Each entry's id is the value to use in a `license_plate_known` alarm-rule condition.
 - `lights: LightPage!`  — List Protect lights (PIR-triggered floodlights).
 - `liveviews: LiveviewPage!`  — List Protect liveviews (multi-camera grid layouts).
 - `recordingStatus: RecordingStatusList`  — Get current recording state for one or all cameras ({cameras, count}).
--- a/apps/api/docs/openapi-reference.md
+++ b/apps/api/docs/openapi-reference.md
@@ -1718,6 +1718,24 @@ No native ``protect_get_liveview`` tool — filter from LIST.
 
 **Returns:** `Page_KnownFaceModel_`
 
+### `GET /v1/sites/{site_id}/known-license-plates` — List Known License Plates
+
+
+**Parameters:**
+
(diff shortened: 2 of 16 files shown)


sirkirby (changes requested):
Thanks Richard — clean mirror of the faces tool, and closing the discovery loop for `license_plate_known` conditions in #310 is exactly the right pairing. The id-format note (UUID vs `vehicle_LPR-<PLATE>`) is a nice catch from live testing.

## CI failures — we'll handle on our end

The `apps/api` serializer-projection gate and `make generate-skill-references` drift are downstream surfaces you wouldn't have hit on the protect side. I'll add the projection in `apps/api/src/unifi_api/serializers/protect/recognition.py` and refresh the skill-reference tables on our end. No action needed from you on those gates.

## Should fix

- Tool-layer test is missing. The faces side has `TestProtectListKnownFacesTool` (success / unknown_group_types / empty / error). Worth at least an error-path test for plates since your pass-through deliberately diverges from the faces tool's double-serialize.

## Nit

- Tool description at `tools/recognition.py:104-106` mentions `vehicle_type` but the controller field surfaced via `metadata` is `vehicleType` (camelCase). Minor mislead.

sirkirby (approved):
Big lift expanding this from the MCP-only tool to the full GraphQL + REST + OpenAPI surface — that's the projection work I'd said we'd pick up on our end, and it's a faithful mirror of the faces pattern across every surface. Appreciate you taking it on.

Two items from the last round were still open, plus the scope expansion left one new test-parity gap, so I pushed them straight to your branch (co-authored) rather than bounce another round:

- **Tool-layer test** — added `TestProtectListKnownLicensePlatesTool` (success/empty/error). This covers the tool's pass-through wrapping, which deliberately diverges from the faces tool's double-serialize and had no tool-layer coverage.
- **REST route test** — added `test_list_known_license_plates_happy_path` + `_group_types_filter`, matching the known-faces REST coverage for the new `/known-license-plates` endpoint.
- **Description fix** — `metadata` is passed through raw, so callers see `vehicleType` (camelCase), not `vehicle_type`. Updated the tool description and regenerated the manifest.

Local: ruff clean, protect unit 30 pass, api recognition (REST + GraphQL) 12 pass, `make check-generated` exit 0. Clearing my change request — good to merge once CI re-confirms green.

PR #316, opened 2026-05-30
feat(protect): add known license plate update/delete tools

rmaher001:
## Summary

Adds write operations for UniFi Protect license-plate identities, completing the CRUD surface that began with `protect_list_known_license_plates`:

- **`protect_update_known_license_plate`** — update a license-plate group's `name`, `description`, or `is_notification_enabled` (preview/confirm).
- **`protect_delete_known_license_plate`** — delete a license-plate group (preview/confirm).

These mirror the existing Known Face write tools (`protect_update_known_face` / `protect_delete_known_face`) against the `recognition/vehicle/groups/{id}` endpoint, so the license-plate identity registry now has the same update/delete parity faces already have.

## Implementation

Three layers, following the Known Face pattern:

- **`packages/unifi-core`** — `KnownLicensePlateUpdate` model + `to_controller_plate_update` translator (snake_case → controller camelCase, rejects unknown/read-only fields); `DeleteKnownLicensePlateInput`; and the manager methods `get_known_license_plate`, `update_known_license_plate` / `apply_update_known_license_plate`, `delete_known_license_plate` / `apply_delete_known_license_plate`.
- **`apps/protect`** — the two `@server.tool` wrappers (preview/confirm) + regenerated `tools_manifest.json`.
- **`apps/api`** — exposed through the generic action endpoint exactly like the Known Face mutations: `DETAIL` render kind in the recognition serializer + dispatch overrides mapping each tool to its `apply_*` manager method.

### One deliberate divergence from the faces path

The vehicle-group `DELETE` returns an **empty body** on success. Routing it through the normal `api_request` path raises a spurious `Could not decode JSON` error *after the delete has already succeeded*. `apply_delete_known_license_plate` therefore uses `api_request_raw` (the same approach the alarm-rule delete already uses), guarded by a dedicated test that asserts `api_request_raw` is the call made.

Mutable fields are limited to `name`, `description`, `is_notification_enabled`; the plate text (`matched_name`) and all system fields are read-only.

**Scope:** update + delete only. A merge operation is intentionally out of scope — the type-agnostic `recognition/v2/merge-group` endpoint has not been verified for vehicle groups, so shipping it would be unverified.

## Tests

- Manager-level unit tests for preview field-isolation, unknown/read-only field rejection, the camelCase PATCH translation, the delete preview, and the `api_request_raw` delete path.
- Tool-level tests for preview/confirm on both tools plus the not-found error paths.
- `make check-generated` passes (manifest + skill references regenerated).

## Verified live

Built and exercised against a live UniFi Protect controller via the MCP gateway, on a disposable unlabeled plate group:

- update preview → confirm: the PATCH persists (`is_notification_enabled` + `description`), confirmed by a fresh re-fetch.
- delete preview → confirm: returns cleanly with no JSON-decode error (the empty-body path), and a follow-up lookup returns not-found — confirming the group was actually removed.


sirkirby (approved):
Reviewed against the community PR gate: clean mirror of the known-face write tools, tool→manager→to_controller wiring correct, manifest/skill-refs regenerated with zero drift, f-string logger clean. The api_request_raw empty-body DELETE divergence is justified and test-guarded. Verified read + both preview paths against a live controller (zero mutation); 396 protect tests pass; CI green. LGTM.

PR #334, opened 2026-06-05
fix(protect): fall back to legacy automations when v2 Alarm Manager is empty/unavailable

rmaher001:
## Summary

- `AlarmRulesFacade` now falls back to the legacy automations API whenever the v2 Alarm Manager (`/api/v2/alarms`) cannot serve rules on this console — an empty result or any 4xx — not only on a 403.
- v2 5xx / transient errors (`NvrError`) now propagate instead of being masked by a legacy fallback.
- Corrected the misleading "Requires a SuperAdmin credential" framing in both the `_meta` coverage notice and the `AlarmManagerService` docstring — an empty v2 result is a not-migrated console, not a permission problem.
- Added five RED-first façade tests covering the empty / 4xx / 5xx / `get_rule` paths.

Regression from #320.

## Root Cause

#320 repointed the alarm-rule tools at the v2 Alarm Manager (`AlarmManagerService` → `/api/v2/alarms/protect`). On a console where Protect still serves rules only through the legacy automations API, `/api/v2/alarms/protect` returns `200 []` — the endpoint exists but is unpopulated. The façade fell back to legacy only on `AlarmManagerPermissionError` (403), so an empty v2 response was surfaced as "no rules," hiding the automations that are actually configured. `protect_alarm_list_rules` returned 0 even with many legacy rules present.

## Fix

`AlarmRulesFacade.list_rules` / `get_rule` prefer v2 when it serves rules, and fall back to legacy when it cannot:

- Empty v2 (`[]`) → legacy (endpoint present but unpopulated, e.g. Protect not yet migrated to `/api/v2/alarms`).
- Any v2 4xx — `BadRequest` (incl. 404 not-found and the 400 global-alarm-manager mode) and `AlarmManagerPermissionError` (403) → legacy.
- v2 5xx / transient (`NvrError`) → propagate; a real v2 outage must not be silently hidden behind legacy.

`complete` is `True` only when v2 served the rules; legacy responses are flagged through the existing `_meta` coverage key.

## Validation

- `uv run --directory packages/unifi-core --frozen --extra protect pytest tests/protect -k alarm -q` — 113 passed
- `uv run --directory apps/protect --frozen pytest tests -k alarm -q` — 22 passed
- `uv run --directory apps/api --frozen pytest tests -k alarm -q` — 21 passed
- `ruff format --check .` and `ruff check .` — clean
- Live, against a UniFi OS console where Protect has not migrated to `/api/v2/alarms`: `protect_alarm_list_rules` returned the full set of configured legacy automations (previously `0`), each normalized to the canonical `AlarmRule` shape with 24-hex legacy ids, and the response carried the `_meta` coverage flag.

### DIFF
--- a/apps/protect/src/unifi_protect_mcp/tools/alarm.py
+++ b/apps/protect/src/unifi_protect_mcp/tools/alarm.py
@@ -36,10 +36,15 @@ from unifi_protect_mcp.runtime import alarm_facade, alarm_manager, server
 logger = logging.getLogger(__name__)
 
 # Standard MCP _meta key (reverse-DNS, project-owned) signalling that the active
-# alarm backend is the limited one — AI-powered alarms are not visible without a
-# SuperAdmin credential. Anchored in the MCP _meta convention (see tasks.py).
+# alarm backend is the legacy one — used when the UniFi-OS Alarm Manager (v2)
+# returns no rules on this console, so any AI-powered alarms are not included.
+# Anchored in the MCP _meta convention (see tasks.py).
 _ALARM_COVERAGE_META = "com.github.sirkirby.unifi-mcp/alarm-coverage"
-_ALARM_COVERAGE_NOTICE = "AI-powered alarms are not included; viewing them requires a SuperAdmin credential."
+_ALARM_COVERAGE_NOTICE = (
+    "Showing legacy Protect automations: the UniFi-OS Alarm Manager (/api/v2/alarms) "
+    "returned no rules or is unavailable on this console, so AI-powered alarms "
+    "(where supported) are not included."
+)
 
 
 @server.tool(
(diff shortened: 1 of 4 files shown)


sirkirby (approved):
Approved. Verified the fix traces correctly through both the alarm façade and upstream `uiprotect` exception mapping:

- v2 `[]` (endpoint present but unpopulated / not-migrated) → legacy fallback — the actual regression from #320.
- 4xx (`BadRequest`, incl. `GlobalAlarmManagerError` and 404) and 403 (`AlarmManagerPermissionError`) → legacy.
- 429/5xx (`NvrError`) → propagate; a real v2 outage is never masked behind legacy.

`get_rule` is symmetric and correctly catches `UniFiNotFoundError` (raised by the service when the rule isn't in the v2 list). RED-first tests cover empty/4xx/5xx/not-found. Ruff clean, no f-string loggers, 156 alarm tests green locally (113 core / 22 protect / 21 api), CI green.

Mutation tools (create/update/delete) remain legacy-rooted by design (v2 CRUD is Phase 3) — unchanged by this PR.

PR #336, opened 2026-06-07
feat(protect): surface recognized license plate identity on events

rmaher001:
## Summary

UniFi Protect surfaces recognized **face** identity (`recognized_person_id` / `_name` / `_confidence`) on event records, but there is no parallel for **License Plate Recognition**. This adds the equivalent plate identity, mirroring the face feature, so consumers can correlate a vehicle event to a recognized plate.

Three new optional fields on the `Event` and `SmartDetection` shapes (and their controller parsers):

- `recognized_plate_text` — the recognized plate string (e.g. `ABC123`)
- `recognized_plate_group_id` — the stable per-plate group UUID (survives OCR drift across captures of the same vehicle)
- `recognized_plate_confidence` — match confidence (0–100)

## How it works

`EventManager._license_plate_recognition_fields()` mirrors `_face_recognition_fields()`. On live payloads the carrying detected-thumbnail's `type` is `vehicle` (not `licensePlate`) while the event-level `smart_detect_types` contains `licensePlate`, so the method gates on the event-level enum and accepts thumbnail type `vehicle`/`licensePlate`. It is wired into **both** event projection paths — `_event_to_dict` (parsed) and `_raw_event_to_dict` (the raw `api_request_list` path that backs `protect_list_events` / `protect_get_event` / `protect_list_smart_detections`) — and reads `smart_detect_types` in either snake_case or camelCase so plate identity also surfaces on the raw list path.

Like faces, the `events/<id>` detail endpoint returns the plate text and confidence but drops the stable group id; `_lookup_event_plate_recognition_fields()` (mirroring the face backfill) recovers it from the list/search path when it is missing.

The `apps/api` GraphQL `Event`/`SmartDetection` types and REST projections expose the three fields; the GraphQL schema, OpenAPI spec, reference docs, and the protect tool manifest are regenerated.

## Testing

- **Unit:** raw camelCase extraction, end-to-end `_raw_event_to_dict`, candidate scoring, and the `get_event` group-id backfill (mirroring the face backfill test); GraphQL (event detail + smart detections) and REST (`get_event`) fixtures.
- **Live:** exercised `protect_list_smart_detections`, `protect_list_events`, and `protect_get_event` against a live controller — plate identity surfaces on the list/search paths, the `get_event` backfill recovers the group id, and existing face identity continues to surface alongside plate identity on mixed events.

## Notes

- `detected_thumbnail_id` is populated by the face extractor (from the matched face thumbnail); a pure-LPR event with no face thumbnail will not carry it. Left as-is for parity with the existing face path rather than guessing at a vehicle-thumbnail id.
- The plate `smart_detect_types` gate reads both snake_case and camelCase (`smartDetectTypes`); the existing face extractor reads snake_case only, which is fine for it because it also matches `type == "face"` thumbnails directly.

### DIFF
--- a/apps/api/docs/graphql-reference.md
+++ b/apps/api/docs/graphql-reference.md
@@ -643,6 +643,9 @@ type Event {
   recognizedPersonId: ID
   recognizedPersonName: String
   recognizedPersonConfidence: Int
+  recognizedPlateText: String
+  recognizedPlateGroupId: ID
+  recognizedPlateConfidence: Int
   detectedThumbnailId: ID
 }
 
@@ -1628,6 +1631,9 @@ type SmartDetection {
   recognizedPersonId: ID
   recognizedPersonName: String
   recognizedPersonConfidence: Int
+  recognizedPlateText: String
+  recognizedPlateGroupId: ID
+  recognizedPlateConfidence: Int
   detectedThumbnailId: ID
 }
 
--- a/apps/api/openapi.json
+++ b/apps/api/openapi.json
@@ -2270,6 +2270,39 @@
             ],
             "title": "Recognized Person Name"
           },
+          "recognized_plate_confidence": {
+            "anyOf": [
+              {
+                "type": "integer"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Recognized Plate Confidence"
+          },
+          "recognized_plate_group_id": {
+            "anyOf": [
+              {
+                "type": "string"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Recognized Plate Group Id"
+          },
+          "recognized_plate_text": {
+            "anyOf": [
+              {
+                "type": "string"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Recognized Plate Text"
+          },
           "score": {
             "anyOf": [
               {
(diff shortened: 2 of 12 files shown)


sirkirby (approved):
Reviewed and approved. This is a clean mirror of the existing face-recognition path, and the design holds up well.

**Quality gates (feature-addition path):**
- Ruff lint clean; no f-string loggers (the one new logger call uses `%s` lazy formatting)
- New fields are read-only (`mutable: False`, `Optional`/`default=None`) — no create/update validation surface
- Generated artifacts (manifest, `schema.graphql`, `openapi.json`, reference docs) regenerate with zero drift
- Not a shared `<Domain>Base`, so no blast-radius default concern

**Tests:** 366 core-protect + 45 protect-app + 26 api, all green.

**Nice call on the casing gate:** reading `smart_detect_types` in both snake_case and camelCase is necessary here (not incidental) — since the carrying thumbnail type is `vehicle` (ambiguous), the `licensePlate` enum gate is the only filter, so it has to catch the raw `smartDetectTypes` path or plate identity silently vanishes on the list/search tools. Both raw-path cases are covered by tests.

**Live verification:** ran the three event tools against a live controller. Tools clean, face identity surfaces with no regression, and the negative path is confirmed — a large batch of `vehicle` detections produced zero false plate fields, exactly as the `licensePlate` gate intends. Our controller has no LPR-capable camera, so the positive path (plate surfacing + `get_event` group-id backfill) leans on your documented live test plus the unit/integration coverage, which exercises both.

Merging via squash.

PR #337, opened 2026-06-07
fix(access): handle structured location object on AccessDevice

rmaher001:
## Summary

`AccessDevice.location` is typed `Optional[str]`, but UNVR's Access API returns `location` as a **structured object** (door / floor / building reference). As a result, `access_list_devices` fails with a Pydantic `ValidationError` on every call against a real controller (`Input should be a valid string ... input_type=dict`) — the device-listing tool is unusable.

## Fix

- **unifi-core**: add an `AccessLocation` model (`unique_id` / `name` / `up_id` / `location_type` / `full_name` / `level`) and retype `AccessDevice.location` to `Optional[AccessLocation]`. A `_coerce_location` helper accepts the structured dict, an existing `AccessLocation`, or a bare string (mapped to `AccessLocation(name=...)`), returning `None` for empty values — so both the API-path object and the proxy-path door-name string parse.
- **apps/api**: mirror `AccessLocation` as a Strawberry type and retype the GraphQL/REST `AccessDevice.location`; `from_manager_output` coerces the typed model object, a nested dict, or the proxy `_door_name` string. Regenerated `schema.graphql` and `graphql-reference.md`.

This is an intentional contract change for the device projection (`location` string → object). `AccessDoor.location` is left `Optional[str]` — the controller returns door locations as plain strings.

## Testing

- **Unit**: `_coerce_location` across dict / string / typed object / empty; the apps/api projection from a typed unifi-core model object and from raw dicts; a GraphQL fixture asserting the nested location surfaces; serializer-shape and `location=None` contracts.
- **Live**: verified against a real UNVR — `access_list_devices` now succeeds and returns the structured location with all fields populated, where previously it raised a hard `ValidationError` on every call.

### DIFF
--- a/apps/api/docs/graphql-reference.md
+++ b/apps/api/docs/graphql-reference.md
@@ -16,7 +16,7 @@ type AccessDevice {
   type: String
   isOnline: Boolean
   firmwareVersion: String
-  location: String
+  location: AccessLocation
 }
 
 """Paginated page of UniFi Access devices."""
@@ -54,6 +54,18 @@ type AccessHealth {
   numDoors: Int
   numDevices: Int
   numOfflineDevices: Int
+}
+
+"""
+Structured location reference (door / floor / building) for an Access device.
+"""
+type AccessLocation {
+  uniqueId: ID
+  name: String
+  upId: ID
+  locationType: String
+  fullName: String
+  level: Int
 }
 
 """Read-only access to UniFi Access resources."""
--- a/apps/api/src/unifi_api/graphql/schema.graphql
+++ b/apps/api/src/unifi_api/graphql/schema.graphql
@@ -5,7 +5,7 @@ type AccessDevice {
   type: String
   isOnline: Boolean
   firmwareVersion: String
-  location: String
+  location: AccessLocation
 }
 
 """Paginated page of UniFi Access devices."""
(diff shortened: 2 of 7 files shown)


sirkirby (approved):
Reviewed and approved. Solid fix for a real, total-failure bug — `access_list_devices` was raising a Pydantic `ValidationError` on every call against a real controller because the Access API returns `location` as a structured object, not a string.

**Quality gates:**
- Ruff lint clean
- New `AccessLocation` is read-only and correctly wired — `access_list_devices` / `access_get_device` run each device through `from_controller(...).model_dump()`, and `_coerce_location` handles dict / typed object / bare string / empty / unexpected-shape defensively
- `schema.graphql` + `graphql-reference.md` regenerated; `openapi.json` correctly needs no change (AccessDevice isn't a typed REST schema). Re-ran full generation — zero drift
- Standalone model, all defaults `None` — no blast-radius concern

**Tests:** core access + access app + api access suites all green, covering every coercion input shape, the typed-core-model projection path, and nested `model_dump`.

**Contract change:** agreed this is an intentional, justified breaking change (`location` string → object) that needs no opt-in flag — the old string contract never succeeded against a real controller for the device-list tool, so the only thing it "breaks" is mock expectations, which this PR updates. Leaving `AccessDoor.location` as a string is the right call since doors return plain strings.

**Live before/after:** confirmed on a real controller via the proxy `devices/topology4` path (the crash path) — `main` raises the `ValidationError`, this branch returns a clean structured location with all modeled fields populated and the controller's extra fields ignored.

Merging via squash.

PR #349, opened 2026-06-12
fix(network): correct traffic-flows action/risk filter enum hints

rmaher001:
## Summary

The `unifi_get_traffic_flows` `action` and `risk` filter descriptions advertise enum values the controller's private v2 `/traffic-flows` endpoint rejects. The manager passes filter strings through verbatim, so the description is the contract — and the documented values produce hard 400s against a real controller:

- **action**: hinted `allow/block`, but the endpoint accepts only `[allowed, blocked]`. Passing `allow` returns `400 ... not one of the values accepted for Enum class: [allowed, blocked]`.
- **risk**: hinted `low/medium/high/critical`, but `critical` returns `400 ... Can not find enum ... by toString() value: critical`. The accepted band is `low/medium/high`.

## Fix

- Correct the `action` and `risk` filter descriptions on the `unifi_get_traffic_flows` tool and the unifi-core `TrafficFlowQuery` model.
- Make the `TrafficFlow.action` output-field description a noun phrase (`"Allow/Block"` → `"Action taken (allowed/blocked)"`), consistent with the sibling field descriptions.
- Regenerate `tools_manifest.json`.

Docs-only — no behavior change (the manager already passes filter values through unchanged).

## Testing

- Verified live against a real UniFi controller: `action=allowed`/`blocked` and `risk=low`/`medium`/`high` are accepted; `action=allow` and `risk=critical` return `400` with the messages quoted above.
- `make check-generated`, `ruff format --check`/`ruff check`, and the unifi-core + apps/network traffic-flow / manifest-drift suites all pass.


sirkirby (approved):
Approved. Verified the enum hint fix through the Docker Compose MCP path against the live Network controller: tool index advertises action=allowed/blocked and risk=low/medium/high; accepted values succeed and the old allow/critical values return controller 400s as expected.

PR #356, opened 2026-06-15
fix(acl): surface, set, and clear MAC ACL netmask across MCP + API

level99:
## Summary

MAC ACL rules can match a leading-bit MAC **prefix** — the UI's optional
"Netmask" dropdown (9–48; 24 = vendor OUI, 48 = full MAC) — but the MCP/API
layer silently dropped it. `from_controller` never read the controller's
`mac_mask` field, so a masked rule looked like a bare-MAC match and the mask
could be **neither seen, set, nor changed**.

This surfaces the netmask end-to-end across both the MCP tools and the API
(GraphQL/REST) server, adds the ability to set and clear it, and hardens the
shared model so the two surfaces can't drift.

**Root cause:** the read translator (`from_controller`) extracted only
`type` + `specific_mac_addresses` from each `traffic_source`/`traffic_destination`
block and discarded the sibling `mac_mask`. Verified against a live controller:
the IGMP/SSDP/mDNS rules carry `mac_mask: ff:ff:ff:00:00:00` (a /24) that the
tools rendered as a bare `01:00:5e:00:00:00`.

## Type of change

- [x] `fix:` bug fix

## Scope

- [x] Network MCP server
- [x] API server
- [x] Shared packages
- [x] Documentation

## What changed

- **Model** (`unifi-core/network/models/acl.py`): `AclRule` gains mutable
  `source_netmask`/`destination_netmask` (int 0–48) plus read-only
  `source_mac_mask`/`destination_mac_mask` (the raw 6-octet bitmask, so a
  non-contiguous mask the int can't express stays visible). Bidirectional
  int↔bitmask converters; read-side derivation shared by `from_controller`
  and the GraphQL type (`read_mask_fields`). Clearing emits a `None` sentinel
  the manager pops after fetch-merge-PUT (the controller 400s on an empty
  `mac_mask` and deep-merge can't delete a key). One shared `build_acl_rule`
  used by both surfaces. Hardened: range-guarded converters, strict 6-octet
  mask parsing, null/non-dict traffic blocks, case-insensitive action on update.
- **MCP tools**: `unifi_create_acl_rule` / `unifi_update_acl_rule` gain
  netmask params + `clear_*_netmask` flags; descriptions warn that `0` matches
  ANY MAC.
- **API dispatch** (`_translate_acl_update`): now reads the nested `rule_data`
  dict (it was reading top-level args and dropping **every** field update).
- GraphQL `AclRule` type + cross-layer symmetry; regenerated manifest, OpenAPI
  spec, and GraphQL reference.

## Files changed

| File | Change |
|---|---|
| `packages/unifi-core/.../models/acl.py` | netmask fields, converters, clear sentinel, shared builder, hardening |
| `packages/unifi-core/.../managers/acl_manager.py` | pop the `None` mask sentinel after merge (clears the mask) |
| `apps/network/.../tools/acl.py` | netmask + clear params on create/update; `0`-matches-any warning |
| `apps/api/.../services/dispatch_overrides.py` | read nested `rule_data`; clean error envelope; clear-flag passthrough |
| `apps/api/.../graphql/types/network/acl.py` | expose netmask fields via shared `read_mask_fields` |
| `tests/` (acl_tools, acl_manager, dispatcher, filtering) | model/manager/tool/API + clear + cross-surface coverage |
| `tools_manifest.json`, `openapi.json`, `schema.graphql`, `graphql-reference.md` | regenerated |

## Checklist

- [x] I ran the focused tests/checks for the files I changed.
- [ ] I ran `make pre-commit` — `make` isn't installed on this Windows dev host, so I ran the equivalents individually: `ruff check` / `ruff format`, per-package `pytest`, and regenerated every generated artifact (`tools_manifest.json`, `openapi.json`, `schema.graphql`, and the OpenAPI/GraphQL reference docs). The full gate — all three app packages, the Node worker typecheck, and the generated-artifact drift checks — runs in CI.
- [x] I regenerated generated manifests (`tools_manifest.json`, `openapi.json`, `graphql-reference.md`).
- [x] I added/updated tests for the behavior changes.
- [x] I updated documentation (regenerated API reference) for user-visible changes.

## Testing

| Environment | Details |
|---|---|
| **UniFi OS** | 5.0.16 |
| **MCP Client** | Claude Code (stdio) |

| Live test | Result |
|---|---|
| Existing IGMP/SSDP/mDNS /24 rules now surface `destination_netmask=24` + raw `mac_mask` | ✅ |
| `unifi_create_acl_rule` with `destination_netmask=24` → reads back 24 (MCP tool path) | ✅ |
| Update 24→16 via the MCP tool **and** the API dispatch translator (nested `rule_data`) | ✅ |
| `clear_destination_netmask` → mask removed, MACs preserved (MCP + API paths) | ✅ |
| Out-of-range netmask (99) → graceful `{"success": false}` error, not an exception | ✅ |
| All disposable test rules created `enabled: false` and deleted after verification | ✅ |

Unit tests (focused, per-package):

```
apps/network  test_acl_tools.py + test_acl_manager.py — 82 passed
apps/api      test_actions_dispatcher / test_network_filtering / test_cross_layer_symmetry / test_openapi_drift — 123 passed
ruff check + format — clean
```

(One local-only failure, `test_build_dispatch_table_finds_real_tools`, is an
environment artifact — the Access app isn't installed in this machine's
api-server venv — unrelated to this change; it passes in CI where all apps
are present.)

## Notes for reviewers

- The **API dispatch fix** (`_translate_acl_update` reading `rule_data`) is the
  highest-impact change — the API update path previously dropped all field
  updates; the dispatcher tests had encoded a flat arg shape that masked it.
- **Clearing** a netmask uses a `None`-mask sentinel popped by the manager
  (the controller rejects an empty `mac_mask`, and deep-merge can't delete a
  key). `netmask=None` on update remains a safe no-op (round-trip friendly);
  removal is explicit via `clear_*_netmask`.
- **Non-contiguous masks** (API-legal, UI can't express) round-trip via the
  read-only raw `*_mac_mask` field; the int `*_netmask` is `null` for them.

### DIFF
--- a/apps/api/docs/graphql-reference.md
+++ b/apps/api/docs/graphql-reference.md
@@ -148,6 +148,10 @@ type AclRule {
   destinationType: String
   sourceMacs: [String!]!
   destinationMacs: [String!]!
+  sourceNetmask: Int
+  destinationNetmask: Int
+  sourceMacMask: String
+  destinationMacMask: String
 }
 
 """Paginated page of ACL rules."""
--- a/apps/api/openapi.json
+++ b/apps/api/openapi.json
@@ -26,12 +26,34 @@
             ],
             "title": "Action"
           },
+          "destination_mac_mask": {
+            "anyOf": [
+              {
+                "type": "string"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Destination Mac Mask"
+          },
           "destination_macs": {
             "items": {
               "type": "string"
             },
             "title": "Destination Macs",
             "type": "array"
+          },
+          "destination_netmask": {
+            "anyOf": [
+              {
+                "type": "integer"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Destination Netmask"
           },
           "destination_type": {
             "anyOf": [
@@ -81,12 +103,34 @@
             ],
             "title": "Network Id"
           },
+          "source_mac_mask": {
+            "anyOf": [
+              {
+                "type": "string"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Source Mac Mask"
+          },
           "source_macs": {
             "items": {
               "type": "string"
             },
             "title": "Source Macs",
             "type": "array"
+          },
+          "source_netmask": {
+            "anyOf": [
+              {
+                "type": "integer"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Source Netmask"
           },
           "source_type": {
             "anyOf": [
--- a/apps/api/src/unifi_api/graphql/schema.graphql
+++ b/apps/api/src/unifi_api/graphql/schema.graphql
@@ -137,6 +137,10 @@ type AclRule {
   destinationType: String
   sourceMacs: [String!]!
   destinationMacs: [String!]!
+  sourceNetmask: Int
+  destinationNetmask: Int
+  sourceMacMask: String
+  destinationMacMask: String
 }
 
 """Paginated page of ACL rules."""
--- a/apps/api/src/unifi_api/graphql/types/network/acl.py
+++ b/apps/api/src/unifi_api/graphql/types/network/acl.py
@@ -19,6 +19,7 @@ from dataclasses import asdict
 from typing import Any
 
 import strawberry
+from unifi_core.network.models.acl import read_mask_fields
 
 
 def _get(obj: Any, key: str, default: Any = None) -> Any:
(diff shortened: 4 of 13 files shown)


sirkirby (approved):
Approved. I pushed one follow-up commit to keep the API action translator consistent with the MCP tool path: nested ACL rule_data now rejects unknown/read-only fields instead of silently dropping them, with dispatcher coverage for that behavior and the intentional netmask=None no-op case.

Focused unit tests passed (network ACL + API dispatcher/cross-layer coverage), ruff passed on the changed files, and local live smoke passed for both MCP Network ACL netmask create/update/clear/delete and the API action path.

PR #372, opened 2026-06-18
fix: verify WLAN/network/AP-group updates persisted before reporting success

### LINKED ISSUE #371: bug(wlan): unifi_update_wlan reports success but does not persist changes (e.g. proxy_arp, minrate_ng_data_rate_kbps)
## Summary

`unifi_update_wlan` returns `success: true` with a populated `updated_fields` list and a matching change `preview`, but the change is **never persisted** to the controller. An immediate, independent `unifi_get_wlan_details` shows every targeted field unchanged. Reproduced across multiple field combinations, including a single *recognized* boolean field (`proxy_arp`) — so this is not explained by #136 (silent drop of *unknown* kwargs).

## Environment

- Package: `unifi-network-mcp` via `uvx unifi-network-mcp@latest` (PyPI latest at report time: `0.20.5`)
- Controller: UniFi Dream Machine Pro (UDMPRO), gateway firmware `5.1.15.33416`
- Tool registration mode: `lazy`; transport: stdio
- Target WLAN: WPA2-PSK, 2.4 GHz-only (`wlan_band: 2g`), `setting_preference: manual`

## Steps to reproduce

1. Note a WLAN's current `proxy_arp` (`false`) via `unifi_get_wlan_details`.
2. Call `unifi_update_wlan` with `update_data: {"proxy_arp": true}, confirm: true`.
3. Tool returns `success: true`, `updated_fields: ["proxy_arp"]`, and a `details` object.
4. Call `unifi_get_wlan_details` again.

## Expected

`proxy_arp` is `true`.

## Actual

`proxy_arp` is still `false` — nothing persisted. Notably, the `details` object returned **by the update call itself** also still shows the OLD value, while `updated_fields` claims the field changed. So `updated_fields` appears to echo the requested keys rather than reflect a verified write.

Also reproduced with `minrate_ng_data_rate_kbps: 6000` (stays `1000`) and `minrate_setting_preference: "manual"` (stays `"auto"`; this field never appears in the `preview` diff, consistent with #136 — but `proxy_arp` *is* recognized and still fails).

## Suspected cause

The returned `details` looks like the pre-write object and `updated_fields` is derived from the request, not a read-back — so success is reported even when the controller write is a no-op/rejected. Possibly the underlying `PUT` targets a path/shape this Network app version ignores. Possibly related: #116 (shallow merge), #136 (unknown-kwarg drop).

## Suggested fix

After writing, re-fetch and diff actual vs. requested fields; raise a real error if any requested field didn't change. And verify the write endpoint/method for current UniFi OS.

ljmerza:
## Summary

The fetch-merge-put manager methods (`update_wlan`, `update_network`, `update_ap_group`) PUT the full merged object to the controller and then returned success **without inspecting the result**. On some controllers / UniFi OS versions the legacy `/rest/wlanconf` and `/rest/networkconf` endpoints answer `rc: ok` but silently do **not** persist the requested fields. Because `aiounifi` only raises on `rc: error`, the manager returned `True` while nothing changed — and the tool layer reported `success: true` with an `updated_fields` list derived from the request, not a read-back.

Refs #371 (full root-cause trace in the issue thread).

## Change

- Add a shared `_unpersisted_fields(before, after, requested)` helper to `network_manager.py`.
- After each write, re-read the object and diff it: if a field was meant to change (requested value differs from the pre-write value) but did **not** move, return a real failure naming the stuck field(s).
- Applied to `update_wlan`, `update_network`, and `update_ap_group`.

Design choices to avoid false positives:
- **Write-only / redacted keys** (`x_passphrase`, `sae_psk`, `private_preshared_keys`, `x_iapp_key`, …) are skipped — they never round-trip.
- A field whose re-read value is *different but non-original* (controller normalization) is treated as **persisted**; only "did not move at all" is flagged.
- No-op requests (requested value already equals current) are ignored.

This converts the silent false-success into an actionable error. It does **not** change the write path itself — if a controller is ignoring the legacy full-object PUT, callers now get a clear error instead of a phantom success. Fixing the underlying write path (minimal payload / V2 endpoint) can be a follow-up.

## Testing

- New `packages/unifi-core/tests/network/managers/test_network_manager_update_verification.py` (8 cases): helper unit tests + `update_wlan` persisted/not-persisted/write-only + `update_network` not-persisted.
- Updated `test_update_ap_group` to supply the post-write re-read (now 3 controller calls: GET existence → PUT → GET verify).
- `make`-equivalent runs: `unifi-core` network suite **618 passed**; network-app `test_wifi_tools` + `test_network_tools` **45 passed**; `ruff check` + `ruff format --check` clean.

### DIFF
--- a/apps/network/tests/unit/test_wifi_tools.py
+++ b/apps/network/tests/unit/test_wifi_tools.py
@@ -376,8 +376,16 @@ class TestNetworkManagerApGroups:
                 "attr_hidden_id": "default",
             }
         ]
-        # First call: list_ap_groups (GET list), second call: PUT merged data
-        mock_connection.request.side_effect = [existing_groups, {}]
+        updated_groups = [
+            {
+                "_id": "grp1",
+                "name": "Renamed Group",
+                "device_macs": ["aa:bb:cc:dd:ee:01", "aa:bb:cc:dd:ee:02"],
+                "attr_hidden_id": "default",
+            }
+        ]
+        # GET list (existence) -> PUT merged data -> GET list (persistence re-check)
+        mock_connection.request.side_effect = [existing_groups, {}, updated_groups]
 
         result = await network_manager.update_ap_group(
             "grp1",
@@ -385,8 +393,8 @@ class TestNetworkManagerApGroups:
         )
 
         assert result is True
-        # Verify two requests were made: GET list then PUT
-        assert mock_connection.request.call_count == 2
+        # GET (existence) + PUT + GET (persistence verification)
+        assert mock_connection.request.call_count == 3
         # Second call should be the PUT with merged data
         put_call = mock_connection.request.call_args_list[1]
         put_req = put_call[0][0]
--- a/packages/unifi-core/src/unifi_core/network/managers/network_manager.py
+++ b/packages/unifi-core/src/unifi_core/network/managers/network_manager.py
@@ -14,6 +14,40 @@ CACHE_PREFIX_NETWORKS = "networks"
 CACHE_PREFIX_WLANS = "wlans"
 CACHE_PREFIX_AP_GROUPS = "ap_groups"
 
+# Fields the controller never echoes back verbatim (write-only / redacted), so a
+# re-read cannot confirm them. Excluded from post-write persistence verification.
+_UNVERIFIABLE_UPDATE_KEYS = frozenset(
+    {
+        "x_passphrase",
+        "x_password",
+        "sae_psk",
+        "private_preshared_keys",
+        "x_iapp_key",
+        "password",
+    }
+)
+
+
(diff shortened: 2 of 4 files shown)


sirkirby (approved):
Solid root-cause work — the trace through the manager and tool layers was exactly right. I made a few changes on top: rebased onto main, pulled the inline `#371` refs out of the code and comments (we keep issue refs in commits/PRs, not in source), and added an update step to the WLAN and AP-group smoke lifecycles so these paths get live coverage under `--phase approved` going forward. Verified against a controller here — success is only reported when the write actually persists, and you get a real error when it doesn't. Approving. Appreciate the thorough writeup and the offer to fix it.

PR #375, opened 2026-06-20
fix(redaction): catch underscore-split pre_shared_key fields

Aspoust:
## Problem

`x_ipsec_pre_shared_key` (L2TP/IPsec VPN networkconf field) leaked in cleartext through `unifi_get_network_details` and other read paths, despite the redaction policy intending to hide VPN preshared keys.

## Root cause

The field name splits into segments `["x", "ipsec", "pre", "shared", "key"]` because `pre` and `shared` are separated by an underscore in the real controller API field name. The existing qualifier check in `is_sensitive_key` only matched a single `preshared` segment immediately before `key`, so this 3-way split never matched.

## Fix

Added a 3-gram check in `packages/unifi-core/src/unifi_core/redaction.py` that recognizes the ("pre", "shared", "key") sequence alongside the existing single-token qualifier match. This generically covers any future field following the same `_pre_shared_key` naming pattern, not just this one field.

## Testing

TDD: added a failing test (`test_redacts_underscore_split_preshared_key_fields`) reproducing the gap, confirmed RED, applied the minimal fix, confirmed GREEN. Full unifi-core suite passes (873 tests; unrelated pre-existing collection errors for protect/aiounifi-dependent modules due to missing optional dependencies in this environment, not caused by this change).

Found while doing a live firewall and security audit on my own controller via the unifi-network MCP plugin -- the PSK appeared in cleartext in a unifi_get_network_details response.

### DIFF
--- a/packages/unifi-core/src/unifi_core/redaction.py
+++ b/packages/unifi-core/src/unifi_core/redaction.py
@@ -118,6 +118,11 @@ def is_sensitive_key(key: Any) -> bool:
     for left, right in zip(parts, parts[1:]):
         if right == "key" and left in _SENSITIVE_KEY_QUALIFIERS:
             return True
+    # "preshared" sometimes arrives underscore-split as "pre_shared" (e.g. the
+    # controller's x_ipsec_pre_shared_key field) — match that 3-gram too.
+    for first, second, third in zip(parts, parts[1:], parts[2:]):
+        if (first, second, third) == ("pre", "shared", "key"):
+            return True
     return False
 
 
--- a/packages/unifi-core/tests/test_redaction.py
+++ b/packages/unifi-core/tests/test_redaction.py
@@ -128,6 +128,15 @@ def test_redacts_role_infixed_private_and_preshared_keys() -> None:
     assert is_sensitive_key("wireguard_client_public_key") is False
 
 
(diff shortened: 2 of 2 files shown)


sirkirby (approved):
Approved. This is a narrow redaction helper fix and the inline tuple match for the underscore-split pre/shared/key form is appropriate here.

Verified locally against the current PR head (65502715bcc6e4809284a123547a8a3db25e4d16):

- uv run pytest packages/unifi-core/tests/test_redaction.py -q
- uv run pytest packages/unifi-core/tests -q
- uv run pytest apps/network/tests/unit/test_network_tools.py -q
- uv run pytest apps/network/tests/unit/test_vpn_tools.py -q
- uv run ruff check packages/unifi-core/src/unifi_core/redaction.py packages/unifi-core/tests/test_redaction.py
- uv run ruff format --check packages/unifi-core/src/unifi_core/redaction.py packages/unifi-core/tests/test_redaction.py

Also spot-checked a nested payload with x_ipsec_pre_shared_key to confirm the PSK is redacted while nearby non-secret/public-key fields remain visible. I don't see a live-smoke requirement as blocking for this one because it changes deterministic egress redaction behavior rather than controller I/O or tool mutation behavior.

PR #377, opened 2026-06-21
feat(network): allow unifi_update_traffic_route to edit route-match fields

### LINKED ISSUE #376: feat: unifi_update_traffic_route can't edit target devices / route-match fields
### Summary

`unifi_update_traffic_route` only exposes `enabled` and `kill_switch_enabled`. There is no way through the MCP to change **which clients/networks a policy-based (VPN) route applies to** (`target_devices`), nor its `domains` / `ip_addresses` / `ip_ranges` / `regions` / `next_hop` — even though the UniFi UI and controller API fully support editing these.

### Use case

Swapping the device a VPN route targets — e.g. replacing an old phone with a new one in a "route Second Life dev domains over the work VPN" rule — currently forces a drop to the UniFi UI. An agent doing a device migration can update firewall groups and client reservations via MCP, but hits a wall on traffic routes.

### Root cause (tool-layer only)

`TrafficRouteManager.update_traffic_route(route_id, enabled=None, **kwargs)` already:
- fetches the full current route,
- merges any non-`None` `**kwargs` into it,
- and does `PUT /trafficroutes/{route_id}` (V2 API) with the full object.

So the manager can already persist `target_devices` and friends — the **tool signature** simply never declares or forwards them, and hard-errors unless `enabled`/`kill_switch_enabled` is given. No controller-API limitation is involved.

### Proposed change

Widen `unifi_update_traffic_route` to also accept (each replacing the whole value, read-then-write):
- `target_devices` — list of `{type: CLIENT|NETWORK|ALL_CLIENTS, ...}` objects
- `domains`, `ip_addresses`, `ip_ranges` — list-of-object fields
- `regions` — list of ISO codes
- `next_hop` — string

Forward them through the existing manager, surface them in the confirm-preview, validate list shapes client-side, regenerate the manifest, and add tests.

I have an implementation ready and will open a PR referencing this issue.

ljmerza:
Closes #376.

### What

Widens `unifi_update_traffic_route` so it can edit a route's match/targeting, not just toggle it. New optional params:

- `target_devices` — replace which clients/networks the route applies to (`[{"type": "CLIENT", "client_mac": "…"}]`, `[{"type": "NETWORK", "network_id": "…"}]`, `[{"type": "ALL_CLIENTS"}]`)
- `domains`, `ip_addresses`, `ip_ranges` — list-of-object match fields
- `regions` — list of ISO country/region codes
- `next_hop` — static next-hop IP

`enabled` / `kill_switch_enabled` behave exactly as before.

### Why

There was no MCP path to change which device a policy-based (VPN) route targets — that required the UniFi UI. See #376 for the motivating device-swap scenario.

### How

This is a **tool-layer change only**. `TrafficRouteManager.update_traffic_route(route_id, enabled=None, **kwargs)` already fetches the full route, merges non-`None` kwargs, and `PUT`s the whole object to `/trafficroutes/{id}` (V2). The tool just never forwarded anything beyond the two toggles. The diff:

- forwards any provided field into the existing manager call;
- requires **at least one** updatable field (instead of specifically `enabled`/`kill_switch_enabled`);
- fetches the current route so the confirm-preview shows current-vs-proposed for the changed fields;
- validates list-shaped fields (and that each `target_devices` entry is an object with a `type`) client-side;
- updates the tool description and regenerates `tools_manifest.json`.

No manager or controller-API change.

### Testing

- New `apps/network/tests/unit/test_traffic_route_tools.py` covering target-device preview/apply, multi-field forwarding, the enabled-only path, and validation errors.
- `make manifest` regenerated (178 tools).
- `make format` / `make lint` clean; full `make test` → **797 passed** (incl. manifest-sync and tool-map-sync).

### Backward compatibility

Additive — all new params are optional and default `None`; existing toggle behavior is unchanged.

### DIFF
--- a/apps/api/src/unifi_api/services/dispatch_overrides.py
+++ b/apps/api/src/unifi_api/services/dispatch_overrides.py
@@ -78,6 +78,10 @@ DISPATCH_OVERRIDES: dict[str, tuple[str, str]] = {
     "unifi_toggle_qos_rule_enabled": ("qos_manager", "update_qos_rule"),
     "unifi_toggle_oon_policy": ("oon_manager", "toggle_oon_policy"),
     "unifi_toggle_traffic_route": ("traffic_route_manager", "toggle_traffic_route"),
+    # update_traffic_route now pre-fetches the route via get_traffic_route_details
+    # to render a current-vs-proposed preview, so the AST walker captures the read
+    # method first. Pin dispatch to the mutation method.
+    "unifi_update_traffic_route": ("traffic_route_manager", "update_traffic_route"),
     # update_device_radio: tool needs current radio_table to identify target band.
     "unifi_update_device_radio": ("device_manager", "update_device_radio"),
     # Stats: tool combines existence check on client/device with stats fetch.
--- a/apps/api/tests/test_actions_dispatcher.py
+++ b/apps/api/tests/test_actions_dispatcher.py
@@ -292,6 +292,10 @@ def test_dispatch_overrides_specific_targets() -> None:
     # Toggle that needs current state
     assert table["unifi_toggle_firewall_policy"].method == "toggle_firewall_policy"
     assert table["unifi_reorder_firewall_policies"].method == "reorder_firewall_policies"
+    # update_traffic_route pre-fetches the route for its preview, so the AST captures
+    # get_traffic_route_details; the override must pin it to the mutation method.
+    assert table["unifi_update_traffic_route"].manager_attr == "traffic_route_manager"
+    assert table["unifi_update_traffic_route"].method == "update_traffic_route"
     # Stats: list-returning method (was AST-captured as get_X_details, a dict)
     assert table["unifi_get_device_stats"].manager_attr == "stats_manager"
     assert table["unifi_get_device_stats"].method == "get_device_stats"
@@ -318,6 +322,65 @@ def test_dispatch_overrides_specific_targets() -> None:
     assert table["access_lock_door"].method == "apply_lock_door"
     assert table["access_create_credential"].method == "apply_create_credential"
     assert table["access_update_policy"].method == "apply_update_policy"
+
+
+@pytest.mark.asyncio
+async def test_dispatch_update_traffic_route_routes_to_mutation_with_widened_field() -> None:
+    """Regression: unifi_update_traffic_route pre-fetches the route via
+    get_traffic_route_details to render a current-vs-proposed preview, so the AST
+    dispatcher captures the READ method first. The DISPATCH_OVERRIDES entry must
+    redirect dispatch to the mutation method, and a widened match field
+    (target_devices) must reach the manager.
+
+    Uses the REAL build_dispatch_table() so this exercises the override wiring, not
+    just dispatch_action's arg forwarding.
+    """
+    entry = ToolEntry(
+        name="unifi_update_traffic_route",
+        product="network",
+        category="traffic_routes",
+        manager="",
+        method="",
+    )
+    registry = _registry_with(entry)
+
+    domain_manager = MagicMock()
+    domain_manager.update_traffic_route = AsyncMock(return_value=True)
+
+    conn_manager = MagicMock()
+    conn_manager.site = "default"
+    conn_manager.set_site = AsyncMock()
+
+    factory = MagicMock()
+    factory.get_domain_manager = AsyncMock(return_value=domain_manager)
+    factory.get_connection_manager = AsyncMock(return_value=conn_manager)
+
+    session = MagicMock()
+    target_devices = [{"type": "CLIENT", "client_mac": "aa:bb:cc:dd:ee:ff"}]
+    await dispatch_action(
+        registry=registry,
+        factory=factory,
+        session=session,
+        tool_name="unifi_update_traffic_route",
+        controller_id="cid",
+        controller_products=["network"],
+        site="default",
+        args={"route_id": "tr1", "target_devices": target_devices},
+        confirm=True,
+        dispatch_table=build_dispatch_table(),
+    )
+
+    # Override resolved to the traffic_route_manager mutation, NOT the read method
+    # get_traffic_route_details the AST would otherwise capture.
+    factory.get_domain_manager.assert_awaited_once_with(
+        session=session,
+        controller_id="cid",
+        product="network",
+        attr_name="traffic_route_manager",
+    )
+    # The widened match field is forwarded to update_traffic_route as a kwarg
+    # (no arg translator — the manager takes **kwargs; confirm is a separate arg).
+    domain_manager.update_traffic_route.assert_awaited_once_with(route_id="tr1", target_devices=target_devices)
 
 
 @pytest.mark.asyncio
--- a/apps/network/src/unifi_network_mcp/tools/traffic_routes.py
+++ b/apps/network/src/unifi_network_mcp/tools/traffic_routes.py
@@ -6,7 +6,7 @@ on a UniFi Network Controller using the V2 API.
 """
 
 import logging
-from typing import Annotated, Any, Dict, Optional
+from typing import Annotated, Any, Dict, List, Optional
 
 from mcp.types import ToolAnnotations
 from pydantic import Field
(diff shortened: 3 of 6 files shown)


sirkirby (changes requested):

- [ ] Add an API action-dispatch override for `unifi_update_traffic_route`.

  On this branch, `build_dispatch_table()` resolves `unifi_update_traffic_route` to `traffic_route_manager.get_traffic_route_details`, not the mutation method. That changed because the tool now fetches current route state for preview before calling `update_traffic_route`, so the AST dispatcher sees the read call first.

  I verified this locally:

  ```text
  DispatchEntry(manager_attr="traffic_route_manager", method="get_traffic_route_details")
  ```

  The API action endpoint would therefore try to dispatch update args into the read method and fail. Please add a `DISPATCH_OVERRIDES` entry mapping `unifi_update_traffic_route` to `("traffic_route_manager", "update_traffic_route")`.

- [ ] Add a regression test in `apps/api/tests/test_actions_dispatcher.py` proving `unifi_update_traffic_route` dispatches to `TrafficRouteManager.update_traffic_route` with the widened fields.

  The existing action-dispatcher tests still pass because this tool is not covered there. The test should cover at least one new field such as `target_devices` or `domains`, not only the existing `enabled` path.

## Verified Locally

```text
uv run pytest apps/network/tests/unit/test_traffic_route_tools.py -q
# 7 passed

uv run pytest packages/unifi-core/tests/network/models/test_traffic_routes.py -q
# 19 passed

uv run pytest apps/api/tests/test_actions_dispatcher.py -q
# 32 passed, but missing this regression
```

The feature direction looks good and the MCP-side tests are useful; this blocker is specifically about the REST/API action dispatch surface staying aligned with the new preview-fetching tool body.

sirkirby (commented):
Additional required pre-merge validation: this also needs targeted live smoke output for the widened traffic-route update path.

The unit tests are useful, but they only prove the mock/tool path. Because this changes an API-backed mutation that writes full traffic-route objects, I need live controller evidence before merge.

Please add maintainer-verifiable live output covering one of these paths:

1. Preferred: read an existing disposable traffic route, preview one widened field (`target_devices`, `domains`, `ip_addresses`, `ip_ranges`, `regions`, or `next_hop`), apply a low-risk reversible update, re-read to prove it persisted, then revert and re-read again.
2. If there is no safe disposable traffic route: run live read + confirm=false preview for a real route, and explicitly document why confirm=true mutation was deferred.

After the dispatcher override is fixed, the API action path should also be exercised for the same tool so we know the REST/action surface is not bypassing the MCP tool behavior.

ljmerza:
Thanks for the careful review — all three items addressed in `85de3ad`.

## 1. Dispatch override (blocker)

Added the `DISPATCH_OVERRIDES` entry so the action dispatcher pins
`unifi_update_traffic_route` to the mutation method instead of the
preview-fetch read it now AST-resolves to:

```python
# update_traffic_route now pre-fetches the route via get_traffic_route_details
# to render a current-vs-proposed preview, so the AST walker captures the read
# method first. Pin dispatch to the mutation method.
"unifi_update_traffic_route": ("traffic_route_manager", "update_traffic_route"),
```

Same pattern as the sibling `unifi_toggle_traffic_route` / `unifi_update_wlan`
entries. No `DISPATCH_ARG_TRANSLATORS` entry is needed: the manager signature is
`update_traffic_route(route_id, enabled=None, **kwargs)`, so the widened fields
forward straight through, and `confirm` is a separate `dispatch_action` arg (not
in `args`).

Resolved table now:

```text
DispatchEntry(manager_attr="traffic_route_manager", method="update_traffic_route")
```

## 2. Action-dispatcher regression test (blocker)

`apps/api/tests/test_actions_dispatcher.py` —
`test_dispatch_update_traffic_route_routes_to_mutation_with_widened_field` builds
the **real** `build_dispatch_table()` (so it exercises the override wiring, not
just arg forwarding) and asserts both the resolution and a widened field:

```python
factory.get_domain_manager.assert_awaited_once_with(
    session=session, controller_id="cid", product="network",
    attr_name="traffic_route_manager")
domain_manager.update_traffic_route.assert_awaited_once_with(
    route_id="tr1", target_devices=[{"type": "CLIENT", "client_mac": "aa:bb:cc:dd:ee:ff"}])
```

Plus a spot-check line in `test_dispatch_overrides_specific_targets`.

Test runs:

```text
uv run --package unifi-api-server pytest apps/api/tests/test_actions_dispatcher.py -q
# 33 passed

uv run --package unifi-api-server pytest apps/api/tests -q
# 883 passed

uv run --package unifi-network-mcp pytest apps/network/tests/unit/test_traffic_route_tools.py -q
# 7 passed

make manifest   # no diff (dispatcher-side only)
make format / make lint   # clean
```

## 3. Live controller evidence

Added a committed approved-phase lifecycle `lifecycle_network_traffic_route()` to
`scripts/live_smoke.py` (wired into `run_approved()`), plus a widened
preview-args entry. Traffic routes can't be created via MCP (they need a VPN
policy), so it exercises the widened **`domains`** field against an existing
route in a fully reversible way — read → preview → apply → re-read → revert →
re-read — pinnable via `UNIFI_SMOKE_TRAFFIC_ROUTE_ID`. (Network-specific values
below are redacted.)

### MCP-direct (PR-branch tool layer)

```text
approved:get            ok       unifi_get_traffic_route_details
approved:preview        ok       unifi_update_traffic_route   Will update domains on traffic_route '<route>'. Set confirm=true to execute.
approved:update         ok       unifi_update_traffic_route   Traffic route '<route>' updated: domains.
approved:verify         ok       unifi_get_traffic_route_details
approved:verify         ok       unifi_update_traffic_route   probe domain codex-smoke-<ts>.example.com present on '<route>' after update
approved:revert         ok       unifi_update_traffic_route   Traffic route '<route>' updated: domains.
approved:verify-revert  ok       unifi_get_traffic_route_details
approved:verify-revert  ok       unifi_update_traffic_route   probe domain removed; '<route>' domains restored to original entries
FAILED records: 0
```

(First pass also confirmed end-to-end validation: a throwaway `.invalid` domain
was correctly rejected by the controller with `api.err.InvalidDomain`, i.e. the
widened `domains` payload reaches the controller and is validated. Switched the
probe to the IANA documentation domain `example.com`.)

### REST / action surface

After the override, exercised `POST /v1/actions/unifi_update_traffic_route`
against a locally-booted `unifi-api-server` registered to a live controller —
this is what proves the REST surface routes to the mutation method rather than
`get_traffic_route_details`:

```text
[get]            http=200 domains=[<existing>]
[update]         http=200 success=True        # POST /v1/actions/unifi_update_traffic_route (domains += probe)
[verify]         http=200 probe_present=True
[revert]         http=200 success=True
[verify-revert]  http=200 probe_absent=True
REST EVIDENCE PASSED=True
```

The route was restored to its original `domains` and independently re-read to
confirm no probe entry remained.

sirkirby (approved):
Independent validation is complete — both blockers from the prior review are resolved and verified, and the widened-update path holds up against live hardware. Approving.

## Blockers — resolved & verified

**1. Dispatch override.** `DISPATCH_OVERRIDES` now pins `unifi_update_traffic_route → ("traffic_route_manager", "update_traffic_route")`, matching the sibling toggle/update entries. I confirmed the override is actually load-bearing: with it removed, dispatch AST-resolves to the read method `get_traffic_route_details` and the action path can't await the mutation.

**2. Regression test.** `test_dispatch_update_traffic_route_routes_to_mutation_with_widened_field` builds the real `build_dispatch_table()` and asserts both the resolution and that a widened field (`target_devices`) reaches the manager. I verified it genuinely guards the override — it fails when the override is removed — so it's a real regression guard, not a tautology.

## Gates

- `make lint` clean
- 33 dispatcher + 7 tool + 19 core-model tests pass
- No generated-artifact drift

## Independent live smoke (maintainer hardware)

- **Field-name check:** all six widened fields (`target_devices`, `domains`, `ip_addresses`, `ip_ranges`, `regions`, `next_hop`) are real keys on live V2 route objects. This is the one structural risk in widening — the manager merges kwargs by key into the full route payload, so a wrong name would silently inject a bogus key instead of replacing. Confirmed they all match.
- **Reversible mutating cycle** against an existing DOMAIN-type route: read → `confirm=false` preview → apply (append a throwaway `example.com` probe domain) → re-read (probe present) → revert → re-read (probe gone, original entries restored). Field preservation held; 0 failed records.
- **`target_devices`** validation exercised via the preview path.

For the record: only `domains` received a live *write*; the other five forward through the identical manager merge path and are all confirmed real route keys, so the structural risk is covered without mutating each field on a production route.

Nice work on the smoke harness addition — the read→apply→verify→revert→verify lifecycle with a re-read assertion is exactly the right shape for a manager that doesn't verify-after-write.

LGTM.

PR #387, opened 2026-06-27
fix(wlan): accept networkconf_id alias in create_wlan and update_wlan

### LINKED ISSUE #386: bug(wlan): create_wlan silently ignores networkconf_id — SSIDs always land on Default network
## Summary

`unifi_create_wlan` accepts a `networkconf_id` parameter but silently drops it — every SSID created via the MCP is attached to the Default network regardless of the value passed.

## Environment

| Field | Value |
|-------|-------|
| Package | `unifi-network-mcp@latest` via `uvx` |
| Controller | UCG Ultra |
| UniFi OS | 5.1.19 |
| Auth method | API key (`X-API-Key`) |
| API base path | `/proxy/network/api/s/default/` |

## Steps to reproduce

```python
create_wlan(
    name="MySSID",
    passphrase="password1",
    security="wpapsk",
    networkconf_id="<id-of-non-default-network>",
    ap_group_ids=["<ap-group-id>"]
)
```

## Observed

SSID is created successfully and appears in the controller, but is attached to the Default network. The `networkconf_id` field is either silently dropped from the POST body in the MCP layer, or the legacy `/rest/wlanconf` endpoint ignores it on creation.

## Expected

SSID should be created attached to the network specified by `networkconf_id`.

## Workaround

After creation, immediately PATCH the WLAN via a direct REST `PUT` to `/proxy/network/api/s/{site}/rest/wlanconf/{wlan_id}` using API key auth, with a full config payload that includes `networkconf_id`. (Note: cookie/session auth returns 403 on this write path.)

```python
import requests, urllib3
urllib3.disable_warnings()
s = requests.Session()
s.verify = False
s.headers.update({"X-API-Key": API_KEY})
data = s.get(f"https://{HOST}/proxy/network/api/s/{SITE}/rest/wlanconf/{wlan_id}").json()["data"][0]
data["networkconf_id"] = target_network_id
s.put(f"https://{HOST}/proxy/network/api/s/{SITE}/rest/wlanconf/{wlan_id}", json=data)
```

## Proposed fix

Two options:

1. **Verify the POST payload** — confirm `networkconf_id` is included in the body sent to `/rest/wlanconf` on creation. If it's being filtered out in the MCP layer, add it to the allowed fields.
2. **Post-create PATCH** — if the legacy create endpoint simply doesn't accept `networkconf_id` on initial POST (controller limitation), issue an automatic follow-up `PUT` after the create succeeds, similar to the workaround above.

## Notes

- This may share a root cause with a related issue where `update_wlan` also fails to accept `networkconf_id` (returns "Update data is effectively empty or invalid") — both suggest `networkconf_id` is not in the allowed fields list for WLAN operations.
- The direct REST `PUT` workaround confirms the controller *does* accept `networkconf_id` on this endpoint — it's not a controller limitation.

sgubler:
## Summary

Both `unifi_create_wlan` and `unifi_update_wlan` accept a `networkconf_id` parameter in their documentation and descriptions, but the value was silently dropped before reaching the controller — causing all SSIDs to land on the Default network regardless of the parameter passed.

**Root cause:** The shared `Wlan` pydantic model normalises the controller field `networkconf_id` to the model field `network_id` on ingest. Both tool input filters check model field names only, so a caller passing `networkconf_id` (as the docs say to) had it discarded before the model was built.

- `create_wlan`: input was filtered against `WLAN_MUTABLE_FIELDS` (which contains `network_id`, not `networkconf_id`) before the pydantic model was constructed
- `update_wlan`: `to_controller_update` filtered to `MUTABLE_FIELDS` keys only — `networkconf_id` was not one of them, producing an empty dict and the error `"Update data is effectively empty or invalid."`

Closes #386.

## Changes

- **`packages/unifi-core/src/unifi_core/network/models/wlans.py`** — `to_controller_update` normalises `networkconf_id` → `network_id` before filtering, when `network_id` is not already present. Fixes `update_wlan`.
- **`apps/network/src/unifi_network_mcp/tools/network.py`** — `create_wlan` applies the same explicit guard before the `WLAN_MUTABLE_FIELDS` filter. Uses an explicit condition (not a dict comprehension) so `network_id` always wins when both names are passed, regardless of insertion order.
- **`packages/unifi-core/tests/network/models/test_wlans.py`** — model-layer tests for the alias and both-names-passed precedence case
- **`apps/network/tests/unit/test_network_tools.py`** — tool-layer tests covering alias and model-name paths for both `create_wlan` and `update_wlan` (confirm + preview)

Both caller-side names now work; the controller payload always uses `networkconf_id`.

## Type of change

- [x] `fix:` bug fix

## Scope

- [x] Network MCP server
- [x] Shared packages

## Checklist

## Testing

```
cd apps/network && make pre-commit
# 801 passed
```

## Notes for reviewers

The fix is intentionally minimal — two explicit alias-normalisation guards, one in `to_controller_update` (shared model layer) and one in `create_wlan` (tool layer). No schema changes, no manifest regeneration needed (tool signatures are unchanged).

The same aliasing pattern exists for `vlan`/`vlan_id` but is out of scope for this PR.

Reproduced on UCG Ultra running UniFi OS 5.1.19 with `unifi-network-mcp` via `uvx`.

### DIFF
--- a/apps/network/src/unifi_network_mcp/tools/network.py
+++ b/apps/network/src/unifi_network_mcp/tools/network.py
@@ -1193,7 +1193,9 @@ async def create_wlan(
     from unifi_core.network.models.wlans import Wlan as WlanModel
 
     try:
-        wlan_model = WlanModel(**{k: v for k, v in wlan_data.items() if k in WLAN_MUTABLE_FIELDS})
+        # Accept networkconf_id as an alias for network_id (controller name vs model name)
+        normalized = {("network_id" if k == "networkconf_id" else k): v for k, v in wlan_data.items()}
+        wlan_model = WlanModel(**{k: v for k, v in normalized.items() if k in WLAN_MUTABLE_FIELDS})
     except Exception as exc:
         return {"success": False, "error": f"Invalid WLAN data: {exc}"}
     validated_data = wlan_to_create(wlan_model)
--- a/apps/network/src/unifi_network_mcp/tools/network.py
+++ b/apps/network/src/unifi_network_mcp/tools/network.py
@@ -1193,9 +1193,12 @@ async def create_wlan(
     from unifi_core.network.models.wlans import Wlan as WlanModel
 
     try:
-        # Accept networkconf_id as an alias for network_id (controller name vs model name)
-        normalized = {("network_id" if k == "networkconf_id" else k): v for k, v in wlan_data.items()}
-        wlan_model = WlanModel(**{k: v for k, v in normalized.items() if k in WLAN_MUTABLE_FIELDS})
+        # Accept networkconf_id as an alias for network_id (controller name vs model name).
+        # Only remap when network_id is not already present — mirrors to_controller_update behaviour.
+        if "networkconf_id" in wlan_data and "network_id" not in wlan_data:
+            wlan_data = {**wlan_data, "network_id": wlan_data["networkconf_id"]}
+            wlan_data = {k: v for k, v in wlan_data.items() if k != "networkconf_id"}
+        wlan_model = WlanModel(**{k: v for k, v in wlan_data.items() if k in WLAN_MUTABLE_FIELDS})
     except Exception as exc:
         return {"success": False, "error": f"Invalid WLAN data: {exc}"}
     validated_data = wlan_to_create(wlan_model)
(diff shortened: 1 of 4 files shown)


sirkirby (commented):
This is looking right. I confirmed the reported failure mode against the current code, and the alias fix is scoped the way I would expect: callers can use `networkconf_id`, while the controller payload still gets the controller field name.

I'm not approving yet because final approval for this mutation path needs the full live-smoke gate, not just a targeted controller probe. No code changes requested from this review; I'll hold approval until we run that full smoke pass.

sirkirby (approved):
This looks good. I ran maintainer-side live smoke against the branch and the WLAN create/update/delete lifecycle passed, including the update path this fixes. The remaining smoke noise I saw is unrelated to this WLAN change, so I am not going to hold this PR on it.

PR #389, opened 2026-06-27
fix(network): mark mdns_enabled read-only; surface informative error on OS 5+

### LINKED ISSUE #388: bug(network): mdns_enabled silently ignored by controller on UniFi OS 5+ — update_network reports wrong error
## Summary

`unifi_update_network` lists `mdns_enabled` as a writable field, but on UniFi OS 5+ the controller accepts `PUT /rest/networkconf/{id}` with `mdns_enabled` set, returns HTTP 200, and then silently ignores the value — re-fetching the network immediately shows the original value unchanged.

## Environment

| Field | Value |
|-------|-------|
| Package | `unifi-network-mcp` via `uvx unifi-network-mcp@latest` |
| Controller | UCG Ultra |
| UniFi OS | 5.1.19 |
| Auth method | API key (`X-API-Key`) |

## Steps to reproduce

```python
update_network(
    network_id="<network-id>",
    update_data={"mdns_enabled": False}
)
```

## Observed

As of 0.20.7 (verify-after-write from #371), the manager correctly detects non-persistence and returns an error: `"controller did not persist: mdns_enabled"`. However:

1. The field is still advertised as writable in the `unifi_update_network` tool description
2. The error arrives after a wasted round-trip to the controller
3. No guidance on *why* it fails or what to do instead

Before 0.20.7 the failure was completely silent (success returned despite no change).

## Expected

Either:
1. The field is rejected immediately with a clear error explaining the OS 5+ controller limitation
2. Or it works — the controller actually persists the value

## Context

On UniFi OS 5, mDNS appears to be controlled at the site level via:

```
GET/PUT /proxy/network/api/s/{site}/rest/setting/{mdns_id}
```

Key fields on that object: `key: "mdns"`, `enabled_for`, `enabled_for_network_ids`. Attempts to set `enabled_for_network_ids` via this endpoint were also rejected in testing — it's unclear whether per-network mDNS control is possible via the API at all on OS 5, or whether it is UI-only.

## Proposed fix

Mark `mdns_enabled` as read-only in the Network model with an explanatory description, and surface a specific, informative error when a caller attempts to update a read-only field (rather than the generic "No valid mutable fields provided for update"). Remove the field from the writable field list in the tool description.

A future PR could implement the site-level mDNS API if valid values for `enabled_for_network_ids` can be determined.

sgubler:
## Summary

- `mdns_enabled` is marked read-only in the Network model — the controller silently ignores it on the per-network write path on UniFi OS 5+
- `unifi_update_network` now returns a specific, actionable error naming which fields are read-only when they're the only thing passed, instead of the generic "No valid mutable fields provided for update"
- `mdns_enabled` removed from the writable field list in the tool description

## Type of change

- [x] `fix:` bug fix

## Scope

- [x] Network MCP server
- [x] Shared packages

## Background

As of 0.20.7 (verify-after-write from #371), calling `unifi_update_network` with `{"mdns_enabled": False}` would attempt the write, get HTTP 200, re-read the network, detect the field didn't change, and return `"controller did not persist: mdns_enabled"`. That's better than silent success, but it still wastes a controller round-trip and gives no explanation.

On UniFi OS 5+, per-network mDNS appears to be controlled at the site level via `GET/PUT /rest/setting/{mdns_id}` rather than the per-network networkconf endpoint. Attempts to set `enabled_for_network_ids` via that endpoint were also rejected in our testing, so it's unclear whether per-network mDNS control is possible via the API at all. A future PR could explore the site-level API further.

Closes #388.

## Checklist

## Testing

```
cd apps/network && make pre-commit
# 801 passed
```

Four test gaps were closed after the initial implementation commit:

- **Model layer**: `to_controller_update({"mdns_enabled": False}) == {}` pins the actual filtering function, not just field-set membership — catches a regression if the function logic changes independently of the field sets
- **Preview mode**: `confirm=False` with a read-only field returns the rejection error immediately, not a misleading preview — pins that the read-only check fires before the preview gate
- **Code path separation**: passing a completely unknown field produces the generic "No valid mutable fields" error, not the read-only error — the two empty-payload paths must stay distinct
- **`updated_fields` accuracy**: when `mdns_enabled` is mixed with a valid field, it is absent from `result["updated_fields"]` — derived from `validated_data.keys()`, which must not contain the read-only field

## Notes for reviewers

No manifest regeneration needed — the tool signature is unchanged (no parameters added or removed). The field still appears in `unifi_list_networks` / `unifi_get_network_details` read responses so callers can observe the current mDNS state; it simply can't be changed via this tool on OS 5+.


sirkirby (changes requested):
This direction looks right, and marking `mdns_enabled` read-only is the honest behavior for the current per-network write path. I do need a couple of cleanup changes before approval:

- Please regenerate and commit the Network tool manifest. The decorator description removes `mdns_enabled`, but the committed `apps/network/src/unifi_network_mcp/tools_manifest.json` still advertises it as writable for `unifi_update_network`. `make -C apps/network manifest` should produce the needed diff.
- Please also remove `mdns_enabled` from the long `update_network` function docstring so the source docs do not keep listing it as an allowed update field.

One sequencing note: if #387 merges first, this branch will need a small rebase in `apps/network/tests/unit/test_network_tools.py` because both PRs append test classes at the same spot. Final approval will also need the full live-smoke gate after the branch is updated.

sgubler (commented):
Thanks for the thorough review.

- Manifest regenerated and committed — `mdns_enabled` no longer appears as writable in `tools_manifest.json` for `unifi_update_network`
- `mdns_enabled` also removed from the `update_network` function docstring (it was still in the long field list there)
- Noted on the sequencing. If #387 merges before this, I'll rebase the test file to avoid the append conflict at `TestUpdateNetworkReadOnlyFields`.

sgubler:
Rebased onto main after #387 merged — conflict in `test_network_tools.py` resolved by appending `TestUpdateNetworkReadOnlyFields` after the WLAN test classes. 805 tests passing. Ready for the live-smoke gate.

sirkirby (approved):
Thanks for the update. This addresses the review blockers: the long update_network docstring no longer advertises mdns_enabled as writable, and the generated network tools manifest is now in sync.

I re-reviewed the current diff, regenerated the network manifest locally with no resulting diff, and ran the focused model/tool tests. I also ran a targeted maintainer live controller check through the branch-local smoke runner: listed a real network, then called unifi_update_network with only mdns_enabled in both preview and confirm modes. Both paths rejected immediately with the new read-only error before any controller write path.

CI is green on the updated branch. Approved.

PR #392, opened 2026-06-28
fix: support superHdr and fix boolean handling in protect hdr_mode

cd1zz:
## Problem

`protect_update_camera_settings` advertised `hdr_mode` as a boolean and passed the raw value straight to uiprotect's `Camera.set_hdr_mode(mode: Literal["auto", "off", "always"])`. Two bugs resulted:

1. **No way to request the highest-quality HDR (`superHdr`)**, even though the controller and uiprotect support it via the `"always"` mode (which sets ISP `HDRMode.ALWAYS_ON` / `"superHdr"`).
2. **Passing `false` did not turn HDR off.** `set_hdr_mode(False)` falls through to the `else` branch because `False == "off"` is `False` in Python, so it enables superHdr. In effect *both* `true` and `false` enabled HDR.

## Fix

- Add `normalize_hdr_mode()` in `camera_manager.py`, mapping booleans and friendly aliases to uiprotect's literals:
  - `off` ← `False`, `"off"`, `"none"`, `"disabled"`
  - `auto` ← `True`, `"auto"`, `"on"`, `"normal"`
  - `always` ← `"always"`, `"super"`, `"superHdr"`, `"super_hdr"`
- Use it in **both** the preview and apply paths, so the preview reflects what will actually be applied and invalid values raise before any mutation.
- Update the `protect_update_camera_settings` schema description and regenerate `tools_manifest.json`.
- Add unit tests (alias table, bool handling regression, superHdr apply, invalid value).

## Testing

- `uv run --package unifi-protect-mcp pytest apps/protect/tests/unit/test_cameras.py` → **84 passed**
- `ruff check` clean on changed files
- `tools_manifest.json` regenerated via `generate_tool_manifest.py` (only the hdr_mode description line changes)

Verified against a live NVR (UDM, Protect) across AI Pro, AI Turret, G6 Dome, and G6 Instant cameras: `{"ispSettings":{"hdrMode":"superHdr"}}` is accepted and persists.

### DIFF
--- a/apps/protect/src/unifi_protect_mcp/tools/cameras.py
+++ b/apps/protect/src/unifi_protect_mcp/tools/cameras.py
@@ -215,7 +215,9 @@ async def protect_update_camera_settings(
         Field(
             description=(
                 "Dictionary of settings to update. Supported keys: "
-                "ir_led_mode (auto, on, autoFilterOnly), hdr_mode (true/false), "
+                "ir_led_mode (auto, on, autoFilterOnly), "
+                "hdr_mode (off, auto, or always; always = superHdr highest quality. "
+                "Booleans accepted: true=auto, false=off), "
                 "mic_enabled (true/false), mic_volume (0-100), "
                 "status_light_on (true/false), speaker_volume (0-100), "
                 "name (string), motion_detection (true/false)."
--- a/apps/protect/src/unifi_protect_mcp/tools_manifest.json
+++ b/apps/protect/src/unifi_protect_mcp/tools_manifest.json
@@ -4998,7 +4998,7 @@
               "type": "boolean"
             },
             "settings": {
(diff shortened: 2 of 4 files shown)


sirkirby (approved):
Reviewed and validated end-to-end. Correct diagnosis: `set_hdr_mode(False)` fell through to the `else` branch (`False == "off"` is `False`), so it set `hdr_mode=True` + ISP `ALWAYS_ON` — both `true` and `false` enabled super HDR. The `normalize_hdr_mode()` mapping fixes this and unlocks `always`/superHdr.

Verification performed:
- Unit tests: 84 passed
- Ruff clean; `tools_manifest.json` regenerates with zero drift
- Model wiring confirmed: `hdr_mode` is `Optional[str]` and mutable, so booleans and string aliases survive `to_controller_update` into the manager normalizer
- Independent live validation against the reference NVR (13 HDR-capable cameras), with a fresh per-step re-fetch to prove persistence and original mode restored afterward:
  - `superHdr` → ISP `ALWAYS_ON`, persists ✔
  - `false` → HDR genuinely off (legacy bool `False`, not super) ✔
  - `true` → auto (HDR on, ISP normal) ✔
  - invalid value → predictable error, no mutation ✔

Clean, well-tested, no blockers.

PR #393, opened 2026-06-28
feat: add protect_toggle_rtsp tool to enable/disable camera RTSP streams

cd1zz:
## Problem

The Protect MCP server can read stream URLs (`protect_get_camera_streams`) but has **no way to enable or disable RTSP** on a camera channel. Turning a stream on/off therefore requires editing the controller directly — the MCP can show you a stream only once you've already enabled it elsewhere.

## What this adds

A new mutating tool **`protect_toggle_rtsp(camera_id, enabled, quality="high")`** (`quality` ∈ high/medium/low), built on the `protect_toggle_recording` golden path:

- **Manager methods** `toggle_rtsp` (preview) and `apply_toggle_rtsp` (confirm) on `CameraManager`. Apply mutates `channel.is_rtsp_enabled` and persists via the diff-based `save_device` flow — the same pattern the manager already uses for `mic_enabled`. It is a **no-op** when the channel is already in the requested state.
- On enable, returns the controller-assigned `rtsp_alias` and the RTSP/RTSPS URLs. These are secret-bearing, so the tool **redacts them at the egress boundary** via the existing response-redaction policy (the `rtsp_alias`/`rtsp_url`/`rtsps_url` keys are already in the sensitive-key set), exactly like `protect_get_camera_streams`.
- `ToggleRtspInput` pydantic model in `_actions.py`, with `quality` constrained to a `Literal["high","medium","low"]`.
- `ToolAnnotations(readOnlyHint=False, destructiveHint=False, idempotentHint=True, openWorldHint=False)`; `permission_category="camera"`, `permission_action="update"`.

### Why the private-API `save_device` path (not `create_rtsps_streams`)

uiprotect's public-API `Camera.create_rtsps_streams()` requires an `X-API-Key`. The default Protect MCP setup authenticates with username/password and has no API key, so the channel-level `is_rtsp_enabled` + `save_device` approach keeps the tool working for the common configuration.

## Testing

- New unit tests: manager preview/apply, no-op when already enabled, unknown-quality error; tool preview, confirm-with-redaction, invalid quality, and error path.
- `uv run --package unifi-protect-mcp pytest apps/protect/tests/unit` → **464 passed**
- `uv run --package unifi-core pytest packages/unifi-core/tests/protect` → **373 passed**
- `ruff check` + `ruff format --check` clean
- `tools_manifest.json` regenerated via `generate_tool_manifest.py`; `apps/protect/docs/tools.md` updated

Verified against a live NVR (UDM, Protect) across AI Pro, AI Turret, G6 Dome, and G6 Instant: toggling `channels[].isRtspEnabled` publishes/unpublishes the stream and the controller assigns the alias as expected.


sirkirby (approved):
Reviewed, fixed up, and validated end-to-end.

The tool itself is well-built: follows the `protect_toggle_recording` golden path, uses the diff-based `save_device` flow, is a proper no-op when the channel is already in the requested state, and redacts the secret-bearing `rtsp_alias`/`rtsp_url`/`rtsps_url` at the egress boundary (confirmed those keys are in the redaction sensitive-set). Adding a dedicated tool is the right call here — it mirrors the existing toggle-tool precedent, the operation is channel-scoped and returns secret-bearing output that wouldn't fit the flat `protect_update_camera_settings` settings dict, and it closes a real read/write gap (streams were readable but not toggleable).

Pushed fixes for the CI failures (cross-cutting API-layer registrations the change was missing):
- Registered `protect_toggle_rtsp` in the camera mutation serializer projections (the serializer registry enforces a projection per tool)
- Added the dispatch override -> `camera_manager.apply_toggle_rtsp`, mirroring `protect_toggle_recording`
- Merged current `main` and regenerated the skill reference tables (Protect now 59 tools)

Verification:
- `apps/api` 884 passed, protect unit 464, core protect 373; ruff check + format clean; no manifest/skill-ref drift
- Live validation against a real NVR (toggling the Low channel, state restored afterward): preview shape, unknown-quality error, enable persists, disable persists with no URL, and — the case the unit tests can't cover — enabling a previously-disabled channel returns the controller-assigned alias and a usable RTSPS URL on the same call.

Minor (non-blocking, left as-is): the URL/host construction duplicates `get_camera_streams`; a shared helper would be a nice future cleanup.

PR #397, opened 2026-06-29
fix(network): persist manual min-rate by coupling preference + enable

### LINKED ISSUE #396: bug(wlan): update_wlan silently drops manual min-rate (missing minrate_setting_preference)
## Summary

`unifi_update_wlan` returns the #372 "did not persist" error when setting `minrate_ng_data_rate_kbps` / `minrate_na_data_rate_kbps`, and the requested rate never lands. Follow-up to #371 / #372 (the verify-after-write that surfaced this).

The root cause is **field coupling**, not the write path. The min-rate data-rate fields are only honored when `minrate_setting_preference` is `manual` **and** the band's `minrate_*_enabled` is true. `update_wlan` sends only the rate field, so on a controller in `auto` rate mode the controller answers `rc: ok` but recomputes the rate away. The #372 verify step then (correctly) flags it as unpersisted — but the error message attributes it to "the legacy `/rest/wlanconf` write path," which is misleading in this case.

## Environment

- `unifi-network-mcp` (latest), lazy registration, stdio
- UniFi Dream Machine Pro (UDM Pro)
- Target: a 2.4 GHz WLAN

## Live evidence (isolates the cause)

Holding everything else constant and varying only the preference / payload shape:

1. **Rate-only write while `minrate_setting_preference: auto`** → not persisted (rate stays at the old value). *(original report)*
2. **Set `minrate_setting_preference` to `manual` + the rate** → persists.
3. **With preference already `manual`, write the rate via the legacy `/rest/wlanconf/{id}` PUT** — tried both the full merged object *and* a minimal `{minrate_ng_data_rate_kbps: N}` payload → **both persist**.

Step (3) shows this is **not** a write-path / payload-shape problem on this controller: the legacy full-object PUT persists fine once the rate mode is `manual`. The only differentiator is `minrate_setting_preference`.

## Proposed fix

When a caller sets `minrate_{ng,na}_data_rate_kbps`, also set `minrate_setting_preference="manual"` and the band's `minrate_*_enabled=true`, without overriding values the caller set explicitly. PR incoming.

ljmerza:
## Summary

`unifi_update_wlan` could not persist a manual 2.4/5 GHz minimum data rate. It sent `minrate_{ng,na}_data_rate_kbps` without `minrate_setting_preference`, so a controller in `auto` rate mode answered `rc: ok` and silently recomputed the rate away. The #372 verify-after-write surfaced this as a "did not persist" error (mis-attributed to the legacy write path).

Closes #396.

## Root cause (live-verified on a UDM Pro)

The min-rate data-rate fields are only honored in **manual** mode with the band enabled. Holding everything else constant and varying only the preference / payload shape:

- rate-only write while `minrate_setting_preference: auto` → not persisted
- set preference to `manual`, then the rate → persists
- with `manual`, **both** the full-object PUT and a minimal `{rate}` payload persist

So this is field coupling, not a write-path / payload-shape bug.

## Change

- New pure helper `_apply_minrate_dependencies(update_data)` in `network_manager.py`: when a `*_data_rate_kbps` is requested, inject `minrate_setting_preference="manual"` and the band's `minrate_*_enabled`, **preserving any value the caller set explicitly**.
- `update_wlan` applies it before the merge/PUT. No tool signature or manifest change.

## Tests

- 6 new cases in `test_network_manager_update_verification.py` (helper unit tests + an `update_wlan` test asserting the coupling reaches the PUT payload).
- The `live_smoke` WLAN lifecycle now issues a rate-only update on its throwaway SSID, so the silent-recompute path is covered by verify-after-write.
- `make core-test` → **1285 passed**; `ruff check` / `ruff format --check` clean.

## Live verification (UDM Pro, through the patched manager)

Reproduced the broken `auto` baseline, ran the patched `update_wlan` with a rate-only payload, then restored:

```
baseline:        {'rate': 6000,  'preference': 'manual', 'ng_enabled': True}
set auto:        {'rate': 1000,  'preference': 'auto',   'ng_enabled': True}   # broken precondition
update_wlan({minrate_ng_data_rate_kbps: 12000}) -> (True, None)
                 {'rate': 12000, 'preference': 'manual', 'ng_enabled': True}   # fix flips to manual + enables, rate persists
restore:         {'rate': 6000,  'preference': 'manual', 'ng_enabled': True}
```

From the `auto` state the rate-only update now persists; pre-fix the same call left `auto` and the rate was recomputed away.

## Note

`create_wlan` has the same latent coupling (a WLAN created with a manual rate but `auto` preference). Left as a follow-up to keep this PR focused on the reported `update_wlan` path.

### DIFF
--- a/packages/unifi-core/src/unifi_core/network/managers/network_manager.py
+++ b/packages/unifi-core/src/unifi_core/network/managers/network_manager.py
@@ -49,6 +49,35 @@ def _unpersisted_fields(before: Dict[str, Any], after: Dict[str, Any], requested
     return stuck
 
 
+# Manual min-rate data-rate fields are honored only when the controller's rate
+# mode is "manual" and the band's min-rate is enabled. Setting a *_data_rate_kbps
+# while minrate_setting_preference is "auto" is answered rc:ok but silently
+# recomputed away — which the verify step above would (correctly) flag as a
+# non-persisted write. Map each rate field to the enable flag it requires.
+_MINRATE_RATE_FIELDS = {
+    "minrate_ng_data_rate_kbps": "minrate_ng_enabled",
+    "minrate_na_data_rate_kbps": "minrate_na_enabled",
+}
+
+
+def _apply_minrate_dependencies(update_data: Dict[str, Any]) -> Dict[str, Any]:
+    """Return update_data with the dependencies a manual min-rate write needs.
+
+    When a caller sets ``minrate_{ng,na}_data_rate_kbps``, also force
+    ``minrate_setting_preference="manual"`` and the band's ``minrate_*_enabled``
+    so the controller persists the requested rate instead of recomputing it
+    under auto mode. Values the caller set explicitly are left untouched.
+    """
+    rate_fields = [field for field in _MINRATE_RATE_FIELDS if field in update_data]
+    if not rate_fields:
+        return update_data
+    patched = dict(update_data)
+    patched.setdefault("minrate_setting_preference", "manual")
+    for rate_field in rate_fields:
+        patched.setdefault(_MINRATE_RATE_FIELDS[rate_field], True)
+    return patched
+
+
 class NetworkManager:
     """Manages network (LAN/VLAN) and WLAN operations on the Unifi Controller."""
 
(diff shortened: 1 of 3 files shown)


sirkirby (approved):
Approving. Root cause is correct and matches the live evidence: the `*_data_rate_kbps` fields are only honored in `manual` mode with the band enabled, so a rate-only write under `auto` is acked then recomputed away.

The placement is the strong point — `_apply_minrate_dependencies` sits inside `manager.update_wlan`, which is the common chokepoint for both the MCP tool and the API dispatch path, so both surfaces are fixed at once. Helper is pure (copies, `setdefault` preserves explicit caller values), and `_unpersisted_fields` reads the raw controller dict so the injected `minrate_setting_preference` is verified end-to-end.

Validation:
- ruff check + format clean; 14/14 unit tests (6 new) pass.
- Independent maintainer live smoke on a UDM Pro: create → hide_ssid update → **rate-only min-rate update** → delete, all verified-after-write. The rate-only update returned ok (verify-after-write confirmed it persisted) where pre-fix it would have reported "did not persist". Throwaway SSID fully cleaned up.
- Rebased onto current main (clean auto-merge over the live_smoke.py overlap).

Pushed one trivial follow-up commit fixing the step-comment numbering in `update_wlan` (two `# 3.`s → `# 3.`/`# 4.`).

PR #399, opened 2026-06-29
fix(setup): merge settings.local.json with python3 instead of sed (fixes #398)

### LINKED ISSUE #398: bug(setup): set-env.sh sed-based JSON merge corrupts settings.local.json when a string value contains ",}" (e.g. regex permission {2,})
## Summary

`scripts/set-env.sh` builds the merged `.claude/settings.local.json` with line-oriented `sed`, which has no concept of JSON string boundaries. Its trailing-comma cleanup rewrites any literal `,}` that happens to appear **inside a string value**, producing invalid JSON. The script then (correctly) aborts and refuses to write — but it can *never* succeed as long as the existing settings file contains such a string, so Claude Code setup is blocked.

Observed with plugin `unifi-protect` v0.6.0 (same `set-env.sh` is shared across targets).

## Root cause

In `write_claude_settings()`:

```sh
# line ~154
existing=$(echo "$existing" | sed 's/,[[:space:]]*}/\n  }/g')
```

This is meant to strip the trailing comma the merge inserts before `}`. But `,}` also occurs legitimately *inside* string values — most commonly in a regex-bearing permission entry. For example an existing permission:

```json
"Bash(grep -inE \"VM[A-Z0-9?]{2,}|...\")"
```

contains the substring `,}` inside the regex quantifier `{2,}`. The `sed` rewrites it to:

```json
"Bash(grep -inE \"VM[A-Z0-9?]{2
  }|...\")"
```

i.e. a raw newline is injected into the string, giving `Invalid control character at ...`. The validation at line ~160 then fails:

```
ERROR: produced invalid JSON. Bad output left at .claude/settings.local.json.tmp.NNN for inspection.
       Original .claude/settings.local.json was not modified.
```

The same class of bug applies to the other `sed` JSON edits in this function (lines ~138-151): key matching, value escaping (`s/&/\\&/g` etc.), and comma handling are all string-unaware.

## Repro

1. Have a `.claude/settings.local.json` whose permission list contains any string with `,}` in it, e.g. a regex permission using `{2,}` / `{1,3}`:
   ```json
   { "permissions": { "allow": ["Bash(grep -inE \"x{2,}\")"] } }
   ```
2. Run:
   ```sh
   bash scripts/set-env.sh --target claude UNIFI_PROTECT_HOST=192.168.1.1 \
     UNIFI_PROTECT_USERNAME=admin UNIFI_PROTECT_PASSWORD=secret
   ```
3. Result: `ERROR: produced invalid JSON ...`. Setup cannot complete until the offending permission is removed.

## Suggested fix

`python3` is already a hard dependency in `write_claude_settings()` (it's used to validate input and output). Do the whole merge in Python instead of `sed`, so string contents are never touched:

```sh
python3 - "$SETTINGS_FILE" "$@" <<'PY'
import json, sys
path, *pairs = sys.argv[1:]
try:
    with open(path) as f: data = json.load(f)
except FileNotFoundError:
    data = {}
env = data.setdefault("env", {})
for p in pairs:
    k, _, v = p.partition("=")
    env[k] = v
with open(path, "w") as f:
    json.dump(data, f, indent=2)
    f.write("\n")
PY
```

This is robust to arbitrary string content (regex permissions, escaped quotes, `&`, etc.) and removes the need for the manual escaping / trailing-comma passes entirely.

## Workaround

Merge the `UNIFI_PROTECT_*` keys into the settings `env` block by hand (or with a small python snippet) instead of running the script.

ebaschiera:
## What

Replace the `sed`-based JSON merge in `set-env.sh` (`write_claude_settings`) with a `python3` merge for the Claude target. Applied identically to all three plugin copies (`unifi-protect`, `unifi-network`, `unifi-access`).

Fixes #398.

## Why

The `sed` merge had no concept of JSON string boundaries. The trailing-comma cleanup

```sh
existing=$(echo "$existing" | sed 's/,[[:space:]]*}/\n  }/g')
```

rewrites any literal `,}` that appears **inside a string value**, not just structural trailing commas. The most common trigger is a regex-bearing permission already present in the user's `settings.local.json`, e.g.:

```json
"Bash(grep -inE \"VM[A-Z0-9?]{2,}|...\")"
```

The `{2,}` quantifier contains `,}`, so `sed` injects a newline into the string and the result fails JSON validation:

```
ERROR: produced invalid JSON. Bad output left at .claude/settings.local.json.tmp.NNN for inspection.
       Original .claude/settings.local.json was not modified.
```

Because the offending string is pre-existing, the script can never succeed until the user manually edits their settings — setup is effectively blocked. The same string-unaware fragility applied to the key-match and value-escaping passes too.

## How

`python3` is already a hard dependency of this code path (it was used to validate input and output). The fix does the entire merge in `python3`: load the existing object, set each `KEY=VALUE` under `env`, and write back with `indent=2`. This is robust to arbitrary string content — regex permissions, embedded quotes, `&`, backslashes — none of which are touched.

The previous `sed` logic is retained only as a fallback for environments where `python3` is unavailable, so behavior on python-less systems is unchanged.

## Testing

`bash -n` passes on all three copies. Manual repro against a `settings.local.json` containing both `{2,}` and `{1,3}` regex permissions, merging a password containing `"`, `,}`, `&`, and `\`:

- Before: `ERROR: produced invalid JSON ...`
- After: valid JSON, existing regex permissions byte-for-byte intact, password value correctly escaped via `json.dump`.

### DIFF
--- a/plugins/unifi-access/scripts/set-env.sh
+++ b/plugins/unifi-access/scripts/set-env.sh
@@ -129,6 +129,64 @@ write_claude_settings() {
     fi
   fi
 
+  tmp_file="${SETTINGS_FILE}.tmp.$$"
+
+  # Preferred path: merge with python3, which understands JSON string
+  # boundaries. The previous sed-based approach corrupted any existing string
+  # value that happened to contain ",}" (e.g. a regex permission like
+  # `Bash(grep -inE "x{2,}")`) by rewriting the trailing-comma cleanup inside
+  # the string.
+  if command -v python3 >/dev/null 2>&1; then
+    if ! python3 - "$SETTINGS_FILE" "$tmp_file" "$@" <<'PY'
+import json
+import sys
+
+settings_file, tmp_file = sys.argv[1], sys.argv[2]
+pairs = sys.argv[3:]
+
+try:
+    with open(settings_file) as handle:
+        data = json.load(handle)
+except FileNotFoundError:
+    data = {}
+
+if not isinstance(data, dict):
+    sys.stderr.write("ERROR: %s does not contain a JSON object.\n" % settings_file)
+    sys.exit(1)
+
+env = data.setdefault("env", {})
+if not isinstance(env, dict):
+    sys.stderr.write("ERROR: \"env\" in %s is not a JSON object.\n" % settings_file)
+    sys.exit(1)
+
+for pair in pairs:
+    key, sep, value = pair.partition("=")
+    if not sep:
+        sys.stderr.write("ERROR: invalid KEY=VALUE argument: %s\n" % pair)
+        sys.exit(1)
+    env[key] = value
+
+with open(tmp_file, "w") as handle:
+    json.dump(data, handle, indent=2)
+    handle.write("\n")
+PY
+    then
+      echo "ERROR: failed to merge values into $SETTINGS_FILE." >&2
+      echo "       Original $SETTINGS_FILE was not modified." >&2
+      rm -f "$tmp_file"
+      exit 1
+    fi
+
+    mv "$tmp_file" "$SETTINGS_FILE"
+    print_values "$@"
+    echo ""
+    echo "Saved to $SETTINGS_FILE"
+    return
+  fi
+
+  # Fallback path (python3 unavailable): best-effort sed-based merge. Note this
+  # cannot safely handle string values containing ",}" and will be rejected by
+  # the validation below in that case.
   if [ -f "$SETTINGS_FILE" ]; then
     existing=$(cat "$SETTINGS_FILE")
   else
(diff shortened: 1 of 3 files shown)


sirkirby (approved):
Approving. Clean fix for a real setup blocker.

Reproduced both directions locally: the old sed merge fails with `ERROR: produced invalid JSON` on an existing `settings.local.json` whose permission list contains a `{2,}` regex (the `,}` substring), which leaves setup permanently blocked; the new python3 merge succeeds, keeps the regex permission byte-for-byte, preserves existing env keys, and correctly escapes a password containing `"`, `,}`, `&`, and `\`.

The removed inline JSON validation lived inside a python3 guard, so the python-less sed fallback behaves exactly as before — the claim of unchanged behavior on python-less systems holds.

Validation:
- All three `set-env.sh` copies are byte-identical; `bash -n` clean; LF endings preserved.
- Project setup smoke suite: 44 passed, 0 failed (including the "refuses malformed / leaves original untouched" cases).
- CI green; rebased onto current main cleanly.

PR #413, opened 2026-07-02
feat(network): add Dynamic DNS management tools

rmaher001:
## Summary

Adds Dynamic DNS management to the Network MCP server. The controller natively manages Dynamic DNS (Settings → Internet → Dynamic DNS, backed by `/rest/dynamicdns`), but there were no tools for it — DDNS entries could only be configured by hand in the web UI.

New tools (provider-agnostic — `service` is a parameter, e.g. `dyndns`, `noip`, `namecheap`, `cloudflare`, `custom`):

- `unifi_list_dynamic_dns`
- `unifi_get_dynamic_dns_entry_details`
- `unifi_create_dynamic_dns`
- `unifi_update_dynamic_dns`
- `unifi_delete_dynamic_dns`

Implementation mirrors the DNS-record golden path:

- `DynamicDnsManager` in `unifi-core` over V1 `/rest/dynamicdns` (`get` implemented via list-and-filter since the endpoint 405s on GET-by-id)
- `DynamicDns` pydantic model with `MUTABLE_FIELDS`/`READ_ONLY_FIELDS` and `from_controller` / `to_controller_create` / `to_controller_update`
- Update uses fetch-merge-put — pass only the fields to change, current values are preserved
- Mutations follow preview-then-confirm
- The provider secret (`x_password`) is redacted on every egress path, including create/update previews and results
- Full API surface in the same PR: Strawberry `DynamicDns` type + `dynamicDns` / `dynamicDnsEntry` query fields, REST `GET /sites/{site_id}/dynamic-dns[/{entry_id}]`, mutation-ack serializer, cross-layer symmetry registration, and regenerated `schema.graphql` / `openapi.json` / reference docs / tool manifest / skill references

Note on tool naming: the list/get stems are deliberately distinct (`dynamic_dns` vs `dynamic_dns_entry_details`) because the GraphQL naming convention derives query field names from tool names and identical stems collide.

## Type of change

- [x] `feat:` new feature

## Scope

- [x] Network MCP server
- [x] API server
- [x] Shared packages

## Checklist

## Testing

- `make ci` green at the branch tip: ruff format/lint clean, `make check-generated` no drift
- `apps/api` 908 passed (includes SerializerRegistry boot, REST route coverage, SDL/openapi/docs drift gates, new GraphQL fixture tests incl. redaction)
- `apps/network` 860 passed (manager, model, and tool tests incl. preview-redaction cases)
- `packages/unifi-core` 1332 passed
- Live-tested against a real controller (Network 10.x on UniFi OS): full create → list → get → update → delete round-trip on a throwaway entry; verified partial update preserves unmentioned fields and `x_password` is redacted in list/get/create outputs

## Notes for reviewers

- New-tool PRs can't be split smaller: the manifest, API projection, REST route, and regenerated artifacts are a single CI-enforced contract, so most of the diff is generated files (`openapi.json`, `schema.graphql`, manifests, reference docs).
- `unifi_get_dynamic_dns_entry_details` normalizes controller `_id` to `id` via `from_controller` so list/get shapes match.
- Groundwork for a follow-up: a VPN-server "alternate address" update tool (independent, ships separately).

### DIFF
--- a/apps/api/docs/graphql-reference.md
+++ b/apps/api/docs/graphql-reference.md
@@ -669,6 +669,26 @@ type DpiStats {
   categories: JSON!
 }
 
+"""A Dynamic DNS provider entry configured on the controller."""
+type DynamicDns {
+  id: ID
+  siteId: String
+  hostName: String
+  service: String
+  server: String
+  login: String
+  xPassword: String
+  interface: String
+  customService: String
+  options: [String!]
+}
+
+"""Paginated page of Dynamic DNS entries."""
+type DynamicDnsPage {
+  items: [DynamicDns!]!
+  nextCursor: String
+}
+
 """A UniFi Protect event row (list + detail share this shape)."""
 type Event {
   id: ID
@@ -1123,6 +1143,14 @@ type NetworkQuery {
 
   """Look up a single DNS record by id."""
   dnsRecord(controller: ID!, id: ID!, site: String! = "default"): DnsRecord
+
+  """
+  List Dynamic DNS provider entries on the given controller/site (paginated).
+  """
+  dynamicDns(controller: ID!, site: String! = "default", limit: Int! = 50, cursor: String = null): DynamicDnsPage!
+
+  """Look up a single Dynamic DNS entry by id."""
+  dynamicDnsEntry(controller: ID!, id: ID!, site: String! = "default"): DynamicDns
 
   """List configured static routes (paginated)."""
   routes(controller: ID!, site: String! = "default", limit: Int! = 50, cursor: String = null): RoutePage!
@@ -2176,6 +2204,8 @@ Read-only access to UniFi Network resources.
 - `dpiApplications: DpiApplicationPage!`  — List DPI applications (paginated).
 - `dpiCategories: DpiCategoryPage!`  — List DPI categories (paginated).
 - `dpiStats: DpiStats`  — DPI stats catalog (applications + categories).
+- `dynamicDns: DynamicDnsPage!`  — List Dynamic DNS provider entries on the given controller/site (paginated).
+- `dynamicDnsEntry: DynamicDns`  — Look up a single Dynamic DNS entry by id.
 - `eventLog: EventLogPage!`  — List recent controller events (paginated).
 - `eventTypes: EventTypes`  — Get the controller's event-type prefix catalog.
 - `firewallGroup: FirewallGroup`  — Look up a single firewall group by id.
--- a/apps/api/docs/openapi-reference.md
+++ b/apps/api/docs/openapi-reference.md
@@ -469,6 +469,34 @@ Fetch a user by listing then filtering — no native get_user method.
 **Returns:** `Detail_DnsRecordModel_`
 
 
+## network/dynamic_dns
+
+### `GET /v1/sites/{site_id}/dynamic-dns` — List Dynamic Dns
+
+
+**Parameters:**
+
+- `site_id` (path) (required)
+- `limit` (query)
+- `cursor` (query)
+- `controller` (query)
+
+
+**Returns:** `Page_DynamicDnsModel_`
+
+### `GET /v1/sites/{site_id}/dynamic-dns/{entry_id}` — Get Dynamic Dns Entry Details
+
+
+**Parameters:**
+
+- `site_id` (path) (required)
+- `entry_id` (path) (required)
+- `controller` (query)
+
+
+**Returns:** `Detail_DynamicDnsModel_`
+
+
 ## network/firewall
 
 ### `GET /v1/sites/{site_id}/acl-rules` — List Acl Rules
--- a/apps/api/openapi.json
+++ b/apps/api/openapi.json
@@ -1377,6 +1377,31 @@
         "title": "Detail[DoorModel]",
         "type": "object"
       },
+      "Detail_DynamicDnsModel_": {
+        "additionalProperties": true,
+        "properties": {
+          "data": {
+            "$ref": "#/components/schemas/DynamicDnsModel"
+          },
+          "render_hint": {
+            "anyOf": [
+              {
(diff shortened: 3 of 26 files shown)


rmaher001:
Filed #416 for the direction discussion per CONTRIBUTING's issue-first guidance (should have preceded this PR — apologies for the ordering). This PR implements item 1 (DDNS); item 2 (VPN alternate-address) is a possible follow-up awaiting direction.

sirkirby (changes requested):
## Hard blockers before merge

I found three issues that need fixing before this can land:

- [ ] **API action dispatch is broken for `unifi_update_dynamic_dns`.** The tool exposes `update_data`, but `DynamicDnsManager.update_dynamic_dns()` expects `entry_data`, and there isn't a `DISPATCH_ARG_TRANSLATORS` entry for this tool. The `/v1/actions/unifi_update_dynamic_dns` path will call the manager with an unexpected `update_data=` kwarg.
  - `packages/unifi-core/src/unifi_core/network/managers/dynamic_dns_manager.py:90`
  - `apps/api/src/unifi_api/services/dispatch_overrides.py`

- [ ] **API GraphQL/REST redaction policy override is not honored.** `DynamicDns.from_manager_output(..., redact_sensitive=...)` supports policy-controlled redaction, but the new GraphQL resolver and REST route call it without passing `ctx.redact_sensitive_fields` / `request.app.state.config.policy.response.redact_sensitive_fields`. That means `UNIFI_API_REDACT_SENSITIVE_FIELDS=false` still redacts `x_password` on these surfaces.
  - `apps/api/src/unifi_api/graphql/resolvers/network.py:2759`
  - `apps/api/src/unifi_api/graphql/resolvers/network.py:2783`
  - `apps/api/src/unifi_api/routes/resources/network/dynamic_dns.py:75`
  - `apps/api/src/unifi_api/routes/resources/network/dynamic_dns.py:127`

- [ ] **Create/update silently drop unknown fields inside the free-form data dicts.** The tests currently assert this behavior (`test_create_drops_unknown_fields`, `test_drops_unrecognised_keys`), but this reintroduces the silent-drop class we added strict-kwargs/additionalProperties protections to avoid. For `entry_data` / `update_data`, please reject unknown or read-only keys with an actionable error instead of filtering them out silently.

Minor but worth fixing while you're in here:

- `unifi_update_dynamic_dns` preview currently uses `current_state={}` instead of fetching the current entry and showing a real before/after preview.

Focused checks I ran locally:

- `uv run --package unifi-network-mcp pytest apps/network/tests/unit/test_dynamic_dns_manager.py apps/network/tests/unit/test_dynamic_dns_tools.py`
- `uv run --package unifi-core pytest packages/unifi-core/tests/network/models/test_dynamic_dns.py`
- `uv run --package unifi-api-server pytest apps/api/tests/graphql/fixtures/network/test_dynamic_dns.py`
- `uv run --package unifi-api-server pytest apps/api/tests/unit/test_cross_layer_symmetry.py apps/api/tests/test_resource_route_coverage.py apps/api/tests/test_serializer_coverage.py apps/api/tests/test_serializer_registry.py apps/api/tests/serializers/test_network_serializers.py`
- `uv run --package unifi-api-server pytest apps/api/tests/graphql/test_sdl_drift.py apps/api/tests/graphql/test_openapi_drift.py apps/api/tests/graphql/test_docs_drift.py apps/api/tests/graphql/test_openapi_docs_drift.py apps/api/tests/graphql/test_redaction_contract.py`
- `uv run ruff check ...` on the touched implementation/test files
- `make check-generated`

Once these are fixed, I'll rerun the focused checks and do the live validation pass on my hardware before merge.

rmaher001:
Thanks for the detailed review — all three blockers and the minor are addressed, and I rebased onto `main` + regenerated the artifacts (the branch is mergeable again).

**Blockers**

- **`unifi_update_dynamic_dns` dispatch:** added create/update arg-translators in `dispatch_overrides.py` so the manager receives the shape it expects. The update preview now pre-fetches the current entry, which made the AST dispatch-builder capture the read method first — pinned dispatch to the mutation method via `DISPATCH_OVERRIDES` (mirroring `unifi_update_traffic_route`) and added a regression test that uses the real `build_dispatch_table()` rather than a hand-built table.
- **Redaction policy:** threaded `redact_sensitive` through the GraphQL resolver and the REST route (matching `Wlan` / `wlans.py`); added policy-off tests on both surfaces.
- **Silent field drop:** create/update now reject unknown or read-only keys with an actionable error (`reject_unknown_fields`) on both the MCP and `/v1/actions` paths, and the old silent-drop tests were replaced with rejection tests.

**Minor:** the update preview now renders a real current-vs-proposed diff instead of an empty `current_state`.

**Also fixed while in here:** `dynamic_dns_manager` wasn't registered in the apps/api manager factory, so the REST route and GraphQL resolver raised `UnknownManager` at request time (the fixture tests stub the manager, which hid it). Registered it, and added a REST route test that goes through the real factory.

Verified the full CRUD cycle end-to-end against a live controller — create/update/get/delete, `x_password` redaction, and unknown-field rejection all behave as expected. Ready for another look.

sirkirby (approved):
Re-reviewed the full Dynamic DNS change set after the follow-up fixes and the maintainer test commit at a4a987d. The prior blockers are resolved: API dispatch now has the needed argument translators, REST/GraphQL redaction honors the request policy instead of hardcoding redaction, unknown/read-only Dynamic DNS fields are rejected through the shared model, and update preview now uses current controller state.

I also rebased the fork branch onto current main and pushed a focused REST detail redaction-policy coverage commit. Verification on my side:

- make check-generated
- focused Dynamic DNS MCP/core/API/dispatch/cross-layer/drift/redaction tests
- live Network controller Dynamic DNS create -> list/get -> reject unknown update field -> update -> get -> delete -> deleted-get lifecycle smoke
- make pre-commit

GitHub CI is green on the updated head. Approved.

PR #414, opened 2026-07-03
feat(network): add unifi_delete_port_forward tool

level99:
## Summary

Adds `unifi_delete_port_forward` — a confirmation-gated, destructive delete for port
forwarding rules. Port forwards were the **only mutating network resource without a delete
tool** (11 siblings — ACL rules, firewall groups/policies, WLANs, DNS records, etc. — already
have one). `FirewallManager.delete_port_forward` has existed since the initial commit but was
never surfaced as a tool, so callers could *disable* a rule via `unifi_toggle_port_forward`
but not *remove* it (GUI-only until now).

No new manager code — the existing method is now exposed through the MCP tool layer and the
API `/v1/actions` path.

## Type of change
- [x] `feat:` new feature

## Scope
- [x] Network MCP server  [x] API server

## What changed

**MCP (`apps/network`):** new `unifi_delete_port_forward` tool —
`permission_category=port_forwards` / `action=delete`, `destructiveHint=True`,
`idempotentHint=True`. Preview (`confirm=false`) uses `preview_response(action="delete")` with
an irreversibility warning; the confirmed path delegates to
`firewall_manager.delete_port_forward`. Thin wrapper, no business logic.

**API (`apps/api`):** register the delete tool's mutation-ack projection (DETAIL) so
`test_serializer_coverage` passes and `/v1/actions` serializes it. Generalized the toggle
dispatch-arg translator into a shared `_rename_port_forward_id_to_rule_id`
(`port_forward_id` -> `rule_id`), now used by both toggle and delete — mirroring the existing
`_rename_mac_address_to_client_mac` shared helper. No Strawberry type / REST route / SDL
change — the serializer is projection config, not schema.

## Files changed

| Area | Files |
|---|---|
| MCP | `tools/port_forwards.py` (new tool) |
| API | `serializers/network/port_forwards.py` (ack projection), `services/dispatch_overrides.py` (shared id translator) |
| Tests | `test_port_forwards.py` (tool), `test_firewall_manager.py::TestDeletePortForward` (manager), `test_actions_dispatcher.py` (translator), `serializers/test_network_misc.py` (ack coverage) |
| Generated | `tools_manifest.json` (180 -> 181), network skill reference |

## Checklist
- [x] Read `AGENTS.md` / `CONTRIBUTING.md`.
- [x] Added/updated tests for the new behavior.
- [x] Regenerated all generated artifacts (`make manifest` -> tool manifest + skill reference).
- [x] No credentials, tokens, controller addresses, or private data (test fixtures use synthetic IDs).
- [ ] `make pre-commit` — `make` isn't available on my host; ran the equivalents individually
      (workspace `ruff check`/`format`, per-package `pytest`, regenerated + drift-checked every
      artifact). Full gate runs in CI.

## Testing

| Environment | |
|---|---|
| **Device** | UniFi Dream Machine Pro Max |
| **UniFi OS** | 5.0.16 |
| **Network application** | 10.1.89 |
| **MCP client** | Claude Code (stdio) |

| Live test (real tool code vs the UDM; rule created + deleted, nothing left behind) | Result |
|---|---|
| Create a disposable **disabled** rule on an unused high port | ✅ |
| Preview (`confirm=false`) — no write; irreversibility warning fired; rule still present on re-read | ✅ |
| Confirmed delete (`confirm=true`) — `deleted successfully` | ✅ |
| Re-read — rule gone from the controller | ✅ |

Unit/static (per package; `uv run pytest` trampoline-fails on my host so the workspace venv is used directly):
- `apps/network` — tool tests (preview / success / manager-failure / not-found / empty-id / generic-exception) + direct manager coverage for `delete_port_forward` (success / not-connected / re-raise) + tool-map sync.
- `apps/api` — cross-layer symmetry + all four `*_drift` gates + serializer coverage + the dispatch-translator test.
- `ruff check` + `ruff format --check` clean workspace-wide; working tree clean after regenerating every artifact.

## Notes for reviewers
- **Preview helper divergence (intentional).** This tool uses `preview_response(action="delete", ...)`,
  whereas the three existing delete tools (`delete_firewall_policy`, `delete_firewall_group`,
  `delete_acl_rule`) call `create_preview(...)`, which hard-codes `action="create"` / `will_create`
  in the response — semantically wrong for a delete. I used the correct helper here rather than copy
  the siblings. Happy to file a follow-up adding a `delete_preview` helper and migrating the three
  siblings if you'd prefer consistency now.
- **`confirm` on `/v1/actions`.** `unifi_delete_port_forward` is not in `CONFIRM_REQUIRED_TOOLS`,
  consistent with every sibling delete tool — the preview-then-confirm gate is an MCP-layer contract;
  the API layer treats `confirm` as caller-supplied. No behavior change vs siblings; flagging for awareness.
- **Shared translator.** `_translate_toggle_port_forward` -> `_rename_port_forward_id_to_rule_id`; the
  body is byte-identical, so toggle behavior is unchanged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,7 +241,7 @@ This is a monorepo with shared packages:
 
 ```
 apps/
-  network/          # UniFi Network MCP server (stable, 180 tools)
+  network/          # UniFi Network MCP server (stable, 181 tools)
   protect/          # UniFi Protect MCP server (beta, 59 tools)
   access/           # UniFi Access MCP server (beta, 34 tools)
   worker/           # Cloudflare Worker gateway + npm CLI
--- a/apps/api/src/unifi_api/serializers/network/port_forwards.py
+++ b/apps/api/src/unifi_api/serializers/network/port_forwards.py
@@ -2,8 +2,8 @@
 
 Phase 6 PR2 Task 22 migrated the read shape (PortForward LIST + DETAIL) to
 a Strawberry type at ``unifi_api.graphql.types.network.port_forward``. Only
-the mutation ack remains here — it covers create/update/toggle for port
-forward rules.
+the mutation ack remains here — it covers create/update/toggle/delete for
+port forward rules.
 """
 
 from typing import Any
@@ -26,13 +26,15 @@ def _get(obj: Any, key: str, default: Any = None) -> Any:
         "unifi_create_simple_port_forward": {"kind": RenderKind.DETAIL},
         "unifi_update_port_forward": {"kind": RenderKind.DETAIL},
         "unifi_toggle_port_forward": {"kind": RenderKind.DETAIL},
+        "unifi_delete_port_forward": {"kind": RenderKind.DETAIL},
     },
 )
 class PortForwardMutationAckSerializer(Serializer):
     """DETAIL ack for port forward mutations.
 
-    ``create_*`` returns a dict (or None); ``update_*`` / ``toggle_*``
-    return ``bool``."""
+    ``create_*`` returns a dict (or None); ``update_*`` / ``toggle_*`` /
+    ``delete_*`` return ``bool`` (the manager result the /v1/actions
+    dispatcher serializes)."""
 
     @staticmethod
     def serialize(obj) -> dict:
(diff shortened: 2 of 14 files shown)


sirkirby (approved):
Looks good. I pushed a small docs follow-up to cover the stale static references alongside the generated tool reference.

Maintainer validation:
- Targeted unit/API checks passed locally for the new tool, dispatch translator, serializer ack, manifest coverage, and tool-map sync.
- CI is green after the docs follow-up.
- Ran a live disposable port-forward cycle on maintainer hardware: created a disabled test rule, verified `confirm=false` returned a delete preview without mutating, confirmed delete with `confirm=true`, and verified the rule was gone afterward.

Approved.

PR #415, opened 2026-07-03
chore: add pre-pr push gate and committed git hooks

rmaher001:
## Summary

Adds an opt-in pre-push quality gate so formatting, lint, and generated-artifact drift are caught locally before a push instead of surfacing as a red PR.

- `make pre-pr` — fast, check-only gate: `format-check` + `lint` + `check-generated`, plus an assertion that the tracked working tree is clean (untracked files are ignored). Fail-closed; no mutations.
- `.githooks/pre-push` — committed hook that runs `make pre-pr` and blocks the push on failure.
- `make install-hooks` — run once after cloning; sets `core.hooksPath=.githooks` (and re-applies the hook's executable bit, since git silently skips a non-executable hook).
- CONTRIBUTING.md documents the setup.

The full test suite intentionally stays in CI — the gate covers only the fast checks that most often slip through (the same drift class that `make check-generated` exists for), keeping the hook cheap enough to leave enabled.

## Type of change

- [x] `chore:` maintenance, dependencies, CI, or configuration

## Scope

- [x] Documentation

(Also touches the root `Makefile` and adds `.githooks/` — repo tooling; no server/package code changes.)

## Checklist

`make pre-commit` not run in full: this change is Makefile/docs/hook-only (no Python or worker code), so the relevant checks were the gate itself plus format/lint/drift, all run as described below.

## Testing

- `make pre-pr` on a dirty tree: correctly fails with the offending `git status --short` output (exit 1)
- `make pre-pr` on a clean tree: passes (format-check, lint, check-generated all green)
- `make install-hooks` then a real `git push`: the hook fires, runs the gate, and the push proceeds only after it passes
- Hook file mode verified `100755` in the index (git silently skips non-executable hooks)

## Notes for reviewers

- Opt-in by design: contributors who never run `make install-hooks` see no behavior change. A single push can bypass with `git push --no-verify`; CI remains the authoritative gate.
- The clean-tree assertion is there to stop half-committed regenerated artifacts from being pushed alongside stale committed ones.

### DIFF
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+# Pre-push gate. Runs `make pre-pr` (format-check + lint + generated-artifact
+# drift + clean-tree) before allowing a push, so drift and formatting misses
+# are caught locally regardless of which editor or tooling is in use. The
+# full test suite runs in CI.
+#
+# Install with `make install-hooks` (sets core.hooksPath=.githooks).
+# Bypass for work-in-progress with `git push --no-verify`.
+set -euo pipefail
+
+repo_root="$(git rev-parse --show-toplevel)"
+cd "$repo_root"
+
+if ! make pre-pr; then
+	echo ""
+	echo "pre-push blocked: 'make pre-pr' failed. Fix the issues above, or run"
+	echo "'git push --no-verify' to bypass (discouraged on shared branches)."
+	exit 1
+fi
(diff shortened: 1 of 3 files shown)


rmaher001:
Closing this one on reflection: it's a contributor-workflow change that per CONTRIBUTING should have started as an issue/discussion rather than a PR, and the repo's existing enforcement direction is CI-side gates rather than local hooks. Happy to reopen or rework as a discussion if there's interest.

PR #420, opened 2026-07-05
feat(access): add device config read/update tools across all surfaces

### LINKED ISSUE #419: feat: expose Access device settings/configs (read + update)
### Component

Access MCP server

### Problem or use case

UniFi Access readers (e.g. G6 Pro Entry) have per-device settings that are first-class in the Access web UI — Devices → reader → Settings covers the voice greeting, sounds, display options, and more. The Access MCP server currently exposes none of this, in either direction:

- **Read**: `access_get_device` / `access_list_devices` reduce a device to id/name/type/online/firmware. The controller's `devices/topology4` response (which the server already fetches) carries each device's `configs[]` — `{key, value, tag, update_time}` entries — but they are dropped by the compact-keys filter and the `AccessDevice` model.
- **Write**: there is no device-settings mutation at all (device mutations stop at reboot).

Concrete agent use case: toggling a reader's voice greeting from an automation (e.g. suppress the greeting while the door is standing open, restore it after) — impossible today even though the UI exposes the setting.

### Proposed solution

Two additions, mirroring the existing device/policy golden paths. All controller behavior below was verified live against a real controller (UNVR, current firmware) by capturing the Access web UI's own requests:

1. **`access_get_device_configs`** (read-only) — return a device's `configs[]` from the already-fetched topology. Entries are tagged by category (`device_setting`, `device_extra`, `hub_action`, `hub_power`, `wiring_state`, `credential`); **`credential`-tagged entries carry live secrets (e.g. `ssh_password`) and must be redacted at egress** per the repo's response-redaction policy. This tool doubles as the safe discovery step: read the real keys on a live device before ever writing one.

2. **`access_update_device_config`** (mutating, preview/confirm) — update device settings the same way the web UI does, on the internal v2 API family the server already uses for reboot/lock/unlock/policy:

   `PUT /proxy/access/api/v2/device/{device_id}/configs?is_camera=<bool>`

   Body is a JSON array of `{key, tag, value}` entries; partial updates verified (a single-entry array succeeds). `is_camera=true` applies to camera-class readers (e.g. G6 Pro Entry); response is the standard `{"code":1,"codeS":"SUCCESS"}` envelope. Example verified round-trip — the reader greeting: `show_entry_greet` (`yes`/`no`), `greeting_text` (preset phrase, e.g. `welcome`/`hello`/`hi`/`none`), `greeting_broadcast_name` (`first_name_only`/`none`), all tag `device_setting`.

   Note: the reverse-engineered hjdhjd/unifi-access lib describes a `/configs` (hub) vs `/settings` (non-hub) split; on current firmware the UI uses `/configs` for both with the `is_camera` query parameter as the discriminator — the live capture is authoritative here.

   Safety posture: preview/confirm like all mutations, and the tool validates keys against the device's current `configs[]` (no writing keys the device doesn't already expose).

Full multi-surface treatment (GraphQL + REST + regenerated artifacts) in the same PR, per the repo's serializer-registry CI gate.

### Alternatives considered

- Hand-toggling in the Access app — the status quo; not automatable, and there is no native schedule/door-state gating for reader settings.
- Official Developer API — it has no device-settings surface (unlock/credentials/users/visitors/webhooks only), so the internal v2 API path the server already uses is the only workable route.

### Contribution

- [x] Willing to help implement and test this change (I run Access hardware and can live-verify against a real controller).

rmaher001:
## Summary

Exposes a UniFi Access device's `configs[]` settings — the per-device settings the Access web UI edits (voice greeting, sounds, display, etc.) — which no tool surfaced before. Two new tools, wired through MCP, GraphQL, and REST:

- **`access_get_device_configs`** (read) — returns a device's config entries (`{key, value, tag}`). Credential-tagged and secret-named values are redacted at every egress (MCP tool, GraphQL type, REST route).
- **`access_update_device_config`** (preview/confirm write) — set config keys the device already exposes. Keys are validated against the device's live `configs[]`, each entry's `tag` is looked up from that live config, and credential/secret keys are refused. Writes `PUT device/{id}/configs?is_camera=<bool>` with a partial `{key, tag, value}` array (upsert).

Resolves #419 (opened first, per CONTRIBUTING).

## Layers

- **unifi-core**: `DeviceConfigEntry` model + redaction/validation/`is_camera` helpers; `DeviceManager.get_device_configs` / `update_device_config` / `apply_update_device_config`.
- **apps/access**: the two MCP tools (proxy-only); redaction gated by the response policy.
- **apps/api**: `AccessDeviceConfig` Strawberry type + `deviceConfigs` resolver, `GET /v1/sites/{site}/access-devices/{id}/configs` REST route, serializer + action dispatch, with the redaction policy threaded through both read surfaces. Regenerated `schema.graphql`, `openapi.json`, reference docs, and tool manifests.

## Testing

- Unit: full suites green (`unifi-core` + `apps/access` + `apps/api`), including the multi-surface projection + redaction contract gates.
- Live-verified against a real UNVR (current firmware) on **all three surfaces** — the read returns the device's configs with credential/secret values redacted on MCP, REST, and GraphQL; the write round-trip (`show_entry_greet` yes→no→yes) was confirmed via the action endpoint.

## Notes

- The internal v2 API uses `PUT /proxy/access/api/v2/device/{id}/configs?is_camera=<bool>`; the `?is_camera` query param is the discriminator for camera-class readers vs hubs on current firmware (this differs from the `/configs` vs `/settings` split described by the reverse-engineered `hjdhjd/unifi-access` lib — the `?is_camera` param is what the current web UI uses).
- `is_camera` is auto-derived from the device id (camera-class readers carry a 24-hex Protect-style id), with an explicit override parameter.

### DIFF
--- a/apps/access/src/unifi_access_mcp/tools/devices.py
+++ b/apps/access/src/unifi_access_mcp/tools/devices.py
@@ -5,13 +5,15 @@ Provides tools for listing, inspecting, and rebooting Access hardware devices
 """
 
 import logging
-from typing import Annotated, Any, Dict
+from typing import Annotated, Any, Dict, Optional
 
 from mcp.types import ToolAnnotations
 from pydantic import Field, ValidationError
 
-from unifi_access_mcp.runtime import device_manager, server
+from unifi_access_mcp.runtime import device_manager, server, should_redact_sensitive_fields
 from unifi_core.access.models._actions import RebootDeviceInput
+from unifi_core.access.models.device_configs import from_controller as device_config_from_controller
+from unifi_core.access.models.device_configs import redact_config_entries
 from unifi_core.access.models.devices import from_controller as access_device_from_controller
 from unifi_core.confirmation import preview_response
 from unifi_core.exceptions import UniFiNotFoundError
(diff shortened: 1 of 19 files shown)


sirkirby (approved):
Looks good now. I rebased this onto current `main` and pushed the rebased branch so the required checks reran cleanly.

Validation I ran locally:

- Focused unit/API suites for the touched Access + API paths:
  - `apps/access/tests/unit/test_devices.py`
  - `packages/unifi-core/tests/access/models/test_device_configs.py`
  - `apps/api/tests/graphql/fixtures/access/test_devices.py`
  - `apps/api/tests/routes/resources/test_access_doors_policies.py`
  - API resource/serializer/drift/redaction gates
- `ruff check` over the touched implementation/test files
- `make check-generated`

Live validation on my Access hardware:

- `access_get_device_configs` passed through the live smoke harness.
- Targeted MCP read/update cycle passed: read configs, verified secret redaction, previewed a writable setting update, applied it, verified the changed value, restored the original value, and verified restore.
- Targeted API surface validation passed: REST read, GraphQL read, and `/v1/actions/access_update_device_config` apply/restore all worked, with sensitive config values redacted on read surfaces.

The full Access safe smoke run still reports unrelated/pre-existing Access failures (activity histogram controller error, API-client detail methods, visitor endpoint), but the new device-config read/update path passes on the changed code paths.

PR #447, opened 2026-07-14
fix(protect): send full automations POST body on legacy alarm rule create

rmaher001:
## Problem

`protect_alarm_create_rule` fails with `400 "Failed to parse 'request-body'"` from `POST /proxy/protect/api/automations` whenever the active backend is the legacy automations API. `list_rules` / `get_rule` / `update_rule` / `delete_rule` all work — only creating a new rule fails.

## Root cause

The legacy create path builds its POST body via `alarm_rule_to_legacy_body()`, which emits only the canonical write subset — `name` / `enable` / `sources` / `conditions` / `actions`. The raw automations endpoint requires the full structural envelope, so the body is missing `historyConditions`, `schedules`, `cooldown`, and `isCreatedBySystem`, which is what the controller rejects.

`update_rule` is unaffected because it read-modify-writes the full existing rule (`deep_merge(current_raw, translated)`), so those fields are always present. Create has no base to merge onto and shipped the sparse body straight to the controller.

## Fix

- Add `alarm_rule_to_legacy_create_body()`, which scaffolds the structural fields the canonical write shape can't express — `isCreatedBySystem`, `historyConditions`, `schedules`, `cooldown` (defaults matched to a live Protect 7.0 rule and the `Hovborg/unifi-protect-bridge` `automation_payloads.py` reference) — then overlays the caller's write fields on top.
- `AlarmRulesFacade.create_rule`'s legacy branch uses it; `alarm_rule_to_legacy_body` is left sparse so `update` semantics are unchanged (it keeps merging onto the current rule).

The fix is centralized in `unifi-core`, so every caller that reaches `AlarmRulesFacade.create_rule` shares the corrected legacy payload construction. The generic API action dispatcher has a separate pre-existing `body` → `fields` argument-translation gap; that broader dispatcher issue is outside this PR.

## Tests

- New unit tests assert the actual body sent to the controller (not just which backend is chosen), and specifically cover the `license_plate_known` trigger's `value` round-tripping verbatim through create.
- `unifi-core` and `apps/protect` suites pass; ruff (format + check) clean.

## Verification

Deployed and exercised against a real Protect controller (legacy automations backend): `create_rule` now succeeds, `get_rule` round-trips identically, and `delete_rule` cleans up. No more 400.

### Maintainer verification

I independently validated the branch against my Protect hardware. I forced only backend selection to the legacy facade path while keeping the real controller transport, cloned a controller-accepted non-system rule as disabled, and confirmed create → get → delete with legacy coverage metadata and cleanup. I also ran the full Protect safe phase and the repository pre-commit gate.

<details>
<summary>Targeted legacy create/get/delete output</summary>

```text
protect targeted:create ok: protect_alarm_create_rule
protect targeted:get ok: protect_alarm_get_rule
protect targeted:delete ok: protect_alarm_delete_rule
coverage complete: false
cleanup_confirmed: true
```

</details>

<details>
<summary>Full Protect safe-phase lifecycle output</summary>

```text
protect lifecycle:create ok: protect_alarm_create_rule
protect lifecycle:update ok: protect_alarm_update_rule
protect lifecycle:get ok: protect_alarm_get_rule
protect lifecycle:delete ok: protect_alarm_delete_rule

56 ok, 16 expected skips, 11 correctly held for approval
created resource id matched cleaned resource id
```

</details>

<details>
<summary>Repository gate</summary>

```text
$ make pre-commit
ruff: passed
generated artifact drift: passed
all Python package/app suites: passed
MCP registration modes: passed
Worker tests and typecheck: passed
exit code: 0
```

</details>

### DIFF
--- a/packages/unifi-core/src/unifi_core/protect/managers/alarm_facade.py
+++ b/packages/unifi-core/src/unifi_core/protect/managers/alarm_facade.py
@@ -27,6 +27,7 @@ from unifi_core.protect.models.alarm_rules import (
     alarm_rule_from_controller,
     alarm_rule_from_legacy,
     alarm_rule_to_legacy_body,
+    alarm_rule_to_legacy_create_body,
     alarm_rule_to_v2_body,
 )
 
@@ -88,7 +89,7 @@ class AlarmRulesFacade:
             raw = await self._service.create_rule(body)
             return alarm_rule_from_controller(raw).model_dump(exclude_none=True), True
 
-        body = alarm_rule_to_legacy_body(fields)
+        body = alarm_rule_to_legacy_create_body(fields)
         raw = await self._legacy.create_rule(body)
         return alarm_rule_from_legacy(raw).model_dump(exclude_none=True), False
 
--- a/packages/unifi-core/src/unifi_core/protect/models/alarm_rules.py
+++ b/packages/unifi-core/src/unifi_core/protect/models/alarm_rules.py
@@ -245,6 +245,37 @@ def alarm_rule_to_legacy_body(fields: dict[str, Any]) -> dict[str, Any]:
     return body
 
 
+def alarm_rule_to_legacy_create_body(fields: dict[str, Any]) -> dict[str, Any]:
+    """Serialize canonical write fields to a COMPLETE legacy automations POST body.
+
+    ``POST /proxy/protect/api/automations`` rejects an incomplete body with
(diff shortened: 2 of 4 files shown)


sirkirby (approved):
I independently validated both the normal Protect safe lifecycle and the forced legacy create/get/delete path against my hardware. The legacy path returned complete:false and cleaned up its disposable rule; make pre-commit and CI are green. The payload fix is correctly scoped, and I updated the test labels to follow repository policy.

PR #448, opened 2026-07-14
fix(tool-index): tokenize search so multi-word/LLM queries find tools

ldenaeyer:
## Problem

`get_tool_index` (in `packages/unifi-mcp-shared`, used by all servers) filters the `search`
argument with a **whole-string substring match**:

```python
if search:
    search_lower = search.lower()
    all_tools = [
        t for t in all_tools
        if search_lower in t.get("name", "").lower()
        or search_lower in t.get("description", "").lower()
    ]
```

Because it looks for the entire query string as one substring, any **multi-word query
returns nothing** unless that exact phrase appears verbatim in a single tool's name or
description. This hits LLM/agent callers hardest, since they naturally issue semantic
multi-word queries.

Real-world case that motivated this: an agent searched for
`"update wlan radio tx power"`, `"update tx power"`, and `"wifi update tx power"` to find a
radio-power tool. Every query returned `{"tools": [], "count": 0}`, even though
`unifi_update_device_radio` (whose description covers `tx_power`) exists and is correctly
categorized. The agent concluded the capability did not exist and gave up. Discovery, not
the tool, was the failure.

## Fix

Tokenize the query and rank tools by how many query tokens appear in name + description;
keep any match (best first) with an exact-phrase boost so precise queries stay precise.

- Minimal, dependency-free (`re` only), backward compatible.
- Single-token queries behave as before, just ranked.
- Return shape (`tools`/`count`/`categories`/`filtered`) unchanged.
- One change in the shared package fixes network, access, and protect (they re-export it).

## Notes

- No synonym expansion here on purpose; that is opinionated and better left to callers.
- Separately (not addressed here): several read tools have no permission_category, so
  category-filtered discovery can't find them.

Verified locally: the three previously-empty queries above now return
`unifi_update_device_radio` at the top.


sirkirby (changes requested):
Thanks for the contribution, Lee — the problem statement is clear, and I reproduced the core improvement locally: `unifi_update_device_radio` ranks first for each of the three motivating queries.

## Hard blockers (must fix before merge)

- [ ] **Keep the Worker/cloud discovery path in parity.** `apps/worker/worker/src/relay-object.ts` still applies the old whole-string substring filter in `handleToolIndex`, so the same multi-word searches continue to fail when discovery is served through the Worker. Please mirror the ranking behavior there and add Worker coverage.
- [ ] **Update the public MCP schema description.** `packages/unifi-mcp-shared/src/unifi_mcp_shared/meta_tools.py` still advertises `search` as a "Case-insensitive substring match." That is the schema agents actually read, so it needs to describe the new semantics and be covered by the metadata tests.
- [ ] **Define and test low-signal/tokenless behavior.** Because tokens shorter than two characters are discarded, searches such as `"x"` or whitespace currently skip filtering and return the entire index with `filtered: true` (186 Network tools in my run). That is a regression from the old behavior. Please preserve the prior match behavior or return an empty result when a non-empty search produces no usable tokens.
- [ ] **Bound the result broadening.** The target tool is ranked first, but the three motivating queries return 53, 40, and 44 tools. `"to update firewall"` returns 119 because query tokens are substring-matched inside words such as `tool` and `controller`. Since this filter exists to reduce discovery payload size, please avoid low-signal token matches (for example through haystack tokenization, stop-word handling, a score threshold, or another documented approach) and add a regression assertion for result breadth.
- [ ] **Add behavior tests on both implementations.** The PR changes shared ranking semantics but modifies no tests. At minimum I’d cover multi-word ranking, exact-phrase priority, single-token compatibility, stable tie ordering, no-match/tokenless input, and Worker parity.
- [ ] **Fix the Ruff failure.** CI reports an unsorted import block in `tool_index.py` (`logging` must sort before `re`).

I ran the full shared-package suite: all 337 existing tests pass, which also confirms the new behavior currently has no regression coverage. The branch is now one commit behind `main` after #447 merged, so please rebase while preparing the update.

No controller smoke test is needed for this PR: the change is confined to discovery filtering and does not call a UniFi API.

sirkirby:
Thanks for identifying this, Lee. I independently validated the discovery failure and agree it should be fixed. The implementation needs broader cross-surface changes than the current patch covers, including Worker parity, bounded matching semantics, public schema updates, and regression tests. Since there hasn’t been an update following the requested changes, I’m closing this PR, capturing the validated bug in #478, and will implement it in a maintainer PR. I’ve credited you there for the original report and reproduction.

PR #449, opened 2026-07-14
feat(network): expose 802.11k + roaming-assistant fields on unifi_update_wlan

ljmerza:
## Summary

`unifi_update_wlan` cannot write 802.11k (`rrm_enabled`) or either band's roaming-assistant threshold. These fields are absent from the `Wlan` pydantic model, so `to_controller_update()` filters them out against `MUTABLE_FIELDS` and the tool answers:

```
Update data is effectively empty or invalid.
```

That message reads as a malformed request rather than "this field isn't supported", so the caller has no signal that the field simply isn't modelled. Passing them as top-level args instead fails earlier with `unknown arguments {rrm_enabled, roaming_assistant_6e_rssi}`.

**The write path was never the problem.** `network_manager.update_wlan` deep-merges the update into the existing object and PUTs the whole thing, so these fields only ever needed to reach it. This is a model gap, not a `/rest/wlanconf` limitation.

## Type of change

- [ ] `fix:` bug fix
- [x] `feat:` new feature

## Scope

- [x] Network MCP server
- [ ] Protect MCP server
- [ ] Access MCP server
- [ ] Relay sidecar
- [ ] Cloudflare Worker / CLI
- [x] API server
- [x] Shared packages
- [ ] Plugin packaging
- [x] Documentation

## Change

Five fields added to the `Wlan` model (and `from_controller`), mirrored on the Strawberry type for cross-layer symmetry, and documented on the tool:

| Field | Type | Meaning |
|---|---|---|
| `rrm_enabled` | bool | 802.11k RRM / neighbour reports |
| `roaming_assistant_na_enabled` | bool | 5GHz assistant on/off |
| `roaming_assistant_na_rssi` | int (dBm) | 5GHz kick threshold |
| `roaming_assistant_6e_enabled` | bool | 6GHz assistant on/off |
| `roaming_assistant_6e_rssi` | int (dBm) | 6GHz kick threshold |

Field names and types are read from a live `get_wlan_details` on a **UDM Pro running Network 10.4.57**, not inferred:

```
rrm_enabled: False
roaming_assistant_na_enabled: True    roaming_assistant_na_rssi: -77
roaming_assistant_6e_enabled: True    roaming_assistant_6e_rssi: -88
```

Adding to the model also enables them on `create_wlan`, since `to_controller_create` reads the same `MUTABLE_FIELDS`.

No tool signature change. Two follow-up commits regenerate artifacts that the source change drifts (`tools_manifest.json`; and `schema.graphql` / `graphql-reference.md` / `openapi.json`, each of which has its own CI gate).

## Checklist

## Testing

Ran each `pre-commit` stage individually rather than `make pre-commit` as a whole, because `worker-typecheck` needs `npm ci` for the Cloudflare worker, which is untouched here. Everything else in the target ran:

```
core-test (packages/unifi-core)      1408 passed
apps/api                              942 passed
apps/network                          877 passed
shared-test                           337 passed
relay-test                            111 passed
cross-layer symmetry                   76 passed
ruff check                            All checks passed!
ruff format --check                   870 files already formatted
make manifest / check-generated       clean, regeneration idempotent
```

7 new cases in `test_wlans.py` (`TestRoamingFields`) covering `MUTABLE_FIELDS` membership, `from_controller` reads, survival through `to_controller_update`, `False` not being dropped as if it were `None`, and the create round-trip.

Replaying the exact payload that fails on `main`:

```python
>>> to_controller_update({"rrm_enabled": True, "roaming_assistant_6e_rssi": -70})
{'rrm_enabled': True, 'roaming_assistant_6e_rssi': -70}     # main: {} -> "effectively empty or invalid"
```

The filter's other guarantees still hold: `False` preserved, `None` dropped, unknown keys dropped.

## Notes for reviewers

**No coupling helper, deliberately.** #397 added `_apply_minrate_dependencies` because a rate-only write is live-verified to be recomputed away under `auto`. The analogous question here — does `roaming_assistant_*_rssi` need `roaming_assistant_*_enabled: true` to persist? — is **unverified**. On the controller I have access to both are already `true`, so the failure isn't observable, and I'd rather not ship a dependency injection on the assumption it behaves like minrate. If the coupling is real, #372's verify-after-write reports it as "did not persist" rather than letting it pass silently. Happy to add the helper as a follow-up if someone can reproduce a non-persisting rssi-only write.

**Band coverage.** Only `na` (5GHz) and `6e` (6GHz) exist — the controller exposes no `roaming_assistant_ng`; 2.4GHz sticky-client handling is `min_rssi` on the radio, already writable via `unifi_update_device_radio`.

**Still unwritable on this tool** (out of scope, flagging as a possible follow-up): `bss_transition`, `wpa3_fast_roaming`, `no2ghz_oui`, `minrate_setting_preference`, `dtim_6e`.


sirkirby (changes requested):
Thanks, Leonardo — the implementation itself behaved correctly in independent validation. I ran the full repository pre-commit gate, Network readonly/safe live smoke, and a targeted disposable-WLAN cycle on a lab controller. All five fields were accepted individually on create, a combined update read back exactly, a partial update preserved sibling and unrelated WLAN fields, and all 6 temporary WLANs were deleted.

I have two durable coverage/discoverability gaps to close before merge:

1. Because adding these fields to `Wlan.MUTABLE_FIELDS` also makes them accepted by `unifi_create_wlan`, please add all five to `create_wlan`'s `wlan_data` Field description/docstring and regenerate the Network manifest. Right now create support is real (and advertised in the PR body) but not discoverable from that tool's public schema.
2. Please add boundary-level regression coverage for the new wiring: a Network tool test proving `create_wlan` forwards the new fields, and assertions in the API WLAN serializer/projection test that `Wlan.from_manager_output` maps all five values. The model conversion tests are useful, but they would not catch either boundary being accidentally omitted.

The branch is also one commit behind current main; please rebase/update it after these changes, regenerate artifacts, and let CI rerun.

ljmerza:
Thanks for the thorough validation — and for catching the `create_wlan` gap. You're right that it was real-but-undiscoverable; I noted the create side-effect in the PR body and then never documented it where a caller would actually look.

All three addressed:

**1. `create_wlan` discoverability.** Added the five fields to the `wlan_data` `Field(description=...)` and to the `Optional parameters` list in the docstring, mirroring the wording used on `unifi_update_wlan`. Network manifest regenerated via `make manifest`.

**2. Boundary coverage.** One test per boundary:

- `apps/network/tests/unit/test_network_tools.py::TestCreateWlanRoamingFields::test_roaming_fields_reach_the_manager` — asserts all five reach the manager payload via `create_wlan.call_args[0][0]`.
- `apps/api/tests/serializers/test_network_serializers.py::test_wlan_projection_maps_roaming_fields` — asserts `Wlan.from_manager_output` maps all five out of the raw dict.

Both use `rrm_enabled=False` and negative RSSIs deliberately: a truthiness filter or an `or None` on either path passes with `True`/positive values and silently eats these. I verified the projection test actually fails without the wiring — nulling `rrm_enabled` in `from_manager_output` reds it — rather than just confirming it passes.

**3. Rebased** onto `228f6d8` (it had drifted to 4 behind), artifacts regenerated, CI rerun.

Local: `apps/network` 878, `apps/api` 944, `unifi-core` 1415, ruff clean, `make manifest` idempotent.

sirkirby (approved):
Thanks, Leonardo — this closes the gaps from the requested-changes review.

I verified the updated head (`58ad38f`) against the request:

- `unifi_create_wlan` now documents all five fields in its schema description and docstring.
- The Network boundary test proves all five values reach the manager payload, including `False` booleans and negative RSSI values.
- The API projection test covers all five fields with the same falsey-value cases.
- The branch is rebased on current `main`, generated artifacts are up to date, and all 13 GitHub checks are green.

I also reran the full local `make pre-commit` gate plus Network read-only and safe live smoke on this exact head. Both live phases had zero failures/exceptions, and the safe phase cleaned all three resources it created. Together with the earlier disposable-WLAN cycle that directly exercised these five create/update fields, this is sufficient controller-backed coverage for the revised PR.

Approved — thank you for the thorough follow-up.

PR #453, opened 2026-07-17
feat(network): add unifi_search_device_by_mac cross-site lookup

marcelbohmer:
## Summary

Adds `unifi_search_device_by_mac`, a read-only tool that locates a device across **all** sites by MAC address. UniFi's native search is per-site only — there's no built-in way to find which site adopted a device when you only have its MAC.

It iterates every site from `/api/self/sites` (`system_manager.get_sites()`) and matches each site's `/api/s/{site}/stat/device-basic` list, returning the owning site's name/description plus basic device info (model, type, status). A single unreadable site (e.g. permissions) is logged and skipped rather than aborting the whole search.

Also deduplicates the device state-code map into a module-level `_DEVICE_STATE_MAP` reused by `list_devices` and `get_device_details` (was previously inlined twice).

## Testing

- 3 new unit tests: found-on-later-site (with mixed-case MAC normalisation), not-found, skips-failing-site.
- `pytest apps/network/tests/unit/test_devices_filtering.py` -> 20 passed.
- `ruff check` clean.
- Regenerated `tools_manifest.json` and the network skill reference (186 -> 187 tools).

### DIFF
--- a/apps/network/src/unifi_network_mcp/tools/devices.py
+++ b/apps/network/src/unifi_network_mcp/tools/devices.py
@@ -8,6 +8,7 @@ import logging
 from datetime import datetime, timedelta
 from typing import Annotated, Any, Dict, List, Optional
 
+from aiounifi.models.api import ApiRequest
 from mcp.types import ToolAnnotations
 from pydantic import Field, ValidationError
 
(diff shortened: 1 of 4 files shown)


sirkirby:
Thanks for putting this together and for surfacing the unknown-site lookup scenario. After reviewing the product fit, I do not think we should add this tool.

The Network MCP is site-scoped today, and the existing device list and detail tools already cover MAC lookup within an explicit site. The new behavior is primarily controller-wide site discovery and fan-out, which introduces broader scope, authentication, latency, and partial-result semantics that we have not designed as a general capability.

If we decide to support cross-site discovery, we should start with first-class site enumeration and consistent controller and site scoping across the MCP and API, then determine whether a dedicated device locator is still warranted. I do not want to ask you to rework the implementation against a contract we have not decided to support, so I am closing this rather than requesting implementation changes.

I appreciate the idea and the effort behind the contribution.

PR #463, opened 2026-07-26
fix: translate content filter categories for controller

### LINKED ISSUE #462: Content filter updates send blocked_categories instead of categories
### Problem

`unifi_update_content_filter` publicly accepts `blocked_categories`, and preview mode preserves that field, but confirmed updates pass it unchanged into the V2 controller PUT payload. UniFi rejects the request with HTTP 400 because the controller field is `categories`.

### Reproduction

1. Call `unifi_update_content_filter` with `filter_data={"blocked_categories":[...]}` and `confirm=false`; preview succeeds.
2. Repeat with `confirm=true`.
3. The controller rejects `blocked_categories` as an unrecognized field.

### Expected

The MCP-facing alias remains `blocked_categories`, while the manager translates it to `categories` before fetch-merge-PUT. Unmentioned fields must remain preserved.

### Proposed fix

Normalize both the fetched object and partial update to controller field names in `ContentFilterManager.update_content_filter`, then merge and PUT. Add tool-level preview coverage and manager-level exact-payload coverage.

richardpowellus:
Fixes #462

### DIFF
--- a/apps/network/src/unifi_network_mcp/tools/content_filtering.py
+++ b/apps/network/src/unifi_network_mcp/tools/content_filtering.py
@@ -20,6 +20,9 @@ from pydantic import Field
 from unifi_core.confirmation import create_preview, update_preview
 from unifi_core.exceptions import UniFiNotFoundError
 from unifi_core.network.models.content_filter import (
+    MUTABLE_FIELDS as CF_MUTABLE_FIELDS,
+)
+from unifi_core.network.models.content_filter import (
     from_controller as cf_from_controller,
 )
 from unifi_core.network.models.content_filter import (
@@ -137,8 +140,10 @@ async def update_content_filter(
     if not filter_data:
         return {"success": False, "error": "filter_data cannot be empty"}
 
-    validated_data = cf_to_update(filter_data)
-    if not validated_data:
+    # Keep the preview in the caller-facing dialect; translate to the controller
+    # dialect only on the write path.
+    public_updates = {k: v for k, v in filter_data.items() if k in CF_MUTABLE_FIELDS and v is not None}
+    if not public_updates:
         return {"success": False, "error": "Update data is effectively empty or invalid."}
 
     if not confirm:
@@ -147,11 +152,11 @@ async def update_content_filter(
             resource_id=filter_id,
             resource_name=filter_id,
             current_state={},
-            updates=validated_data,
+            updates=public_updates,
         )
 
     try:
-        merged = await content_filter_manager.update_content_filter(filter_id, validated_data)
+        merged = await content_filter_manager.update_content_filter(filter_id, cf_to_update(public_updates))
         return {
             "success": True,
             "message": f"Content filter '{merged.get('name', filter_id)}' updated successfully.",
--- a/apps/network/tests/unit/test_content_filter_manager.py
+++ b/apps/network/tests/unit/test_content_filter_manager.py
@@ -180,7 +180,35 @@ class TestContentFilterManager:
         assert put_request.method == "put"
         assert put_request.data["name"] == "Family Filter"
         assert put_request.data["safe_search"] == ["GOOGLE", "YOUTUBE", "BING"]
-        assert put_request.data["blocked_categories"] == ["adult", "gambling"]
+        assert put_request.data["categories"] == ["adult", "gambling"]
+        assert "blocked_categories" not in put_request.data
+
+    @pytest.mark.asyncio
+    async def test_update_content_filter_translates_blocked_categories(self, content_filter_manager, mock_connection):
+        """The public blocked_categories alias becomes categories in the PUT payload."""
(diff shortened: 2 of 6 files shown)


sirkirby (changes requested):
I confirmed the underlying #462 diagnosis: the controller returns and accepts categories, while the current confirmed update path sends the MCP-facing blocked_categories alias unchanged. The translation and exact-PUT regression coverage in the first commit are the right shape.

This PR has since expanded into three separate changes: the #462 bug fix, domain-list exposure, and a new content-filter creation tool. The latter two introduce additional repository-wide contracts that are currently failing, including API serializer registration, API action dispatch, generated skill references, formatting, and documentation.

Please scope this PR back to the #462 fix represented by the first commit, retain its focused tests, run formatting and make pre-commit, and update the PR body to only Fixes #462. Please remove Closes #464; I will keep that feature request open and track the API-dispatch and domain-list follow-ups separately.

Once the PR is narrowed and green, I expect the #462 fix to be straightforward to re-review and merge.

richardpowellus:
Thanks for the review. I narrowed this back to the #462 fix from the first commit only: the manager translates the public `blocked_categories` alias to controller `categories` before fetch-merge-PUT, with the focused manager exact-payload and tool preview/confirm regressions retained. The domain-list exposure, creation tool, API/generated artifact changes, and `Closes #464` are removed.

Checks run:
- `make format` — 871 files unchanged
- `uv run --package unifi-network-mcp pytest apps/network/tests/unit/test_content_filter_manager.py apps/network/tests/unit/test_content_filter_tools.py -v` — 24 passed
- `make pre-commit` — passed (Ruff, generated manifests/references, all repository test suites, drift checks, and Worker TypeScript typecheck)

Re-requesting review on the narrowed, green-ready branch.

richardpowellus:
The replacement fork workflow runs are all waiting for upstream approval (`action_required`) before any jobs start. The REST approval endpoint returns `403 Must have admin rights to Repository` for this token. Please approve the pending Actions runs when convenient; I’ll monitor them through completion once started.

sirkirby (approved):
The narrowing is exactly what I asked for — single commit, the #462 fix plus its focused regressions, `Closes #464` dropped. I re-ran everything independently rather than relying on the results in your comment, and it all holds up: Ruff clean, 24/24 content-filter tests, no generated-artifact drift, clean rebase onto `main`, 12/12 CI.

I also validated it against real hardware, and the live behaviour is worth recording because it's more severe than #462 implied. The content-filtering endpoint is strict-schema — it rejects unrecognised fields outright rather than ignoring them:

```
400 JSON parse error: Unrecognized field "blocked_categories" ..., not marked as ignorable
```

So this was never silent data loss. `unifi_update_content_filter` was returning a hard 400 for every category update. Your fix is correct, and I confirmed a full mutating cycle end to end: the alias lands as `categories`, nothing leaks the public name into the payload, and `name`, `enabled`, `network_ids`, `safe_search`, `client_macs`, `allow_list`, `block_list` and `schedule` all survive the fetch-merge-PUT untouched.

I pushed one follow-up commit to your branch rather than sending it back a third time, since this is a placement question rather than anything wrong with the fix.

I moved the translation out of the manager and into `to_controller_update()` in `packages/unifi-core/src/unifi_core/network/models/content_filter.py`. That's where every other Network domain does public-alias to controller-dialect mapping — `wlans` (`network_id` to `networkconf_id`), `vpn` (`type` to `vpn_type`), `traffic_routes` (`name` to `description`), `port_forwards` (`fwd_protocol` to `proto`). Keeping it there leaves `ContentFilterManager` dialect-agnostic, so fetch-merge-PUT doesn't need to translate the profile it just fetched. That also removes the `_to_controller_payload(existing)` call, which was dead code in practice — the controller returns `categories` and never the alias.

Two things that pushed me toward the model layer specifically:

- `to_controller_update()` had a test, `test_passes_blocked_categories`, asserting the alias passed through untranslated. That test encoded the exact bug you were fixing, so the model layer was quietly locking in the old behaviour. It now asserts the rename.
- The manager fetch-merge fixture used `blocked_categories` for the *fetched* profile, which no controller actually returns. It now uses `categories`, matching real hardware.

The tool still previews in the caller-facing dialect and translates only on the write path, so a preview shows `blocked_categories` exactly as the caller wrote it — I kept the intent of your preview test.

Re-verified after the move: Ruff format and lint clean, 1421 unifi-core tests, 887 network tests, 96 API symmetry/serializer tests, and another live mutating cycle through the tool path with the profile restored to baseline afterwards.

One thing I found that is **not** yours and that I'm tracking separately: `schedule_mode` is in `MUTABLE_FIELDS` and exposed via GraphQL, but the controller stores it nested as `schedule: {mode: ...}`. Updating it hits the same 400. Same bug family, out of scope here — I'm opening an issue for it.

Merging once CI is green.

Issue #440, opened 2026-07-09
Firewall policy/zone tools return empty on legacy (pre-ZBF) sites, masking real rules and causing false "critical" audit findings


mundijr:
## Summary

`unifi_list_firewall_policies` and `unifi_list_firewall_zones` only read the V2 zone-based firewall API (`/firewall-policies`, `/firewall/zone-matrix`). On a controller still running the classic (pre-Zone-Based-Firewall) rule engine, both endpoints return empty — indistinguishable from "no firewall rules configured." The legacy rule engine (`/rest/firewallrule`) has no read path anywhere in this server.

This is different from the `unifi_create`/`unifi_update_firewall_policy` legacy-field removal in #210 — that was about a broken *write* path. This issue is about a *read* gap that was apparently never covered at all (`git log` shows no prior implementation for `/rest/firewallrule`), while `get_firewall_groups()` already reads the sibling legacy endpoint `/rest/firewallgroup` correctly.

## Real-world impact

Ran a live read-only audit against a home controller: `unifi_list_firewall_policies` and `unifi_list_firewall_zones` both returned 0 results, which read as "no VLAN segmentation enforced" — a critical finding. After the user migrated the site to Zone-Based Firewalls (through the UniFi UI, unrelated to this MCP server), the same tools suddenly surfaced 77 policies, including several custom rules (e.g. blocking one VLAN from reaching another) that had existed the entire time under the legacy engine. The original "critical" finding was a false positive caused entirely by the read gap, not a real misconfiguration.

## Why this matters beyond one audit

This repo's `firewall-auditor` skill (`plugins/unifi-network/skills/firewall-auditor/`) runs essentially the same check — SEG-01/02/03 in `security-benchmarks.md` score VLAN segmentation using only `unifi_list_firewall_policies`/`unifi_list_firewall_zones`. Any site still on the legacy engine would get scored as having zero segmentation regardless of its actual ruleset, since the skill has no way to see legacy rules either.

## Suggested fix

- Add a read-only `FirewallManager.get_legacy_firewall_rules()` (`GET /rest/firewallrule`), mirroring the existing `get_firewall_groups()` pattern against `/rest/firewallgroup`.
- Expose it as a new read-only tool (`unifi_list_legacy_firewall_rules`, maybe `unifi_get_legacy_firewall_rule_details`), consistent with the `FirewallGroup`/`FirewallZone` model pattern in `unifi_core/network/models/firewall.py`.
[...]

PR #471, opened 2026-07-29
Add legacy firewall rule read support

rohitsud04:
## Summary

Adds read-only support for legacy pre-zone-based UniFi firewall rules.

## Changes

- Add `FirewallManager.get_legacy_firewall_rules()`
- Fetch legacy rules through `/rest/firewallrule`
- Cache results per site
- Accept list and `{"data": [...]}` controller responses
- Add the `unifi_list_legacy_firewall_rules` MCP tool
- Redact sensitive fields using the existing redaction path
- Add manager and tool unit tests

## Validation

- 86 unit tests passed
- Ruff checks passed
- `git diff --check` passed

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,7 +251,7 @@ This is a monorepo with shared packages:
 
 ```
 apps/
-  network/          # UniFi Network MCP server (stable, 186 tools)
+  network/          # UniFi Network MCP server (stable, 187 tools)
   protect/          # UniFi Protect MCP server (beta, 61 tools)
   access/           # UniFi Access MCP server (beta, 36 tools)
   api/              # Independent REST + GraphQL API server (beta)
--- a/apps/api/docs/graphql-reference.md
+++ b/apps/api/docs/graphql-reference.md
@@ -919,6 +919,45 @@ type KnownLicensePlatePage {
   nextCursor: String
 }
 
+"""
+A rule from the legacy (pre-zone-based) firewall engine, read from V1 /rest/firewallrule. Distinct from FirewallRule, which models the V2 zone-based engine: actions are lowercase (accept/drop/reject), rules belong to a ruleset rather than a zone pair, and source/destination are flat fields rather than nested objects. Read-only.
+"""
+type LegacyFirewallRule {
+  id: ID
+  name: String
+  ruleset: String
+  ruleIndex: Int
+  action: String
+  enabled: Boolean
+  protocol: String
+  protocolV6: String
+  protocolMatchExcepted: Boolean
+  srcAddress: String
+  srcAddressIpv6: String
+  srcPort: String
+  srcMacAddress: String
+  srcFirewallgroupIds: [String!]!
+  srcNetworkconfId: String
+  srcNetworkconfType: String
+  dstAddress: String
+  dstAddressIpv6: String
+  dstPort: String
+  dstFirewallgroupIds: [String!]!
+  dstNetworkconfId: String
+  dstNetworkconfType: String
+  stateNew: Boolean
+  stateEstablished: Boolean
+  stateRelated: Boolean
+  stateInvalid: Boolean
+  icmpTypename: String
+  icmpv6Typename: String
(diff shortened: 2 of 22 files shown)


sirkirby:
Nice first PR — and a well-chosen one. The manager method mirrors `get_firewall_groups()` correctly, the caching and response normalisation are right, logging uses lazy `%s` formatting, and the redaction path is wired properly. 86 tests passing on a first contribution is a good showing.

I pushed a follow-up commit to your branch rather than sending it back, because the gaps needed hardware and repo context you'd have had no way to get at. Here's all of it, so nothing is a mystery.

**Two CI failures.** `ruff format` flagged `test_firewall_tools.py` — your PR body said Ruff passed, which was true of `ruff check` but `make format` is a separate gate. The more consequential one: `make generate` hadn't been run, so `test_tool_map_sync` failed with `Tools in TOOL_MODULE_MAP but not in manifest`. That matters beyond the red X — `runtime.py`, `categories.py` and `tool_index.py` all read `tools_manifest.json` for discovery, so as shipped the tool existed in code but was **invisible to MCP clients**. Worth knowing for next time: `make pre-commit` runs format, generate, lint, tests and the drift checks in one go.

**The model.** The tool returned raw controller dicts, but every sibling read tool projects through a domain model — `firewall_zone_from_controller(...).model_dump()`, same for groups — and the API layer needs a typed shape to expose. So I added a `LegacyFirewallRule` model.

I deliberately did *not* reuse the existing `FirewallRule`. It models the V2 zone-based engine: uppercase actions, nested source/destination objects, no ruleset. Mapping legacy rules through it would have silently dropped `ruleset`, every `src_`/`dst_` field, the firewall-group and network references, and the connection-state flags — which are exactly the fields an audit needs.

The field set comes from the controller-generated schema behind the Terraform and Pulumi providers, since I had no legacy hardware to read a live payload from. That source also corrected something: the ruleset enum has **18** values including the IPv6 variants (`LANv6_IN`, `GUESTv6_OUT`, …), and `VPN_IN`/`VPN_OUT`/`VPN_LOCAL` — which an older version of this repo listed — were never real.

**The part that actually fixes the reported harm.** #440 wasn't only about legacy rules being unreadable; it was about an empty V2 response being indistinguishable from "no firewall rules configured", which produced a false *critical* segmentation finding. Adding a tool alone doesn't fix that, because nothing tells a caller to go look. `unifi_list_firewall_policies` and `unifi_list_firewall_zones` now include a note pointing at the legacy tool when the controller returns nothing.

One subtlety there, caught by a test I wrote and initially got wrong: `total_count` in the policies tool is computed *after* filtering, so keying the hint off it would fire on any search that matched nothing. It now uses the pre-filter count — a filter miss says nothing about which engine you're on.

**API surface.** Followed `firewall_zones` end to end: Strawberry type, type-registry entry, GraphQL resolver, and a paginated REST route ordered by ruleset then `rule_index`. That tripped five more generated-artifact gates (SDL, OpenAPI, both reference docs, and the resolver-coverage fixture gate), all regenerated.

**Verification.** Ruff format and lint clean, `make check-generated` green, 1429 unifi-core / 886 network / 945 API tests. Against a live UDM SE: `/rest/firewallrule` returns 200 with an empty list rather than a 404, so the tool degrades cleanly on a migrated controller, and the hint correctly stays silent there (8 zones, 51 policies) and on a filter miss.

I could not exercise the populated path — this controller is zone-based-migrated, so it has no legacy rules. If you or anyone reading has a site still on the legacy engine, a sanitised `/rest/firewallrule` payload would let us confirm the model against real data.

Leaving #440 open on purpose. The firewall-auditor skill and the SEG-01–03 benchmarks still can't see legacy rules, so the false-finding path is only half closed. That changes audit scoring, so it gets its own PR rather than riding along here.

Merging once CI is green. Thanks for picking up a real gap — this one had teeth.

sirkirby (approved):
Verified end to end — see the detailed comment above. Merging.

PR #474, opened 2026-08-01
fix(network): reach the v2 security log when the legacy IPS endpoint answers empty

### LINKED ISSUE #473: unifi_get_ips_events returns count=0 on controllers that moved IPS events to the v2 system log: the v2 fallback only runs on exception
## Summary

`unifi_get_ips_events` reports `success: true, count: 0` on a controller that does have threat data. The v2 fallback that exists precisely for these controllers is never reached, because it is guarded by `except Exception` and the legacy endpoint does not raise — it answers `200` with an empty list.

## Environment

- `unifi-network-mcp` 0.24.1, HTTP mode, container `ghcr.io/sirkirby/unifi-network-mcp:0.24.1`
- Controller: UDM SE, Network application **10.4.57**
- Site: `default`

## Actual behaviour

```json
{"success": true, "site": "default", "duration": "weekly", "limit": 50, "count": 0, "events": []}
```

Same result for `hourly`, `daily`, `weekly`. No error, no warning — the tool reports an all-clear.

For contrast, on the same controller in the same session `unifi_list_alarms` answers `success: true, count: 0` as well rather than failing, and `unifi_get_system_info` / `unifi_get_anomalies` both return real data (the latter with 8 entries), so connectivity and auth are fine.

## Cause

`packages/unifi-core/src/unifi_core/network/managers/stats_manager.py`, `get_ips_events`:

```python
try:
    response = await self._connection.request(api_request)      # POST /stat/ips/event
except Exception as legacy_error:
    logger.debug("Legacy IPS endpoint failed, falling back to v2 security logs: %s", legacy_error)
    response = await self._get_security_events_v2(start_time, end_time, limit)
result = response if isinstance(response, list) else []
```

`_get_security_events_v2` is reached only when the legacy request **raises**. A controller that has moved IPS events to the v2 system log does not raise — `/stat/ips/event` still exists and answers `200 []`. So the fallback never fires on exactly the versions it was written for, and the empty list is returned as a successful answer.

The tool's own description already states the premise:

> Note: on newer UniFi OS / Network Application versions, IPS events may not populate this legacy endpoint.

That is the case the fallback cannot reach.

## Suggested fix

Treat an empty (or non-list) legacy response as inconclusive and try v2 as well. PR to follow.

Two details worth stating, because the obvious implementations get them wrong:

- A v2 failure after an empty legacy answer should **not** raise. A controller old enough to have no v2 endpoint answers exactly this way — legacy `200 []`, v2 unavailable — so propagating would break setups the legacy endpoint still serves correctly today. This makes the branch deliberately asymmetric with the existing `except` branch, which does propagate.
- The degraded empty result should not be written to the 5-minute cache, or a transient v2 outage is held as a confirmed quiet period for the whole TTL. On a security-event endpoint that is the difference between "no threats" and "we could not tell".

## Related, for context — none of these is the same bug

- #461 — `alarms/events` 404 on UDM Pro Max **10.4.57**, `EventManager._detect_api_version()` failing silently. Same controller version as here and adjacent in spirit (a v2 migration handled by a fallback that does not fire), but a different symptom and a different manager: we do not reproduce the 404, our `unifi_list_alarms` answers `200`.
- #14 — `SSE mode fails when UNIFI_MCP_HTTP_ENABLED=true`, closed. Recorded because that variable no longer appears to do anything in the current tree; if that is intentional, its remaining references may be worth dropping. Happy to open a separate issue if useful rather than clutter this one.
- #8 — `Unable to pull ghcr.io/sirkirby/unifi-network-mcp:latest`, closed. Same area as a README/GHCR discoverability nit we hit; again separate, mentioning it only so the history is linked.

nicodarge:
Fixes #473.

## The problem

`get_ips_events` guards its v2 fallback with `except Exception`, so it runs only when `/stat/ips/event` raises. Controllers that moved IPS events to the v2 system log do not raise — they answer `200` with an empty list. The fallback therefore never fires on the versions it exists for, and the tool returns `success: true, count: 0` on a controller that has threat data.

The tool description already documents the premise ("on newer UniFi OS / Network Application versions, IPS events may not populate this legacy endpoint"), so the code states the case it cannot reach.

Observed on a UDM SE running Network 10.4.57, `unifi-network-mcp` 0.24.1: `count: 0` for hourly, daily and weekly, while `unifi_get_anomalies` returns 8 entries and `unifi_get_system_info` is fine on the same session.

## The change

An empty — or non-list — legacy response is treated as inconclusive and v2 is tried as well.

Two decisions that are less obvious than they look:

**A v2 failure after an empty legacy answer does not raise.** This is deliberately asymmetric with the existing `except` branch, which propagates. A controller old enough to have no v2 endpoint answers exactly this shape — legacy `200 []`, then v2 unavailable — so propagating would turn a setup the legacy endpoint still serves correctly into a hard failure. The asymmetry is stated in a comment rather than left to be re-derived.

**The degraded result is returned early, so it is not cached.** Writing it to the 5-minute cache would hold a transient v2 outage as a confirmed quiet period, and on a security-event endpoint that is the difference between "no threats" and "we could not tell". A retry within the TTL now re-probes instead of replaying the empty answer.

The trigger is `not isinstance(response, list) or not response` rather than `not response`, to match the normalisation two lines below: a malformed non-list answer was previously coerced to `[]` without ever reaching v2 — the same silent-empty bug in a different shape.

## Tests

`apps/network/tests/unit/test_stats_enhanced.py`, alongside the existing ones for this file:

- empty legacy → v2 is reached, its events are returned
- non-empty legacy → v2 is **not** called (no double query for controllers still populating the legacy endpoint)
- non-list legacy → v2 is reached rather than the answer being coerced to `[]`
- v2 failure after empty legacy → returns `[]` rather than raising
- v2 failure after empty legacy → nothing is written to the cache

`test_get_ips_events_custom_limit` asserted on the **last** request, which is now the v2 one. It asserts on the legacy call explicitly instead, and gains the path assertion it was missing.

## Verification

```
uv run pytest apps/network/tests/unit/test_stats_enhanced.py   →  47 passed
make lint                                                      →  All checks passed
make check-generated                                           →  no drift
```

One caveat, stated rather than hidden: `make core-test` fails in my environment on 7 collection errors, `ModuleNotFoundError: No module named 'uiprotect'`, in `packages/unifi-core/tests/protect/`. I confirmed it fails identically on `main` without this change, so it is not a regression here — but it does mean the documented target for a `packages/unifi-core/` change could not exercise this code. The tests that do cover this file live under `apps/network/tests/unit/`, following the existing layout for `stats_manager`. I have a one-line fix for that target and will send it separately rather than mix it into this PR.

### DIFF
--- a/apps/network/tests/unit/test_stats_enhanced.py
+++ b/apps/network/tests/unit/test_stats_enhanced.py
@@ -265,9 +265,50 @@ class TestStatsManagerEnhanced:
 
         await stats_manager.get_ips_events(limit=10)
 
-        call_args = mock_connection.request.call_args
-        api_request = call_args[0][0]
+        # An empty legacy answer now triggers the v2 fallback, so the legacy
+        # request is the first call rather than the last one.
+        api_request = mock_connection.request.call_args_list[0][0][0]
+        assert api_request.path == "/stat/ips/event"
         assert api_request.data["_limit"] == 10
+
+    @pytest.mark.asyncio
+    async def test_get_ips_events_empty_legacy_falls_back_to_v2(self, stats_manager, mock_connection):
+        """An empty legacy response must still reach the v2 security log.
+
+        Controllers that moved IPS events to the v2 system log answer
+        /stat/ips/event with 200 and an empty list rather than failing, so a
+        fallback guarded only by an exception never runs for them.
+        """
+        v2_events = [{"key": "EVT_IPS_IpsAlert"}]
+        mock_connection.request.side_effect = [[], v2_events]
+
+        result = await stats_manager.get_ips_events()
+
+        assert result == v2_events
+        assert mock_connection.request.call_count == 2
+        assert mock_connection.request.call_args_list[0][0][0].path == "/stat/ips/event"
+        assert mock_connection.request.call_args_list[1][0][0].path == "/system-log/critical"
+
+    @pytest.mark.asyncio
+    async def test_get_ips_events_non_empty_legacy_skips_v2(self, stats_manager, mock_connection):
+        """A controller still populating the legacy endpoint must not be double-queried."""
+        legacy_events = [{"event_type": "alert"}]
+        mock_connection.request.return_value = legacy_events
+
+        result = await stats_manager.get_ips_events()
+
+        assert result == legacy_events
+        assert mock_connection.request.call_count == 1
+
+    @pytest.mark.asyncio
+    async def test_get_ips_events_v2_failure_after_empty_legacy_returns_empty(self, stats_manager, mock_connection):
+        """A failing v2 fallback must not turn an empty result into an exception."""
+        mock_connection.request.side_effect = [[], Exception("v2 unavailable")]
+
+        result = await stats_manager.get_ips_events()
+
+        assert result == []
+        assert mock_connection.request.call_count == 2
 
     @pytest.mark.asyncio
     async def test_get_client_sessions(self, stats_manager, mock_connection):
--- a/apps/network/tests/unit/test_stats_enhanced.py
+++ b/apps/network/tests/unit/test_stats_enhanced.py
@@ -302,12 +302,36 @@ class TestStatsManagerEnhanced:
 
     @pytest.mark.asyncio
     async def test_get_ips_events_v2_failure_after_empty_legacy_returns_empty(self, stats_manager, mock_connection):
-        """A failing v2 fallback must not turn an empty result into an exception."""
+        """A failing v2 fallback must not turn an empty result into an exception.
+
+        A controller old enough to have no v2 endpoint answers exactly this way,
+        so raising here would break setups the legacy endpoint still serves.
+        """
         mock_connection.request.side_effect = [[], Exception("v2 unavailable")]
 
         result = await stats_manager.get_ips_events()
 
         assert result == []
+        assert mock_connection.request.call_count == 2
+
+    @pytest.mark.asyncio
+    async def test_get_ips_events_v2_failure_is_not_cached(self, stats_manager, mock_connection):
+        """A degraded empty result must not be cached as a confirmed quiet period."""
+        mock_connection.request.side_effect = [[], Exception("v2 unavailable")]
+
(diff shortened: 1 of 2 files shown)


sirkirby (approved):
This is a well-argued PR. The diagnosis is right, the asymmetry is the correct call, and stating both non-obvious decisions in comments rather than leaving them to be re-derived is the right instinct.

I verified independently rather than relying on the run in your description: Ruff clean, `make check-generated` clean, your 47 target tests plus the full 883-test network suite green. The branch coverage is precise — `_update_cache.assert_not_called()` pins the uncached-degraded-result claim rather than just asserting the return value, which is the assertion that would actually catch a regression there.

**One thing worth recording, because it says something about how narrow this window is.** I could not reproduce the 200-empty condition on my hardware. My UDM Pro SE runs Network **10.5.67**, and there `/stat/ips/event` is gone entirely — it 404s, so the existing `except` branch fires and the fallback already worked. Yours is 10.4.57, where the endpoint still exists and answers empty.

So the progression looks like: legacy populated → legacy present but empty (your band) → legacy removed (mine). The bug is real but lives in a transitional window, which is exactly the kind of thing that goes unnoticed — the versions on either side of it behave correctly for different reasons.

I confirmed the 404 path is unaffected by your change: `get_ips_events` still returns cleanly on 24h and 168h lookbacks here, going legacy-404 → v2 → empty. What I could not exercise is the populated-v2 path, since `/system-log/critical` returns zero SECURITY events on this controller for both windows. Noting that as a real gap rather than implying full coverage.

**Your caveat about `make core-test` is accurate and I reproduced it.** 7 collection errors, `ModuleNotFoundError: No module named 'uiprotect'`, and it is not a regression from this change — the failing files are all under `packages/unifi-core/tests/protect/`, which you do not touch. Worth noting it is nastier than it looks: the whole suite aborts at collection, so `core-test` reports failure having run **zero** unifi-core tests. It is also order-dependent — in a worktree where I had previously run another package, `uiprotect` was already installed and the same command passed, which is a good way for this to stay hidden. Reviewing #475 next.

Merging.

PR #516, opened 2026-08-09
fix(network): populate event mac/ip/msg from v2 system-log records

ljmerza:
## Summary

`unifi_list_events` and `unifi_recent_events` return events with no `mac`, no `ip`, and no `msg` on modern controllers:

```json
{ "id": "…", "key": "TRAFFIC_BLOCKED_KNOWN_SOURCE_CLIENT", "time": 1786224188475, "severity": "MEDIUM" }
```

Every event is anonymous, so there is no way to tell *which* client connected, disconnected, was blocked by a firewall policy, or tripped IPS. For security events in particular that removes the entire point of the record.

**`EventLog` already declares `mac`, `ip` and `msg`** — they resolve to `None` and `model_dump(exclude_none=True)` strips them. This is a serializer gap, not a model or endpoint limitation.

`get_events` gained a v2 branch (`_get_events_v2` → `/system-log/all`) but `event_log_from_controller` was never taught the v2 record shape. It still looks only for the legacy `/stat/event` flat keys:

| Field | Factory looks for | v2 actually provides |
|---|---|---|
| `mac` | `user`, `mac`, `ap`, `ap_mac`, `device_mac` | `parameters.SRC_CLIENT.id` |
| `ip` | `ip`, `src_ip` | `parameters.SRC_CLIENT.ip` |
| `msg` | `msg`, `message`, `description` | `message_raw` + `parameters` |
| `time` | `time`, `timestamp`, `ts` | `timestamp` ✅ — the only one that matched |

## Type of change

- [x] `fix:` bug fix
- [ ] `feat:` new feature

## Scope

- [x] Network MCP server
- [ ] Protect MCP server
- [ ] Access MCP server
- [ ] Relay sidecar
- [ ] Cloudflare Worker / CLI
- [ ] API server
- [x] Shared packages
- [ ] Plugin packaging
- [ ] Documentation

## Change

v2 nests each actor under `parameters`, keyed by its **role**, and ships a template instead of a rendered message:

```json
{
  "key": "TRAFFIC_BLOCKED_KNOWN_SOURCE_CLIENT",
  "message_raw": "{SRC_CLIENT} was blocked from accessing {DST_IP} by the {TRIGGER} Firewall Policy.",
  "parameters": {
    "SRC_CLIENT": { "id": "aa:bb:cc:00:00:01", "ip": "192.0.2.11", "name": "Lab-Camera" },
    "DST_IP":     { "id": "198.51.100.24", "name": "198.51.100.24" },
    "TRIGGER":    { "id": "trigger-0001", "name": "block cameras to external" }
  }
}
```

Three behaviours in `models/events.py`:

1. **Role-scoped, shape-validated MAC.** `id` holds a MAC for client/device roles but an **IP** for address roles like `DST_IP`, so a naive scan would put `198.51.100.24` in `mac`. Extraction walks a fixed role priority (`SRC_CLIENT`, `CLIENT`, `DST_CLIENT`, `SRC_DEVICE`, `DEVICE`, `AP`, `GATEWAY`) and requires a MAC-shaped value.

2. **`mac` and `ip` come from one actor.** Resolving them independently pairs fields from different devices: on `THREAT_DETECTED_KNOWN_SOURCE_CLIENT`, `SRC_CLIENT` is the offender but carries no `ip`, while `DEVICE` (the reporting console) does — yielding the client's MAC beside the *gateway's* IP, which reads as one device and is wrong. Both are now read from a single primary actor; `ip` falls back across actors only when no MAC-bearing actor exists, where there is no pairing to get wrong.

3. **`message_raw` rendered against `parameters`**, preferring `name` → `hostname` → `id`. Unknown placeholders are left intact so an unrecognised role degrades to the raw template rather than a sentence with holes in it. Falls back to `title_raw`.

Legacy keys are still checked first, so `/stat/event` controllers are unaffected — covered by a test asserting legacy wins when both shapes are present.

No tool signature, field, or description change, so no manifest / GraphQL / OpenAPI regeneration.

### Result

```json
{
  "id": "evt-0001",
  "key": "TRAFFIC_BLOCKED_KNOWN_SOURCE_CLIENT",
  "time": 1786225096952,
  "severity": "MEDIUM",
  "mac": "aa:bb:cc:00:00:01",
  "ip": "192.0.2.11",
  "msg": "Lab-Camera was blocked from accessing 198.51.100.24 by the block cameras to external Firewall Policy."
}
```

## Checklist

- [x] I ran `make pre-commit`.
- [x] I updated generated manifests with `make manifest` when changing MCP tools. *(n/a — no tool surface change; drift check clean.)*
- [x] I updated documentation for user-visible changes. *(n/a — no doc-visible surface change.)*

## Testing

Field names and nesting were read from a **live UDM Pro running Network 5.1.26**,
not inferred — a `POST /v2/api/site/<site>/system-log/all` capture using the same
payload `_get_events_v2` builds.

Replaying 100 consecutive live records through the factory, before vs after:

| | before | after |
|---|---|---|
| `mac` populated | 0 / 100 | **100 / 100** |
| `msg` populated | 0 / 100 | **100 / 100** |
| malformed MACs | — | **0** |

Spanning 7 distinct event keys (`CLIENT_CONNECTED_WIRELESS_2`,
`CLIENT_DISCONNECTED_WIRELESS_2`, `CLIENT_CONNECTED_WIRED_2`,
`CLIENT_DISCONNECTED_WIRED_2`, `THREAT_DETECTED_KNOWN_SOURCE_CLIENT`,
`TRAFFIC_BLOCKED_KNOWN_SOURCE_CLIENT`, `CLIENT_ROAMED_2`).

### New tests

| Layer | File | Cases |
|---|---|---|
| `unifi-core` model | `packages/unifi-core/tests/network/models/test_events.py` *(new)* | 15 |
| Network tool | `apps/network/tests/unit/test_event_tools.py` | +2 |

Core cases run off a sanitised `system_log_events_response.json` fixture and cover
both v2 and legacy shapes, the address-role-is-not-a-MAC guard, the same-actor
`mac`/`ip` pairing guarantee, unknown-placeholder passthrough, `title_raw`
fallback, and malformed `parameters`. The tool-layer cases go end-to-end through
`unifi_list_events` so the assertion is on real tool output, not just the model —
`model_dump(exclude_none=True)` drops unresolved fields, so a mapping gap surfaces
as missing keys rather than nulls.

**Verified the new tests fail without the fix.** Checking `upstream/main`'s
`events.py` into the working tree and re-running:

```
packages/unifi-core/.../test_events.py     8 failed, 7 passed
apps/network/.../test_event_tools.py       1 failed  (KeyError: 'mac')
```

The cases that still pass under the old code are the legacy-shape and defensive
ones, which is the intended result — they assert behaviour this PR preserves.

### Suites

```
core-test (packages/unifi-core)      1504 passed
apps/api                             1185 passed
apps/network                          935 passed
shared-test                           359 passed
relay-test                            111 passed
ruff check / format                  All checks passed!
make pre-commit                      rc=0
check-generated                      "Checked 39 sections; no drift."
```

## Notes for reviewers

**The API layer carries a second copy of this bug — deliberately left alone.**
`unifi_api.graphql.types.network.event.EventLog.from_manager_output` duplicates
the same legacy-only field mapping, so it is blind to v2 in exactly the same way.
Its `render_hint` lists `mac` in `display_columns`, so on a modern controller that
rendered table has a permanently empty column.

I kept it out of this PR to keep the change reviewable and scoped to one
behaviour. The obvious fix is to delegate that method to
`event_log_from_controller` rather than patch the duplicate — `types/network/firewall.py`
already sets that precedent by importing its factory from `unifi-core` — but that
touches all seven fields on a type this PR otherwise doesn't need to change, and
removing the duplication is a maintainer call. Happy to follow up in a separate PR
if that's the direction you want.

**Test path quirk, pre-existing.** `apps/network/tests/unit/test_tool_map_sync.py`
errors with `FileNotFoundError` when pytest runs from `apps/network/` rather than
the repo root — it resolves `apps/network/src/.../tools_manifest.json` relative to
cwd. Unrelated to this PR; passes from the root, which is how `make` invokes it.

**No manifest / GraphQL / OpenAPI regeneration:** no tool signature, field, or
description changed. `check-generated` confirms no drift.

### DIFF
--- a/apps/network/tests/unit/test_event_tools.py
+++ b/apps/network/tests/unit/test_event_tools.py
@@ -33,3 +33,85 @@ async def test_archive_all_alarms_preview_requires_confirmation():
     assert result["resource_type"] == "alarm_collection"
     assert result["resource_id"] == "all_active_alarms"
     assert result["preview"]["proposed"]["archived"] is True
+
+
+class _StubConnection:
+    site = "default"
+
+
+class _StubEventManager:
+    """Minimal stand-in for EventManager returning canned records."""
+
+    _connection = _StubConnection()
+
+    def __init__(self, records):
+        self._records = records
+
+    async def get_events(self, **_kwargs):
+        return self._records
+
+
+_V2_RECORD = {
+    "id": "evt-0001",
+    "key": "TRAFFIC_BLOCKED_KNOWN_SOURCE_CLIENT",
+    "severity": "MEDIUM",
+    "timestamp": 1786225096952,
+    "message_raw": "{SRC_CLIENT} was blocked from accessing {DST_IP} by the {TRIGGER} Firewall Policy.",
+    "parameters": {
+        "SRC_CLIENT": {"id": "aa:bb:cc:00:00:01", "ip": "192.0.2.11", "name": "Lab-Camera"},
+        "DST_IP": {"id": "198.51.100.24", "name": "198.51.100.24"},
+        "TRIGGER": {"id": "trigger-0001", "name": "block cameras to external"},
+    },
+}
+
+
+@pytest.mark.asyncio
+async def test_list_events_surfaces_mac_ip_and_msg_for_v2_records(monkeypatch):
+    """End-to-end through the tool: v2 records must not come back anonymous.
+
+    ``model_dump(exclude_none=True)`` drops unresolved fields, so a mapping
+    gap shows up as missing keys rather than nulls.
+    """
+    from unifi_network_mcp.tools import events as events_module
+
+    monkeypatch.setattr(
+        events_module,
+        "_get_event_manager",
+        lambda: _StubEventManager([_V2_RECORD]),
+    )
+
+    result = await events_module.list_events(within_hours=1, limit=10)
+
+    assert result["success"] is True
+    assert result["count"] == 1
+    event = result["events"][0]
+    assert event["mac"] == "aa:bb:cc:00:00:01"
+    assert event["ip"] == "192.0.2.11"
+    assert event["msg"] == (
+        "Lab-Camera was blocked from accessing 198.51.100.24 by the block cameras to external Firewall Policy."
+    )
+
+
+@pytest.mark.asyncio
+async def test_list_events_still_maps_legacy_records(monkeypatch):
+    from unifi_network_mcp.tools import events as events_module
+
+    legacy = {
+        "_id": "legacy-1",
+        "key": "EVT_WU_Disconnected",
+        "msg": "Client disconnected",
+        "time": 1700000000000,
+        "user": "aa:bb:cc:00:00:09",
+        "ip": "192.0.2.50",
+    }
+    monkeypatch.setattr(
+        events_module,
+        "_get_event_manager",
+        lambda: _StubEventManager([legacy]),
+    )
+
+    event = (await events_module.list_events())["events"][0]
+
+    assert event["mac"] == "aa:bb:cc:00:00:09"
+    assert event["ip"] == "192.0.2.50"
+    assert event["msg"] == "Client disconnected"
--- a/packages/unifi-core/src/unifi_core/network/models/events.py
+++ b/packages/unifi-core/src/unifi_core/network/models/events.py
@@ -15,7 +15,8 @@ MUTABLE_FIELDS = frozenset() (all fields are read-only).
 
 from __future__ import annotations
 
-from typing import Any, Optional
+import re
+from typing import Any, Iterator, Optional
 
 from pydantic import BaseModel, Field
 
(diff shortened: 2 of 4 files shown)


sirkirby (approved):
I ran the full gate checklist and an independent live validation on the reference UDM SE (UniFi OS 5.1.x / Network 10.3.x, v2 system-log path) — everything checks out.

**Independent live results, branch code in-process (not the published package):**

- Baseline on `main`: 0/100 events with `mac`, `ip`, or `msg` populated
- With this branch: **100/100 on all three**, 0 malformed MACs, 0 unresolved placeholders, across 6 distinct event keys (wired/wireless connect/disconnect, roam, firewall block)
- `scripts/live_smoke.py --server network --phase readonly` clean for `unifi_list_events` and `unifi_recent_events`

**Gates:** Ruff clean, no f-string loggers, fixture properly sanitized (documentation MACs / RFC 5737 addresses — appreciated), all defaults on the model remain `None`, no tool-surface change so no manifest/doc drift (`check-generated` confirms), and the new tests fail-without-fix claim reproduces. The same-actor mac/ip pairing guarantee and the address-role MAC guard are exactly the right calls — the `THREAT_DETECTED` case would have silently paired an offender MAC with the gateway IP otherwise.

I verified the API-layer duplicate you flagged: `from_manager_output` in the GraphQL `EventLog` type indeed carries the same legacy-only mapping and doesn't delegate to the core factory. Opened #517 for that, going with your suggested delegation approach — a focused follow-up PR is welcome if you want to take it.

PR #527, opened 2026-08-11
fix(network): make port profiles able to express an access port

### LINKED ISSUE #526: Port profile tools cannot create an access port, and report success on writes the controller rewrites
Cross-posting from my fork, where the full write-up (including a live-tested root-cause refinement in the comments) lives: rmaher001/unifi-mcp#5

## Summary

`unifi_create_port_profile` / `unifi_update_port_profile` cannot produce a working **access port**, and both report success on writes the controller silently rewrites.

## Root cause (confirmed live against a real controller, see fork issue comments)

`PortProfile` (`packages/unifi-core/src/unifi_core/network/models/switch.py`) has no `tagged_vlan_mgmt` field — the UI's "Tagged VLAN Management" control, which is the actual access-vs-trunk switch. Without it, the MCP can only ever send a self-contradictory profile (`forward: "native"` alongside an inherited `allow_all`), and the controller resolves the contradiction by rewriting `forward` to agree with the tagged setting rather than the reverse. Adding `tagged_vlan_mgmt` to the model/mutable-set is expected to fix the `forward` coercion as a side effect — no special-casing of `forward` needed.

## Other gaps found in the same area

1. `tagged_networkconf_ids` exists on the model and is mutable via `update`, but is unreachable from `create` (no parameter) and undocumented on `update` (absent from its `Allowed keys:` list).
2. `create_port_profile` omits `poe_mode` (and `stp_port_mode`, `isolation`) from the payload whenever the caller takes the documented default, so the controller applies its own default instead — for `poe_mode: "auto"` this means the created profile ships with Auto PoE **unchecked**, a silent power-loss risk for anything wired through it.
3. `port_mode` (Infrastructure/Edge) isn't exposed at all.
4. Neither tool reads back the controller's post-write state, so a coerced or ignored write is reported as success.

## Requested fields

`tagged_vlan_mgmt` (str: `allow_all`/`block_all`/`custom`), `port_mode` (str: `infrastructure`/`edge`), `excluded_networkconf_ids` (list[str]), `tagged_networkconf_ids` exposed on `create`.

Plus: always send `poe_mode`/`stp_port_mode`/`isolation` rather than omitting at default, and read back stored state on create/update to surface coercion.

Full repro, evidence, and the live-tested diagnosis refinement are in the fork issue linked above. Happy to send a PR if the direction is agreeable.

rmaher001:
Fixes #526. Reproduced and verified on UniFi Network 10.5.67.

`tagged_vlan_mgmt` was missing from `PortProfile`, and it determines whether the controller stores an access or trunk profile. The controller rewrites `forward` to agree with it, so the previously exposed inputs could report success while storing a trunk.

Changes:

- Add `tagged_vlan_mgmt`, `excluded_networkconf_ids`, and the three fields behind the UI's Port Mode control (`stp_edge_state`, `stp_bpdu_guard_enabled`, `stp_uplink`).
- Expose `tagged_networkconf_ids` on create; it was mutable on update but unreachable from create.
- Always send `poe_mode`, `stp_port_mode`, and `isolation`, including when they equal the tool defaults.
- Build create payloads through the shared model helper so MCP and `/v1/actions` cannot drift.
- Use the shared structured write-verification contract for create and update. Both operations perform an independent read-back, fail on missing or coerced requested fields, and distinguish a rejected write from a mutation that landed but could not be verified.
- Clarify that Port State Active/Disabled is a separate controller setting and remains outside this PR's port-profile surface.

## Maintainer validation (2026-08-13)

- `make pre-commit`: passed (core 1576, relay 111, Network 993, Protect 503, Access 206, API 1217, worker 99; generation, lint, registration modes, and typecheck clean).
- Exact final commit: 25 core switch-model tests and 51 Network switch manager/tool tests passed; generated artifacts and lint clean.
- Network read-only hardware smoke: 121 passed, 7 argument-discovery skips, 0 failures or exceptions.
- Network safe smoke: passed.
- Targeted disposable profile lifecycle: create, read, update, deliberate coercion detection, restore, delete preview, and delete all behaved as expected; cleanup confirmed no orphaned profile.

Live access-port read-back:

```
forward: "native"
tagged_vlan_mgmt: "block_all"
poe_mode: "auto"
stp_edge_state: "enabled"
stp_bpdu_guard_enabled: true
stp_uplink: false
```

Deliberately requesting an inconsistent combination now returns structured failure while accurately reporting that the controller mutation landed and identifying the coerced field.

### DIFF
--- a/apps/api/docs/graphql-reference.md
+++ b/apps/api/docs/graphql-reference.md
@@ -1491,12 +1491,17 @@ type PortProfile {
   id: ID
   name: String
   forward: String
+  taggedVlanMgmt: String
   nativeNetworkconfId: String
   taggedNetworkconfIds: [String!]!
+  excludedNetworkconfIds: [String!]
   voiceNetworkconfId: String
   poeMode: String
   isolation: Boolean
   stpPortMode: Boolean
+  stpEdgeState: String
+  stpBpduGuardEnabled: Boolean
+  stpUplink: Boolean
   dot1xCtrl: String
 }
 
--- a/apps/api/openapi.json
+++ b/apps/api/openapi.json
@@ -6512,6 +6512,20 @@
             ],
             "title": "Dot1X Ctrl"
           },
+          "excluded_networkconf_ids": {
+            "anyOf": [
+              {
+                "items": {
+                  "type": "string"
+                },
+                "type": "array"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Excluded Networkconf Ids"
+          },
           "forward": {
             "anyOf": [
               {
@@ -6578,6 +6592,28 @@
             ],
             "title": "Poe Mode"
           },
+          "stp_bpdu_guard_enabled": {
+            "anyOf": [
+              {
+                "type": "boolean"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Stp Bpdu Guard Enabled"
+          },
+          "stp_edge_state": {
+            "anyOf": [
+              {
+                "type": "string"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Stp Edge State"
+          },
           "stp_port_mode": {
             "anyOf": [
               {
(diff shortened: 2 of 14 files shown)


sirkirby (approved):
Maintainer validation complete. The structured write-verification repair, full local quality gate, real-controller smoke coverage, and all required GitHub checks are green.

PR #528, opened 2026-08-11
fix(network): read and write multicast enhancement under the controller's key

### LINKED ISSUE #524: fix(network): multicast_enhance_enabled unreadable (wrong controller key) and silently unwritable in Auto mode
Cross-posting from my fork, where the full write-up with evidence lives: rmaher001/unifi-mcp#3

## Summary

Two related problems with `multicast_enhance_enabled` on `unifi_update_wlan`:

1. **Read bug** — the field always returns `null` because the controller reports it under a different key (`mcastenhance_enabled`) than the model reads (`multicast_enhance_enabled`).
2. **Write is silently ineffective when the WLAN is in "Auto" mode** (`setting_preference == "auto"`) — the tool reports success but the controller discards the write, and the failure text points at the write path/firmware rather than the actual cause.

## Proposed fix

- `packages/unifi-core/src/unifi_core/network/models/wlans.py`: read `mcastenhance_enabled` from the raw payload, keep the public field name `multicast_enhance_enabled` for API stability.
- When a write targets an Auto-managed field while `setting_preference == "auto"`, either refuse it up front or include `setting_preference` in the failure text so the cause is discoverable from tool output alone.

Full repro, evidence, and code pointers are in the fork issue linked above. Happy to send a PR if the direction is agreeable.

rmaher001:
Fixes #524. Verified on UniFi Network 10.5.67.

The controller stores Multicast Enhancement as `mcastenhance_enabled`. The model used `multicast_enhance_enabled` on both the read and the write, so the typed read always returned null and the write went out under a key the controller ignores.

That also explains the third symptom: the post-write check compares requested keys against the raw object before and after, and neither side ever contains `multicast_enhance_enabled`, so the field was reported unpersisted on every attempt whatever value was sent.

This maps the field in both directions, keeping `multicast_enhance_enabled` as the public name and accepting the controller spelling as an inbound alias, as `networkconf_id` already does. The same wrong key was read by the GraphQL projection.

Note: part 2 of the issue (Auto mode) is invalid—see the correction there. With the key fixed, the write persists on an `auto` WLAN, so no guard is warranted.

## Maintainer validation (2026-08-13)

I independently reviewed the four-file patch and found no actionable issues. I rebased it cleanly onto current `main` after #527 and ran the following gates:

- `make pre-commit`: passed, including formatting, generation, lint, all package/app tests, registration-mode smokes, generated-file checks, worker tests, and worker typecheck.
- Focused suites: 75 core WLAN model/manager tests, 40 Network tool tests, and 12 API projection/serializer tests passed.
- Network read-only hardware smoke: 121 operations passed, 7 expected argument-discovery skips, 0 failures or exceptions; both WLAN list/detail paths passed.
- Network safe hardware smoke: passed, including the repository's safe disposable lifecycles.
- Targeted disposable WLAN lifecycle: create preview, create disabled SSID, read `false`, update to `true`, read through core and API projections, restore to `false`, delete preview, delete, and cleanup verification all passed.

<details>
<summary>Independent live-smoke output</summary>

```text
network readonly ok: unifi_get_wlan_details
network readonly ok: unifi_list_wlans
connected: true
ok: 121
skipped: 7
failed: 0
exceptions: 0
```

</details>

<details>
<summary>Targeted multicast lifecycle output</summary>

```json
{
  "api_read_false": false,
  "api_read_true": true,
  "cleanup_verified": true,
  "core_read_false": false,
  "core_read_true": true,
  "create_preview_key": "mcastenhance_enabled",
  "created_value": false,
  "deleted": true,
  "restored_value": false,
  "updated_value": true
}
```

</details>

### DIFF
--- a/apps/api/src/unifi_api/graphql/types/network/wlan.py
+++ b/apps/api/src/unifi_api/graphql/types/network/wlan.py
@@ -110,7 +110,7 @@ class Wlan:
             schedule_enabled=raw.get("schedule_enabled"),
             l2_isolation=raw.get("l2_isolation"),
             wlan_band=raw.get("wlan_band"),
-            multicast_enhance_enabled=raw.get("multicast_enhance_enabled"),
+            multicast_enhance_enabled=raw.get("mcastenhance_enabled"),
             dtim_mode=raw.get("dtim_mode"),
             dtim_na=raw.get("dtim_na"),
             dtim_ng=raw.get("dtim_ng"),
--- a/apps/api/tests/graphql/fixtures/network/test_wlans.py
+++ b/apps/api/tests/graphql/fixtures/network/test_wlans.py
@@ -94,3 +94,34 @@ async def test_wlan_detail_policy_disabled_returns_raw_passphrase(tmp_path, monk
     )
     assert body.get("errors") is None, body
     assert body["data"]["network"]["wlan"]["xPassphrase"] == "wifi-secret"
+
+
+@pytest.mark.asyncio
+async def test_wlan_detail_reads_multicast_enhance_controller_key(tmp_path, monkeypatch):
+    """The controller reports multicast enhancement as ``mcastenhance_enabled``.
+
+    Reading the public field name instead left this null for every WLAN, so the
+    GraphQL surface reported the setting as unset even where the controller had
+    a definite value for it.
+    """
+    monkeypatch.setenv("UNIFI_API_DB_KEY", "k")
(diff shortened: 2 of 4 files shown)


sirkirby (approved):
I independently reviewed and validated this fix. The rebased patch passes the full repository gate, all required GitHub checks, Network read-only and safe smoke, and a disposable WLAN multicast create/update/read/restore/delete lifecycle on the lab controller.

PR #529, opened 2026-08-11
feat(network): support LAN-side IPv6 fields on unifi_update_network

rmaher001:
Fixes #525. Verified on UniFi Network 10.5.67.

The Network model carried only WAN-scoped IPv6 settings, so every per-LAN IPv6 key was rejected as unknown and no dual-stack LAN change was possible through the MCP — despite `unifi_get_network_details` returning all of them, which makes them look settable.

Adds the 16 LAN-side fields the controller stores on a corporate network: `ipv6_interface_type`, `ipv6_aliases`, `ipv6_ra_priority`, `ipv6_ra_preferred_lifetime`, `ipv6_client_address_assignment`, the prefix-delegation set (`ipv6_pd_interface`, `ipv6_pd_prefixid`, `ipv6_pd_auto_prefixid_enabled`, `ipv6_pd_start`, `ipv6_pd_stop`) and the DHCPv6 server set (`dhcpdv6_enabled`, `dhcpdv6_allow_slaac`, `dhcpdv6_dns_auto`, `dhcpdv6_leasetime`, `dhcpdv6_start`, `dhcpdv6_stop`).

Two need care and are documented accordingly:
- `ipv6_interface_type` is single-valued; moving a delegated network to `static` releases its delegated prefix, so the confirm-preview warns on that transition as it already does for connectivity-critical WAN fields.
- `ipv6_aliases` replaces wholesale — the same behaviour `deep_merge` already gives `ip_aliases` and `mac_filter_list` — so the field description says to pass every prefix you want to keep.

`ipv6_pd_prefixid` is `str()`-coerced on read, matching `vlan`: it is a hex string, but a controller returning it as a number would fail validation for every network in the site listing.

The `ipv6` read-view section returned 4 keys while 16 more were writable; filled out to match.

Verification — unit: unifi-core 1574, apps/network 973, apps/api 1214; `make check-generated` and ruff clean.

Live:
```
ipv6_aliases now accepted (previously "No valid mutable fields provided for update")
dhcpdv6_leasetime write -> persisted_fields ["dhcpdv6_leasetime"], restored after
```

## Maintainer verification — 2026-08-13

Rebased onto current `main` after #527 and #528. Maintainer follow-up adds strict validation for the three documented LAN IPv6 enums and classifies `ipv6_ra_enabled` with the LAN fields across MCP and API surfaces.

<details>
<summary>Independent branch-local and controller-backed evidence</summary>

```text
Targeted enum/discovery/apply-revert probe:
[...]

PR #531, opened 2026-08-11
feat(network): expose WLAN setting_preference read-only

### LINKED ISSUE #530: WLAN setting_preference (Advanced: Auto/Manual) is not exposed at all
A WLAN's `setting_preference` — the UI's **Advanced: Auto | Manual** toggle — is absent from the WLAN model, so it is neither readable as a typed field nor visible on the REST/GraphQL surface. `unifi_get_wlan_details` returns it only because that tool passes the raw controller object through.

It matters because that toggle decides who owns a set of advanced settings. Measured on Network 10.5.67 by flipping one SSID and diffing the object, **11 keys exist only in Manual** and are omitted entirely while the SSID is on Auto:

`advertise_device_name_in_beacon`, `bc_filter_enabled`, `bc_filter_list`, `dns_assistance_mode`, `hotspot2conf_enabled`, `iot_channel_lock`, `iot_dtim_lock`, `mlo_enabled`, `sae_anti_clogging`, `sae_sync`, `schedule_with_duration`

So a caller reading a WLAN cannot currently tell whether those settings are absent because they are unset or because the controller is managing them.

## Proposed: expose it read-only

Read-only rather than mutable, for two reasons found while testing:

1. **Changing the mode makes the controller rebuild the WLAN object.** `external_id` was regenerated on every transition (three flips, three distinct values), and the first Manual→Auto also normalised away stale keys left over from earlier edits. Transitions are idempotent after that, but the first one on any given SSID is not predictable from the outside.
2. **The set of Auto-managed settings is controller-version specific.** Ubiquiti can change what Auto covers in any release, so anything keyed on that list — including a "switch to Manual before writing" helper — would be stale-by-design after an upgrade.

None of the 11 Auto-managed keys is currently writable through `unifi_update_wlan`, so no write path needs guarding today. Exposing the mode read-only closes the visibility gap without adding a footgun.

Happy to send a PR — one is ready.

rmaher001:
Closes #530.

The UI's per-SSID "Advanced: Auto | Manual" toggle (`setting_preference`) was absent from the WLAN model, so it was not a typed field and did not appear on the REST or GraphQL surface. `unifi_get_wlan_details` showed it only incidentally, by passing the raw controller object through.

The mode decides who owns a group of advanced settings. Measured on Network 10.5.67 by flipping one SSID and diffing the object, 11 keys exist only under Manual and are omitted while the SSID is on Auto (`bc_filter_enabled`, `mlo_enabled`, `sae_sync`, `hotspot2conf_enabled`, `iot_channel_lock`, `iot_dtim_lock`, `dns_assistance_mode`, plus mDNS/schedule and beacon-name keys). Without the mode a caller cannot distinguish "unset" from "controller-managed".

Read-only, deliberately. Changing the mode makes the controller rebuild the WLAN — `external_id` was regenerated on every transition observed, and the first Manual→Auto also normalised away keys left by earlier edits — and which settings Auto covers is controller-version specific. None of the 11 is currently writable through this server, so nothing needs guarding.

Verified against a live controller (Network 10.5.67): REST `GET /v1/sites/default/wlans` and GraphQL `network.wlans { settingPreference }` both return `auto` / `manual` matching the UI.

## Maintainer verification — 2026-08-13

Rebased onto current `main` after #527, #528, and #529. The maintainer follow-up preserves the multicast test coverage added by #528, verifies read-only enforcement through the public create/update validators, and directly covers both API projection values.

<details>
<summary>Independent branch-local and controller-backed evidence</summary>

```text
Targeted live controller and API projection probe:
{"api_projection_verified": 5, "controller_wlan_count": 5, "modes": {"manual": 5}, "read_only_guard_verified": true, "success": true}

Network live smoke — readonly:
exit 0; connected=true; ok=121; skipped=7

Network live smoke — safe:
exit 0; connected=true; ok=159; pending_approval=34; skipped=52
created_resources=3; cleaned_resources=3

API resource smoke — WLAN endpoint:
api-resources pass: /v1/sites/default/wlans http=200 items=5

Full repository gate — make pre-commit:
exit 0
unifi-core: 1599 passed
network: 995 passed
protect: 503 passed
access: 206 passed
api: 1220 passed
worker: 81 tests passed
generated references: 39 sections checked, no drift
```

The full API-resource harness also reports the same client/device pagination-order parity mismatch on unchanged `main`; the WLAN endpoint passes on both trees.

</details>

### DIFF
--- a/apps/api/docs/graphql-reference.md
+++ b/apps/api/docs/graphql-reference.md
@@ -2155,6 +2155,7 @@ type VpnServerPage {
 type Wlan {
   id: ID
   name: String
+  settingPreference: String
   enabled: Boolean!
   security: String
   networkId: String
--- a/apps/api/openapi.json
+++ b/apps/api/openapi.json
@@ -8614,6 +8614,17 @@
             ],
             "title": "Security"
           },
+          "setting_preference": {
+            "anyOf": [
+              {
+                "type": "string"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Setting Preference"
+          },
           "uapsd_enabled": {
             "anyOf": [
               {
--- a/apps/api/src/unifi_api/graphql/schema.graphql
+++ b/apps/api/src/unifi_api/graphql/schema.graphql
@@ -2144,6 +2144,7 @@ type VpnServerPage {
 type Wlan {
   id: ID
   name: String
+  settingPreference: String
   enabled: Boolean!
   security: String
   networkId: String
--- a/apps/api/src/unifi_api/graphql/types/network/wlan.py
+++ b/apps/api/src/unifi_api/graphql/types/network/wlan.py
@@ -32,6 +32,7 @@ def _get(obj: Any, *keys: str, default: Any = None) -> Any:
 class Wlan:
     id: strawberry.ID | None
     name: str | None
+    setting_preference: str | None
     enabled: bool
     security: str | None
     network_id: str | None
@@ -87,6 +88,7 @@ class Wlan:
         return cls(
             id=raw.get("_id") or raw.get("id"),
             name=raw.get("name"),
+            setting_preference=raw.get("setting_preference"),
             enabled=bool(raw.get("enabled", False)),
             security=raw.get("security"),
             network_id=raw.get("networkconf_id") or raw.get("network_id"),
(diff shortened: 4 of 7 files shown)


sirkirby (approved):
I independently reviewed this read-surface addition, rebased it onto current main, and added direct coverage for public write rejection plus both API projection values. The final head passed focused tests, the full repository gate, branch-local controller/API projection verification, read-only and safe live smoke, and all 14 fresh GitHub checks. Approving for squash merge.

Issue #533, opened 2026-08-11
Expose typed site-wide mDNS settings tools


rmaher001:
> **Scope revised.** This issue originally asked for a generic `unifi_get_setting(section)` / `unifi_update_setting(section, settings_data)` accessor. That proposal is withdrawn — see the maintainer's ruling below. The reasoning holds: an arbitrary-section setter would bypass the typed field models, the mutable/read-only boundaries, the domain safety warnings and the field-symmetry checks the rest of the project relies on, and `SystemManager.update_settings` injects `_id` and `key` without merging the current section object first, so it does not satisfy the project's fetch-merge-put requirement for partial updates. The body below is rewritten to the narrower typed scope. The original text is preserved in the edit history.

## Summary

There is no way to read or change site-wide mDNS from a session. `NetworkModel.mdns_enabled` is correctly read-only and, as of #535, documented as a stale-able mirror of the site-wide scope rather than per-network state — so a session can observe that mDNS applies to a network, correctly cannot change it there, and has nowhere else to go.

## What exists today

`SystemManager.get_settings("mdns")` reaches the site-wide mDNS section and returns its structured settings (confirmed by the maintainer against a live controller). Nothing is registered on top of it. Grepping for `mdns` returns only the read-only network field and its description.

## Requested

Two narrowly scoped typed tools, backed by a shared `MdnsSettings` model:

- `unifi_get_mdns_settings()`
- `unifi_update_mdns_settings(...)`

Per the maintainer's outline, the model carries known fields and modes with explicit mutability, and the write path gets preview/confirmation, response redaction, and fetch-merge-put preservation so a partial update cannot drop fields the caller did not name.

## Why it matters

Site-wide mDNS is the source of truth that `mdns_enabled` mirrors. Without a tool for it, the read-only field is a dead end: it can be read, it can be wrong for hours after the site-wide setting changes, and the setting that governs it is unreachable. #535 documents that trap; this closes it.

## Notes

- The generic-accessor discussion is retained above and in the comment thread for anyone who finds this issue looking for `/get/setting/{section}`.
- Sections such as `guest_access`, `ntp`, `mgmt` and `rsyslogd` are equally unreachable, but each should get its own typed treatment on its own merits rather than riding on a generic surface. Not requested here.
- Holding off on an implementation PR until the scope here is confirmed.

sirkirby:
[...]

PR #535, opened 2026-08-12
docs(network): correct mdns_enabled description

### LINKED ISSUE #534: mdns_enabled is documented as unwritable, but not as a stale mirror of the site-wide setting
`Network.mdns_enabled` is documented as read-only, but not as unreliable to *read*. It is a stored mirror of the site-wide mDNS scope, and the mirror can lag.

Current description in `packages/unifi-core/src/unifi_core/network/models/networks.py`:

> Whether mDNS reflection is enabled on this network (read-only via this tool). On UniFi OS 5+, the controller accepts this field on the per-network write path but silently ignores it. Per-network mDNS is not currently settable via the API.

That covers the write. It says nothing about the read being derived.

## What was observed

On Network 10.5.67, mDNS is configured site-wide, and the per-network `mdns_enabled` key mirrors whether the network is in scope. A network had been out of scope for **many hours** while `unifi_get_network_details` still returned `mdns_enabled: true`. After an unrelated write to that network, the same read returned `false` — the correct value.

Read as per-network state, that looks exactly like the write clobbered the setting. It didn't; the stale mirror re-synced. It cost me a wrong diagnosis before the site-wide setting was checked.

## Two smaller inaccuracies in the same string

1. "the controller accepts this field ... but silently ignores it" is true of the controller, but a caller of `unifi_update_network` never gets that far — `_validate_payload` rejects it first with `Network field(s) are read-only and cannot be updated: mdns_enabled`. The `to_controller_update` silent-drop path is unreachable from the tool, so "silently ignores" is misleading about what a caller actually sees.
2. "UniFi OS 5+" is the wrong version axis for the site-wide behaviour — the observation is on a Network Application version.

## Suggested

Say that the key mirrors the site-wide scope, can go stale, and is not the source of truth; name both layers of the write behaviour; and note that no tool here currently exposes the site-wide mDNS setting (`unifi_get_site_settings` is shaped down to `site_id`/`name`/`role`/`country`, and `GatewaySettings` has no mDNS key). Related: #533.

Happy to send a PR.

rmaher001:
Fixes #534.

## What changed

- Corrects the `Network.mdns_enabled` description: the field is a read-only, non-authoritative mirror of the current site-wide mDNS setting and can remain stale for hours.
- Makes that warning discoverable through `unifi_get_network_details`, the generated MCP manifest, GraphQL SDL/reference documentation, and the REST/OpenAPI schema.
- Preserves Strawberry field descriptions when the API server derives Pydantic response models, with a regression test at the REST-schema boundary.
- Regenerates the affected MCP, GraphQL, and OpenAPI artifacts.

## Why

The old text documented that the field is unwritable but not that the read itself is derived. Read as per-network state, an unrelated network write that happens to re-sync the mirror can look like it clobbered mDNS.

The warning now names both write layers accurately: the controller silently ignores the field on the per-network write path, while `unifi_update_network` rejects it as read-only before any mutation. It also avoids the inaccurate UniFi OS version claim and directs readers to the site-wide setting as the source of truth.

This PR intentionally does not add general site-settings support. #533 remains the separate broader request for that capability.

## Testing

Richard verified the original diagnosis on a live Network 10.5.67 controller: list/details reads were unaffected, and `unifi_update_network` rejected `mdns_enabled` without mutating the network.

I rebased the branch onto current `main`, added the public-surface propagation and regression coverage, and independently verified the repaired branch.

<details>
<summary>Maintainer verification output</summary>

```text
$ uv run pytest apps/api/tests/graphql/test_pydantic_export.py apps/api/tests/graphql/test_sdl_drift.py apps/api/tests/graphql/test_openapi_drift.py apps/api/tests/graphql/test_docs_drift.py apps/api/tests/graphql/test_openapi_docs_drift.py -q
...........                                                              [100%]
11 passed in 1.09s

$ make lint
All checks passed!

$ make check-generated
Done. Checked 39 sections; no drift.

$ make pre-commit
1599 passed
359 passed, 1 warning
33 passed
111 passed, 1 warning
81 passed
995 passed, 1 warning
503 passed, 1 warning
206 passed, 1 warning
1221 passed
Done. Checked 39 sections; no drift.

$ MCP discovery smoke
# Real stdio sessions covering index search, explicit load + tools/list,
# and unifi_execute auto-promotion + tools/list.
discovery_paths=pass index=pass preload=pass lazy_execute_promotion=pass

$ uv run --package unifi-network-mcp python scripts/live_smoke.py --server network --phase safe
records=245 ok=159 pending_approval=34 skipped=52 unexpected_failures=0
created_resources=3 cleaned_resources=3
gateway_setting_changed=true siblings_preserved=true reverted=true

$ uv run --package unifi-api-server python scripts/live_api_smoke.py --output /tmp/unifi-pr535-live-api.json
Live smoke: products=[network, protect, access] retry=3
Result: 36/36 passed, 0 failed
```

</details>

### DIFF
--- a/apps/api/docs/graphql-reference.md
+++ b/apps/api/docs/graphql-reference.md
@@ -1061,6 +1061,10 @@ type Network {
   igmpSnooping: Boolean
   igmpQuerierSwitches: JSON
   igmpFloodUnknownMulticast: Boolean
+
+  """
+  Whether mDNS is enabled for this network. Read-only and not authoritative: mDNS is configured site-wide on current UniFi Network versions, and this key only mirrors that site-wide scope — the mirror can go stale for hours, so treat the site-level setting as the source of truth (this server does not expose it). The controller accepts this field on the per-network write path but silently ignores it; unifi_update_network rejects it as read-only.
+  """
   mdnsEnabled: Boolean
   networkIsolationEnabled: Boolean
   internetAccessEnabled: Boolean
--- a/apps/api/openapi.json
+++ b/apps/api/openapi.json
@@ -4657,6 +4657,7 @@
                 "type": "null"
               }
             ],
+            "description": "Whether mDNS is enabled for this network. Read-only and not authoritative: mDNS is configured site-wide on current UniFi Network versions, and this key only mirrors that site-wide scope \u2014 the mirror can go stale for hours, so treat the site-level setting as the source of truth (this server does not expose it). The controller accepts this field on the per-network write path but silently ignores it; unifi_update_network rejects it as read-only.",
             "title": "Mdns Enabled"
           },
           "name": {
--- a/apps/api/src/unifi_api/graphql/pydantic_export.py
+++ b/apps/api/src/unifi_api/graphql/pydantic_export.py
@@ -12,7 +12,7 @@ import datetime
 from typing import Any, get_args, get_origin
 
 import strawberry
-from pydantic import BaseModel, ConfigDict, create_model
+from pydantic import BaseModel, ConfigDict, Field, create_model
 from strawberry.types.base import StrawberryList, StrawberryOptional
 from strawberry.types.private import StrawberryPrivate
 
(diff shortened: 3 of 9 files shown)


sirkirby (approved):
I rebased this onto current main, propagated the mDNS warning through MCP, GraphQL, and OpenAPI discovery, added a REST-schema regression test, and regenerated the affected artifacts. I independently ran the full pre-commit gate, both real stdio discovery paths, the network safe smoke with cleanup/revert verification, and the API live smoke (36/36). Fresh CI is green across all 14 checks. The broader site-settings capability remains correctly separated in #533. Approved.

Issue #71, opened 2026-02-18
Feature: Add device radio configuration tool (tx_power, channel, channel_width)


Milofax:
## Summary

The MCP server currently has no tool to update device radio settings (TX power, channel, channel width, min RSSI, etc.). While `unifi_update_wlan` and `unifi_update_network` exist for WLAN and network configuration, there is no equivalent `unifi_update_device` or `unifi_update_device_radio` for device-level radio settings.

## Use Case

Diagnosing and fixing WiFi issues requires changing radio parameters — the most common being:

- **TX Power Mode** (`auto` / `high` / `medium` / `low` / `custom`)
- **Channel** (specific channel or `auto`)
- **Channel Width** (HT20 / HT40 / HT80 / HT160)
- **Min RSSI** (for band steering / client roaming)

In my case, an AP's 5GHz radio was auto-reduced to 17 dBm while another AP ran at 24 dBm, causing 23% packet retransmissions and frequent client disconnects. The fix is a simple TX power change, but I had to leave the MCP and use the UniFi UI because there's no write tool for device radio config.

## Proposed Tool

```
unifi_update_device_radio
```

### Parameters

| Parameter | Type | Required | Description |
|-----------|------|----------|-------------|
| `mac_address` | string | yes | Device MAC address |
| `radio` | string | yes | `ng` (2.4GHz) or `na` (5GHz) |
| `tx_power_mode` | string | no | `auto`, `high`, `medium`, `low`, `custom` |
| `tx_power` | integer | no | Custom TX power in dBm (when mode is `custom`) |
| `channel` | string/int | no | Channel number or `auto` |
| `ht` | string | no | Channel width: `20`, `40`, `80`, `160` |
| `min_rssi_enabled` | boolean | no | Enable minimum RSSI |
| `min_rssi` | integer | no | Min RSSI threshold (e.g., -70) |
| `confirm` | boolean | yes | Safety confirmation |

### UniFi API

This maps to:

```
PUT /api/s/{site}/rest/device/{device_id}

{
  "radio_table": [
    { "name": "wifi1", "radio": "na", "tx_power_mode": "high" }
  ]
}
```

The `radio_table` is an array — the tool should merge the update into the existing radio_table rather than replacing it, to avoid accidentally resetting the other radio.

## Additional Context

- The existing `unifi_list_devices` and `unifi_get_device_details` tools already return `radio_table` data, so the read side is covered
- This is probably the most common operational change network admins need to make via automation
[...]

PR #536, opened 2026-08-13
feat(network): serve radio config for any device that has radios

christensenjames:
## Summary

`unifi_get_device_radio` reports no radios for a UniFi gateway that has built-in WiFi, because `DeviceManager.get_device_radio()` gates on `device_type.startswith("uap")` (`packages/unifi-core/src/unifi_core/network/managers/device_manager.py:201`) and a gateway falls through to `return None`. This replaces that gate with a presence check on `radio_table`:

```python
device = await self.get_device_details(device_mac)  # raises on miss
if "radio_table" not in device.raw:
    return None
```

Both radio tools are affected, not just the read. `unifi_update_device_radio` resolves its target band by calling `get_device_radio()` first (`apps/network/src/unifi_network_mcp/tools/devices.py:516`) and errors out when it gets `None`, so the same gate blocks gateway writes.

The new predicate is the one this package already uses. `read_views.shape_device_details` gates its radio section on `"radio_table" in raw` (`read_views.py:517`) with no device-type check, which is why `unifi_get_device_details` already returns `radio_summary` for a gateway that `unifi_get_device_radio` refuses. The two now agree.

Serving gateway radios also surfaces that controller radio rows differ in shape by device class: an AP row carries an int `channel` and a string width (`"ht": "20"`), while a gateway row carries `"channel": "auto"` on an automatic radio and an int width (`"ht": 160`). Rather than letting each layer see whichever shape the device emits, both fields now normalise to one canonical form at the controller boundary — `channel` as int with 0 meaning auto (the convention `DeviceRadio.channel` already documents), `ht` as the MHz string matching `VALID_HT_VALUES` and the update schema. Two helpers in `unifi_core.network.models.devices` (`normalize_radio_channel`, `normalize_radio_ht`) are applied everywhere a raw row is read: `DeviceManager.get_device_radio()`, `radio_from_controller()`, the `read_views` `radio_summary`, and the GraphQL `RadioEntry`. `RadioEntry.ht` widens from `Int` to `String` in the SDL; the `Int` declaration matched neither device class, since every AP row carries the width as a string. The update input schemas already used exactly this representation (`channel: integer`, `ht: string`), so the write contract is untouched.

Nothing that works today changes. A switch has no `radio_table` and still returns `None`, as does a gateway with no radio hardware, and a device with a present-but-empty `radio_table` keeps returning an empty-radios success — pinned by `test_empty_radio_table_still_succeeds`. AP rows pass through the normalisers unchanged (int channel in, int channel out; string width in, string width out).

Descriptions in the manager, both MCP tools, the radio update model, and the GraphQL resolver and type said access-point-only, which is what callers read. The tools manifest, action catalog, plugin skill reference, GraphQL SDL, and GraphQL reference doc are regenerated to match. Related: #71, where this tool pair was added.

## Type of change

- [ ] `fix:` bug fix
- [x] `feat:` new feature

## Scope

- [x] Network MCP server
- [ ] Protect MCP server
- [ ] Access MCP server
- [ ] Relay sidecar
- [ ] Cloudflare Worker / CLI
- [x] API server
- [x] Shared packages
- [x] Plugin packaging
- [x] Documentation

## Checklist

## Testing

```
make -C apps/network test    998 passed
make core-test              1609 passed
make check                  exit 0 (format-check + lint + generated-drift + all suites, incl. 1221 API tests)
```

`make check` subsumes `make pre-commit`'s lint, test, generated-drift, and worker-typecheck steps; `ruff format --check` reports the tree clean. The SDL and GraphQL reference doc were re-exported after the `RadioEntry.ht` change, and the drift gates in `make check` pass, so the committed artifacts match their sources.

The canonical shape is covered at every layer the review asked for: `radio_from_controller()` over the live gateway row (`"auto"`/`160` → `0`/`"160"`) plus direct normaliser tests in the shared model suite, the manager path in `test_returns_radio_data_for_udm` (raw fixtures stay exactly as captured; assertions pin the normalised output for both the gateway-style and AP-style rows), `radio_summary` normalisation in the read-views suite, and a GraphQL fixture test that pushes both raw row shapes through an executed `radios { radio channel ht currentChannel }` query and asserts the typed results with no errors.

### Live read verification

This was verified against a live controller, not only against mocks. `get_device_radio()` ran over every device on one site, from this branch and from `main`, against the same controller in the same session:

| device | `main` | this branch |
|---|---|---|
| 3x AP (`type=uap`) | 2 radios each | 2 radios each |
| 3x switch (`type=usw`) | `None` | `None` |
| gateway, `type=udm`, built-in WiFi | **`None`** | 2.4GHz + 5GHz + 6GHz radios, each joined to its `radio_table_stats` row |

The gateway's 6GHz radio reads back as `channel: 0`, `ht: "160"` (raw: `"auto"`, `160`) with the resolved channel in `current_channel: 37` from its stats row; the 2.4GHz and 5GHz rows are byte-identical to what `main` returns for an AP-shaped row.

### Live gateway write round-trip

The newly reachable gateway write path was exercised against the same gateway with `confirm=true`: a controlled change of `tx_power_mode` on the 6GHz radio (`auto` → `low`), read back, then reverted (`low` → `auto`), read back again. The read-backs diff the full raw `radio_table` against a baseline captured before any write. The changed field landed, every unmodified field and both other radios were preserved, and after the revert the table is equal to the baseline. Because the PUT sends the entire `radio_table`, both writes also round-tripped the untouched `"channel": "auto"` value through the controller, which accepted it — closing the open question from the earlier revision of this PR.

<details>
<summary>Sanitized terminal output</summary>

```
2026-08-16 13:05:21,985 - unifi-network-mcp - INFO - Using bundled default configuration: .../apps/network/src/unifi_network_mcp/config/config.yaml
2026-08-16 13:05:22,052 - unifi-network-mcp - INFO - Attempting to connect to Unifi controller at <controller>...
2026-08-16 13:05:22,067 - unifi-network-mcp - INFO - Pre-login auto-detected controller type: UniFi OS (proxy)
2026-08-16 13:05:22,416 - unifi-network-mcp - INFO - Detected UniFi OS controller (proxy paths required)
2026-08-16 13:05:22,416 - unifi-network-mcp - INFO - Successfully connected to Unifi controller at <controller> for site 'default'
2026-08-16 13:05:22,642 - unifi-network-mcp - INFO - Radio '6e' updated on device 8c:xx:xx:xx:xx:0a: ['tx_power_mode']
2026-08-16 13:05:32,792 - unifi-network-mcp - INFO - Radio '6e' updated on device 8c:xx:xx:xx:xx:0a: ['tx_power_mode']
baseline captured: 3 radios; 6e tx_power_mode='auto', 6e channel='auto'

--- preview (confirm=False) ---
{
  "success": true,
  "requires_confirmation": true,
  "action": "update",
  "resource_type": "device_radio",
  "resource_id": "8c:xx:xx:xx:xx:0a/6e",
  "preview": {
    "current": {
      "tx_power_mode": "auto"
    },
    "proposed": {
      "tx_power_mode": "low"
    }
  },
  "message": "Will update tx_power_mode on device_radio 'Router (6GHz)'. Set confirm=true to execute.",
  "resource_name": "Router (6GHz)",
  "warnings": [
    "Device radio will restart briefly after changes are applied.",
    "Connected clients may experience a brief disconnection."
  ]
}

--- apply (confirm=True): 6e tx_power_mode auto -> low ---
{
  "success": true,
  "message": "Radio '6e' (6GHz) updated on Router (8c:xx:xx:xx:xx:0a).",
  "updated_fields": [
    "tx_power_mode"
  ]
}

read-back diff vs baseline (expect exactly one field):
  wifi2.tx_power_mode: 'auto' -> 'low'

--- revert (confirm=True): 6e tx_power_mode low -> auto ---
{
  "success": true,
  "message": "Radio '6e' (6GHz) updated on Router (8c:xx:xx:xx:xx:0a).",
  "updated_fields": [
    "tx_power_mode"
  ]
}

read-back diff vs baseline after revert (expect none):
  (no differences)

round-trip complete; radio_table restored: True
```

</details>

## Notes for reviewers

The two open items from the first review round are addressed above: the gateway radio shape now has one canonical representation enforced at the controller boundary and covered by shared-model, manager, read-views, and executed-GraphQL tests; and the gateway mutation path has live round-trip proof, including controller acceptance of the round-tripped `"channel": "auto"` value.

The normalisation is read-side only by design. The update input schemas already match the canonical form, and `DeviceManager.update_device_radio()` still merges user fields into the raw `radio_table` and PUTs it back unmodified, so the write payload for APs is byte-identical to what shipped before this PR — no unverified change rides along on the AP write path.

One device still informs the live evidence, a single gateway model with three built-in radios. I have no UDR, UXG, or UDM Pro to check against, but the radio-less-gateway and AP shapes are pinned by tests.

### DIFF
--- a/apps/api/docs/graphql-reference.md
+++ b/apps/api/docs/graphql-reference.md
@@ -1718,7 +1718,7 @@ type RadioEntry {
   name: String
   radio: String
   channel: Int
-  ht: Int
+  ht: String
   txPower: Int
   txPowerMode: String
   currentChannel: Int
--- a/apps/api/docs/graphql-reference.md
+++ b/apps/api/docs/graphql-reference.md
@@ -1149,7 +1149,7 @@ type NetworkQuery {
   """Look up a single device by MAC address."""
   device(controller: ID!, mac: ID!, site: String! = "default"): Device
 
-  """Get the radio configuration for a UniFi access point."""
+  """Get the radio configuration for a UniFi device with radios."""
   deviceRadio(controller: ID!, mac: ID!, site: String! = "default"): DeviceRadio
 
   """Get LLDP neighbors reported by a switch."""
@@ -1713,7 +1713,7 @@ type Query {
   access: AccessQuery!
 }
 
-"""Radio configuration entry on a UniFi access point."""
+"""Radio configuration entry on a UniFi device with radios."""
 type RadioEntry {
   name: String
   radio: String
@@ -2288,7 +2288,7 @@ Read-only access to UniFi Network resources.
 - `contentFilters: ContentFilterPage!`  — List content filters on the given controller/site (paginated).
 - `dashboardStats: [StatPoint!]!`  — Site dashboard timeseries (all-points).
 - `device: Device`  — Look up a single device by MAC address.
-- `deviceRadio: DeviceRadio`  — Get the radio configuration for a UniFi access point.
+- `deviceRadio: DeviceRadio`  — Get the radio configuration for a UniFi device with radios.
 - `deviceStats: [StatPoint!]!`  — Per-device stats timeseries (by MAC).
 - `devices: DevicePage!`  — List devices on the given controller/site (paginated).
 - `dnsRecord: DnsRecord`  — Look up a single DNS record by id.
--- a/apps/api/src/unifi_api/action_catalog.json
+++ b/apps/api/src/unifi_api/action_catalog.json
@@ -3646,7 +3646,7 @@
       "input_schema": {
         "properties": {
           "mac_address": {
-            "description": "MAC address of the access point, in format AA:BB:CC:DD:EE:FF (from unifi_list_devices with device_type='ap')",
+            "description": "MAC address of a device with radios (AP, or UDM/gateway with built-in WiFi), in format AA:BB:CC:DD:EE:FF (from unifi_list_devices)",
             "type": "string"
           }
         },
@@ -6035,7 +6035,7 @@
             "type": "string"
           },
           "mac_address": {
-            "description": "MAC address of the access point, in format AA:BB:CC:DD:EE:FF (from unifi_list_devices)",
+            "description": "MAC address of a device with radios (AP, or UDM/gateway with built-in WiFi), in format AA:BB:CC:DD:EE:FF (from unifi_list_devices)",
             "type": "string"
           },
           "min_rssi": {
--- a/apps/api/src/unifi_api/graphql/resolvers/network.py
+++ b/apps/api/src/unifi_api/graphql/resolvers/network.py
@@ -2043,7 +2043,7 @@ class NetworkQuery:
 
     @strawberry.field(
         permission_classes=[IsRead],
-        description="Get the radio configuration for a UniFi access point.",
+        description="Get the radio configuration for a UniFi device with radios.",
     )
     async def device_radio(
         self,
--- a/apps/api/src/unifi_api/graphql/schema.graphql
+++ b/apps/api/src/unifi_api/graphql/schema.graphql
@@ -1707,7 +1707,7 @@ type RadioEntry {
   name: String
   radio: String
   channel: Int
-  ht: Int
+  ht: String
   txPower: Int
   txPowerMode: String
   currentChannel: Int
--- a/apps/api/src/unifi_api/graphql/schema.graphql
+++ b/apps/api/src/unifi_api/graphql/schema.graphql
@@ -1138,7 +1138,7 @@ type NetworkQuery {
   """Look up a single device by MAC address."""
   device(controller: ID!, mac: ID!, site: String! = "default"): Device
 
-  """Get the radio configuration for a UniFi access point."""
+  """Get the radio configuration for a UniFi device with radios."""
   deviceRadio(controller: ID!, mac: ID!, site: String! = "default"): DeviceRadio
 
   """Get LLDP neighbors reported by a switch."""
@@ -1702,7 +1702,7 @@ type Query {
   access: AccessQuery!
 }
 
-"""Radio configuration entry on a UniFi access point."""
+"""Radio configuration entry on a UniFi device with radios."""
 type RadioEntry {
   name: String
   radio: String
--- a/apps/api/src/unifi_api/graphql/types/network/device.py
+++ b/apps/api/src/unifi_api/graphql/types/network/device.py
@@ -123,7 +123,7 @@ class Device:
         return out
 
 
-@strawberry.type(description="Radio configuration entry on a UniFi access point.")
+@strawberry.type(description="Radio configuration entry on a UniFi device with radios.")
 class RadioEntry:
     name: str | None
     radio: str | None
--- a/apps/api/src/unifi_api/graphql/types/network/device.py
+++ b/apps/api/src/unifi_api/graphql/types/network/device.py
@@ -23,6 +23,7 @@ from typing import TYPE_CHECKING, Annotated, Any
 
 import strawberry
 from strawberry.types import Info
+from unifi_core.network.models.devices import normalize_radio_channel, normalize_radio_ht
 
 if TYPE_CHECKING:
     from unifi_api.graphql.types.network.client import Client
@@ -128,7 +129,7 @@ class RadioEntry:
     name: str | None
     radio: str | None
     channel: int | None
-    ht: int | None
+    ht: str | None
     tx_power: int | None
     tx_power_mode: str | None
     current_channel: int | None
@@ -140,11 +141,11 @@ class RadioEntry:
         return cls(
             name=_get(r, "name"),
             radio=_get(r, "radio"),
-            channel=_get(r, "channel"),
-            ht=_get(r, "ht"),
+            channel=normalize_radio_channel(_get(r, "channel")),
+            ht=normalize_radio_ht(_get(r, "ht")),
             tx_power=_get(r, "tx_power"),
             tx_power_mode=_get(r, "tx_power_mode"),
-            current_channel=_get(r, "current_channel"),
+            current_channel=normalize_radio_channel(_get(r, "current_channel")),
             current_tx_power=_get(r, "current_tx_power"),
             num_sta=_get(r, "num_sta"),
         )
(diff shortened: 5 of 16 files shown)


sirkirby (changes requested):
Thanks James — this is a thoughtful, well-researched change, and the presence-based capability check looks like the right direction. I ran the branch locally and against my reference controller. I found two issues that need to be resolved before merge.

- [ ] **Normalize the captured gateway radio shape across the canonical model and GraphQL surface.**

  The new live fixture contains `channel: "auto"` and `ht: 160`, while `DeviceRadio` currently declares `channel: int | None` and `ht: str | None`. I passed that fixture through `radio_from_controller()` and reproduced two Pydantic validation errors. I also passed the same shape through the actual GraphQL `RadioEntry`; the query returns `channel: null` with `Int cannot represent non-integer value: 'auto'`.

  This cannot stay as a separate pre-existing API issue because this PR broadens the GraphQL contract to devices with radios and makes this exact gateway shape newly reachable. Please establish one canonical representation at the controller boundary and use it consistently across the manager/MCP/model/API paths, then add coverage for the live `"auto"`/integer-width shape in both the shared model tests and the GraphQL fixture tests.

- [ ] **Provide live round-trip proof for the newly reachable gateway mutation, or keep gateway writes gated out.**

  Removing the read gate also opens `unifi_update_device_radio` for gateways because the tool calls `get_device_radio()` before updating. The manager then PUTs the complete raw `radio_table`, including untouched values such as `channel: "auto"`, back to the controller. The PR notes that this write path was not exercised.

  Please run a controlled/no-op update on the same radio-equipped gateway with `confirm=true`, read the device back, and show that the requested field is accepted while every unmodified field and every other radio is preserved. Include sanitized raw terminal output in a `<details>` block in the PR body. My reference controller has no non-AP device with `radio_table`, so I could verify the full read-only suite but could not supply the required gateway-specific mutation proof. The other valid option is to keep gateway updates unsupported in this PR and scope the change to reads.

## Minor Items

- Change the preview warning from “AP radio will restart” to “Device radio will restart.”
- Update the remaining “MAC address of the AP” and “None for non-AP devices” docstrings.

## Validation I ran

- All 14 GitHub checks are green; the branch is current with `main`.
- `make lint` and `make check-generated` pass.
- Focused tests pass: 24 Network radio tests, 33 core device-model tests, and 35 API/GraphQL/serializer tests.
- The full Network read-only live smoke passes, including `unifi_get_device_radio`.
- My targeted live inventory found zero non-AP devices with `radio_table`, which is why the gateway mutation still needs proof from hardware that has built-in radios.

The core idea is solid. Once the live shape has one end-to-end contract and the write path is either proven or explicitly kept gated, this should be in good shape.

christensenjames:
Thank you for the thorough review. Commits 5e7510c and 25611f1 correct both blockers. The PR body contains the new evidence.

**Canonical radio shape.** Two shared helpers in `unifi_core.network.models.devices` normalize the divergent fields at the controller boundary. `normalize_radio_channel` returns an int and maps `"auto"` to 0. This is the convention that `DeviceRadio.channel` already documents. `normalize_radio_ht` returns the width as an MHz string. The string form matches `VALID_HT_VALUES` and the update schema.

Every read path applies the helpers: `DeviceManager.get_device_radio()`, `radio_from_controller()`, the `radio_summary` in `read_views`, and `RadioEntry.from_manager_output()` in GraphQL. In the SDL, `RadioEntry.ht` widens from `Int` to `String`. The `Int` declaration matched neither device class, so this also corrects the AP-side `ht` error that you noted. The channel helper maps an unrecognized string to `None` and writes a warning to the log.

Tests cover each layer:

- The shared model suite runs the live gateway row through `radio_from_controller()` and tests each normalizer directly.
- The manager test asserts the normalized output for the gateway-style row and the AP-style row.
- A read-views test covers the `radio_summary` path.
- A GraphQL fixture test executes `radios { radio channel ht currentChannel }` over both raw shapes and asserts the typed results.

The raw captured fixtures stay exactly as the controller sent them. Only the assertions pin the canonical output. The update input schemas already use this representation (`channel: integer`, `ht: string`). As a result, the write contract and the AP write payload are unchanged.

**Gateway mutation proof.** I ran the controlled round trip on the radio-equipped gateway with `confirm=true`. The test set `tx_power_mode` on the 6GHz radio from `auto` to `low`. It then read the full raw `radio_table` back and compared it with a baseline from before the write. Exactly one field changed. The two other radios did not change.

The test then reverted the field to `auto` and read the table back again. The table was equal to the baseline. Sanitized terminal output is in a `<details>` block in the PR body. The PUT carries the entire `radio_table`. As a result, both writes also carried the untouched `"channel": "auto"` value, and the controller accepted it. This resolves the open question from my original notes.

**Minor items.** The preview warning now reads "Device radio will restart briefly…". I reworded the `MAC address of the AP` and `None for non-AP devices` docstrings. The RF-scan methods keep `ap_mac` because those methods are AP-only.

Verification on the branch: `make -C apps/network test` reports 998 passed. `make core-test` reports 1609 passed. `make check` exits 0. This run includes format-check, lint, the generated-drift gates, and the 1221-test API suite.

sirkirby (approved):
Re-review complete on head 25611f1. The two requested blockers and both minor items are resolved.

- The gateway and AP radio shapes now normalize to one canonical contract across the shared model, manager, read view, and GraphQL projection. The executed GraphQL regression test covers channel auto and integer channel width without serialization errors.
- The supplied live gateway mutation cycle changes exactly the requested tx_power_mode field, preserves every untouched field and radio, and restores the complete raw radio_table to its baseline after the revert.
- The preview warning and remaining radio-specific AP wording were corrected.

I independently reran the Network read-only live smoke against the reference controller: 121 checks passed, 7 expected hardware/fixture skips, and zero failures. I also called unifi_get_device_radio directly and confirmed that the live AP response contains integer channel/current_channel fields and string ht fields. The reference site has no non-AP device with radio_table, so the gateway write uses the documented hardware-gated exception backed by the contributor live round-trip evidence.

Focused verification passed locally: 54 Core tests, 24 Network radio tests, 10 API GraphQL device tests, make lint, and make check-generated. All 14 current GitHub checks are green.

Approved. Release sequencing note: publish the new Core patch first, then raise the Network and API unifi-core floors before tagging either downstream package.

PR #537, opened 2026-08-17
Re-authenticate on LoginRequired in client/device/DPI list paths

hfreire:
Running the network MCP against a self-hosted UniFi controller, every list/read call (`list_clients`, `list_devices`, DPI stats, and anything built on them) starts returning 401 and keeps returning it until the pod is restarted. It reproduces whenever the controller invalidates the session server-side — the one I hit in practice is the controller restarting on a cert renewal, but a plain logout does it too.

The cause is that the collection helpers reach past `ConnectionManager`. `get_clients`/`get_all_clients`/`get_devices`/`get_dpi_stats` call `controller.clients.update()`, `controller.devices.update()`, etc., which go straight to aiounifi's `controller.request()`. Only `ConnectionManager.request()` catches `aiounifi.errors.LoginRequired` and calls `_reauthenticate()`, so on those paths a `LoginRequired` just propagates. `ensure_connected()` doesn't save us either — it only checks the aiohttp socket is still open, which it is; the *session* is what's dead.

Fix is to give those calls the same recovery `request()` already has. Added `ConnectionManager.run_with_reauth(action)` — runs the call, and on `LoginRequired` re-authenticates once (via the existing generation-locked `_reauthenticate`) and retries, treating a second 401 as terminal so a wedged session can't turn into a login storm. The four managers now wrap their `.update()` calls in it. No behavior change on the happy path.

Verified against a live self-hosted controller (classic `/api/login`, `default` site) by building the network image and running the managers directly: baseline lists 31 clients / 4 devices; after forcing a server-side `/api/logout`, the raw `controller.clients.update()` reproduces the `LoginRequired`, while the patched `get_clients()`/`get_devices()` re-authenticate and return the full list (auth generation increments, confirming a real re-login rather than a cache hit).


sirkirby:
Closing this as superseded by #578, which landed the same expired-session diagnosis at a named handler-refresh seam and re-resolves the handler after re-authentication. Your independent reproduction and `run_with_reauth` implementation helped confirm the missing path; thank you, Hugo.

Issue #540, opened 2026-08-18
MAC lookups are case-sensitive: an uppercase MAC reports 'not found' for hardware that exists


rmaher001:
## Summary

MAC lookups compare raw strings. The controller reports MACs lowercase; a caller passing the uppercase form — what is printed on the device label and quoted in most vendor docs — gets "not found" for hardware that is plainly there. It reads as a broken controller rather than a case mismatch.

A MAC is hexadecimal, so `AA:BB:CC:DD:EE:FF` and `aa:bb:cc:dd:ee:ff` are the same address by definition. There is no device for which case distinguishes identity, so this can only ever produce a false negative.

One path already lowercases both sides, which is what keeps this hidden: whichever tool you happen to try first decides whether you believe the bug exists.

## Sites

All verified against `e53dca4`.

| File | Line |
|---|---|
| `packages/unifi-core/.../network/managers/client_manager.py` | 148, 160 |
| `packages/unifi-core/.../network/managers/device_manager.py` | 55 |
| `packages/unifi-core/.../network/managers/event_manager.py` | 55 |
| `packages/unifi-core/.../access/managers/device_manager.py` | 179, 262 |
| `apps/api/src/unifi_api/graphql/resolvers/network.py` | 1931 (`Query.client`), 2037 (`Query.device`) |
| `apps/api/src/unifi_api/routes/streams/network_per_device.py` | 54 |

`get_client_details` backs block / unblock / rename / forget / force-reconnect / authorize-guest / set-ip; `get_device_details` backs reboot / adopt / upgrade / rename / radio / PDU-outlet. A single comparison takes out a whole family at once.

The SSE route fails worst: `200 OK`, a valid `text/event-stream`, then zero events forever, with nothing to indicate why.

The correct-by-accident site, `stats_manager.py:739`:

```python
target = client_mac.lower()
raw = next((c for c in clients if isinstance(c, dict)
            and str(c.get("mac", "")).lower() == target), None)
```

Note it has its own edge: an empty `client_mac` against a record carrying no `mac` matches on `"" == ""`, and `client_mac=None` raises `AttributeError`.

## Proposed fix

Normalise the caller's MAC **once at the entry point**, and use the normalised value for the lookup *and* for everything sent onward — not just the comparison.

The second half matters. Today an uppercase MAC fails at the lookup and dispatches nothing. Fixing only the comparison makes the lookup succeed and then sends the caller's raw string to the controller anyway:

```python
[...]

PR #541, opened 2026-08-18
fix(network,access,api): normalize MAC addresses at every entry point

rmaher001:
Relates to #540 — **deliberately non-closing**, see "What this PR does not fix".

Normalizes the caller's MAC once at the entry point and uses that value for the lookup **and** for everything sent onward.

The second half is the part that matters. Fixing only the comparison is the more dangerous half-measure: the lookup starts succeeding and the caller's raw uppercase string then goes to the controller in a command body (`data={"mac": client_mac, "cmd": "block-sta"}`) or a URL path (`/stat/device/{device_mac}`), so the failure moves later and gets quieter than today's fast not-found. The controller reports lowercase, so lowercase is certainly a form it accepts.

## What it covers

Lookups at the seven sites from the issue, plus the outbound values, the stats and speedtest queries, the ACL and traffic-flow filters, the OON MAC targets, the GraphQL relationship edges, and the cache keys.

`unifi_core.mac` provides `normalize_mac`, `mac_equal` and `looks_like_mac`:

- `mac_equal` guards against `None == None`. That was live: `Device.portClients` compared a device's absent `mac` to a wired client's absent `ap_mac` and returned **every** wired client as that device's port clients.
- `looks_like_mac` lets the Access reboot path tell an address from an opaque `unique_id`, so a `unique_id` caller keeps its single request. A substring check is not enough — `dev-1` contains a separator.
- Access `unique_id` comparisons stay exact. Those are opaque controller identifiers, not addresses, so there is no case-equivalence rule to apply to them.

## Changes from your review

**The partial OON update no longer carries invented policy.** You were right, and it was worse than a MAC bug: `normalize_oon_update_payload` ran the create-path secure normalizer, which fills in an `internet.mode` when the caller gives none. Because the update is merged over the stored policy and PUT back whole, `{"secure": {"enabled": false}}` replaced a stored `BLOCKLIST` with `TURN_OFF_INTERNET` — a narrow app block silently becoming a total internet cut, with the confirmation preview showing the caller's own input rather than the injected value. `secure` is no longer shaped on the update path at all; only `targets` is, which is what this PR is for. A caller who does want a mode changed still passes it through untouched. Covered by tests that assert the complete outgoing PUT against a stored blocklist policy, including a rename that must leave every other field byte-identical.

**The Access API path now treats `Device.id` as the MAC it is.** Confirmed against the installed dependency: `unifi_access_api.Device` declares `id`, `type`, `name`, `alias`, `location_id`, `connected_uah_id`, `is_online`, `is_adopted`, `is_managed` — and no `mac`, so `mac_equal(getattr(device, "mac", None), device_id)` was comparing against `None` on every row and an uppercase lookup raised `UniFiNotFoundError`. The id is now compared with `mac_equal` and reported as `mac` (falling back to `id` only when the controller supplies no `mac` of its own, so a device that has one keeps exactly the value it sent). The bogus `unique_id: device.id` is gone — that key means the controller's topology identifier, and publishing a MAC under it is what sent an address into the reboot URL. Reboot targets now resolve through the proxy topology even when the API client answered the lookup, via one helper shared by the preview and the write, still opportunistic rather than a gate. Tested dual-auth (`has_api_client` **and** `has_proxy`), constructing the real `Device(id=<mac>)` shape rather than a mock that invents a `.mac`.

**A fourth fix, which your blocker uncovered indirectly.** Live QA of the Access change above failed on my own controller: the API client reports the door hub as `1c0b8beef6b5` while the same controller's `devices/topology4` payload writes it `1c:0b:8b:ee:f6:b5`. `mac_equal` normalised case only, so those two spellings of one hub compared unequal - the topology resolution found nothing and fell back to posting the address, and `get_device_configs` answered "not found" for a device `get_device` had just returned. `mac_equal` now compares the twelve hex digits; non-MAC-shaped values keep an exact comparison, so `dev-1` and `dev1` stay distinct, and `normalize_mac` is untouched because it governs what gets *sent*. My unit fixtures had hidden this by writing the address the same way on both sides of every comparison; the new tests use the shape the controller actually returns.

**The manager-boundary tests assert requests, not source text.** The ones that read `inspect.getsource` and checked helper names and their order are gone, replaced by tests that drive the manager and assert the payload it actually issued — including the sibling fields, since a test that only checks the transformed value would pass with the rest of the object dropped. I verified each by removing the corresponding normalization and confirming the matching test fails. The rename regression no longer swallows exceptions: it requires the call to return `True` and to issue exactly one `put` to `/rest/device/<id>` with the expected body, and a second case asserts a genuinely absent MAC still raises and issues no write. Confirmed the old version's hole was real — with the write suppressed, the previous assertion still passed.

## What this PR does not fix

Three `apps/api` paths from #540 still compare MACs with a raw `==`: `Query.client` and `Query.device` in `graphql/resolvers/network.py`, and the per-device SSE filter in `routes/streams/network_per_device.py`.

They were in an earlier revision and were reverted for a release-ordering reason, not a design one: `apps/api` would import `unifi_core.mac`, which exists in no published `unifi-core`, while pinning a released floor — so the pin-alignment job fails on a fresh PyPI install, and raising the floor to an unreleased version is exactly the premature bump that gate rejects. The follow-up is a small PR once the `unifi-core` half is published. #540 stays open until then; I've commented there with the specifics.

## Behavior changes to be aware of

- `forget_client` (and the other command paths) now send the lowercased MAC rather than the caller's string. `apps/network/tests/unit/test_forget_client.py` asserts that.
- A partial OON update no longer sends a `secure.internet.mode` the caller did not provide. This restores the pre-existing update behavior; only `targets` shaping is added.

## Verification

`unifi-core` 1,679 passed, `apps/network` 1,028 passed, `apps/access` 211 passed, `apps/api` 1,222 passed. `ruff format --check` and `ruff check` clean at the `uv.lock` pin (0.16.2); `make check-generated` exit 0; `check_pin_alignment.py` exit 0.

Exercised against a live controller (Network 10.5.67) through a build of this branch: an uppercase client MAC and an uppercase device MAC both resolve and return the normalized lowercase form, while a genuinely absent MAC still fails with `not found` — so the lookup did not become match-anything. Adjacent read tools and the Access product were re-checked on the same build.

Each fix carries a test confirmed to fail without it.


rmaher001:
Heads up on the one red check — **Pin alignment (PyPI install + smoke import)**. It is red on both commits and I should have caught it when I opened this, not after.

Cause is structural rather than a defect in the diff: `apps/api` imports the new `unifi_core.mac` module in four places, and that job installs `unifi-api` from PyPI against `unifi-core[network,protect,access]>=0.4.29`. Published 0.4.29 has no `unifi_core.mac`, so importing every submodule of `unifi_api` fails on a fresh install.

Nothing I can change inside this branch fixes it. Adding the helpers to an existing module instead would fail identically — `from unifi_core.X import mac_equal` still raises at import time against 0.4.29 — so any `apps/api` use of new `unifi-core` code is red until `unifi-core` is released. That matches how the pin bumps land here as their own commits (`chore(deps): require unifi-core 0.4.29 for network and api`).

Two ways forward, whichever suits your release order:

1. **Land as-is and sequence the release** — cut `unifi-core` with `mac.py`, bump the `apps/api` floor, and this goes green.
2. **I split it** — `unifi-core` + `apps/network` in this PR, and the `apps/api` half (the two GraphQL resolvers, the relationship edges on `Device.portClients` / `Client.device`, the per-device SSE filter, and the cache keys) in a follow-up once the release is out.

Say which and I will do it. Everything else is green: 13 of 14 checks, and `make test` covers all six suites locally.

rmaher001:
Pushed `7fe2b28`, which makes the pin-alignment check pass.

The failure was real rather than a CI artifact: `apps/api` imported `unifi_core.mac`, a module that exists in no published `unifi-core`, while pinning `unifi-core>=0.4.29,<0.5`. I confirmed against the published 0.4.29 wheel — `import unifi_core.mac` raises `ModuleNotFoundError` — so a fresh `pip install unifi-api-server` would fail to boot on first import of any module touching it.

It can't be fixed in place. Bumping the floor to `>=0.4.30` is the premature-pin-bump case `check_pin_alignment.py` warns about, since that version is unpublished and `0.4.30.dev*` doesn't satisfy it. So I've taken the remedy the script prescribes — merge and release the upstream first, then add the downstream imports — which also matches this repo's convention of raising floors in separate `chore(deps)` commits to already-released versions.

`apps/api` is therefore restored byte-for-byte to upstream in this PR; the core, network and access work is unchanged and mergeable on its own. The GraphQL edge matching and per-device stream filtering can follow once a core release carries `unifi_core.mac`. Happy to open that follow-up whenever suits.

Verified on a live controller after deploying this exact SHA: a client lookup by uppercase MAC now resolves the same client the controller stores in lowercase.

sirkirby (changes requested):

- [ ] Preserve partial OON secure updates. I reproduced a case where updating only `{"secure":{"enabled":false}}` passes through the create-time secure normalizer at `oon_manager.py:153`, injects `internet.mode="TURN_OFF_INTERNET"`, and then overwrites an existing `BLOCKLIST` mode during the full-object merge/PUT at line 313. That turns a narrow enable/disable update into a materially different network policy, and the injected change is absent from the confirmation preview. Remove secure shaping from this MAC-focused partial-update path, or make it partial-update-aware so omitted secure fields are preserved and previewed accurately. Add a behavioral regression that asserts the complete outgoing request payload against an existing blocklist policy.

- [ ] Correct the Access API/proxy identifier handling. `py-unifi-access` exposes the device MAC in `Device.id` and normally has no separate `mac` attribute, but `_api_device_to_dict` currently labels `Device.id` as `unique_id`, while `get_device` compares `Device.id` case-sensitively and only applies `mac_equal` to the normally absent `.mac`. I confirmed this against the installed dependency and on the live dual-auth controller: API devices use MAC-shaped `id` values, have no separate `mac`, and an uppercase lookup still raises `UniFiNotFoundError`. In the reboot path that miss is caught and the raw MAC is posted instead of resolving the proxy topology identifier. Treat the public API `Device.id` as a MAC, compare it with `mac_equal`, expose it as `mac` rather than proxy `unique_id`, and resolve MAC reboot targets through proxy topology even when the API client is also active. Add dual-auth tests using the real `Device(id=<mac>)` shape.

- [ ] Do not auto-close #540 while its API failures remain. The final commit intentionally restored `apps/api` to upstream, so `Query.client`, `Query.device`, and the per-device SSE filter still use raw case-sensitive equality at `resolvers/network.py:1931`, `resolvers/network.py:2037`, and `routes/streams/network_per_device.py:54`. Either restore the API slice after the required Core release/pin sequencing, or change `Fixes #540` to a non-closing reference and link an explicit follow-up that keeps #540 open until those paths are fixed.

- [ ] Make the new mutation tests behavioral. Several manager-boundary tests in `packages/unifi-core/tests/test_mac.py` only inspect source text for helper names/order, so they pass even if the transformed value is discarded or sibling fields are lost. The new rename regression also catches every exception and can pass without dispatching a request. Assert the actual `ApiRequest` payloads and field preservation for the changed create/update paths, and require rename to complete successfully with the expected PUT.

## Verification

I reviewed exact head `7fe2b28092a8da97e3e640c264ffceed5f74d145`. All 14 GitHub checks are green. Locally, format, Ruff, and generated-artifact checks passed, along with 1,651 Core, 1,023 Network, 206 Access, and 1,221 API tests. Access read-only smoke completed apart from the already-known activity-summary controller error, and the targeted live Access lookup reproduced the API-path failure above. Network live smoke could not connect after three timeout attempts, so no independent Network controller validation or mutation cycle is being claimed here.

rmaher001:
All four blockers addressed. Each was reproduced as a failing test first, and each fix was then checked by removing it and confirming the test goes red — the two you flagged as vacuous were vacuous exactly as you described.

**Partial OON secure updates.** Confirmed and fixed. `{"secure": {"enabled": false}}` did inject `internet.mode="TURN_OFF_INTERNET"` and overwrite a stored `BLOCKLIST` through the merge, with the preview showing the caller's input rather than the injected value. I took the first of your two options: `secure` is no longer shaped on the update path at all. That shaping is a create-time concern and this PR has no business touching it — only `targets` does. A caller who genuinely wants a mode change still passes it straight through. The regression asserts the complete outgoing PUT against a stored blocklist policy, plus a rename that must leave every other field byte-identical.

**Access API/proxy identifiers.** Confirmed against the installed dependency, not just taken on trust: `unifi_access_api.Device` declares `id`, `type`, `name`, `alias`, `location_id`, `connected_uah_id`, `is_online`, `is_adopted`, `is_managed` — no `mac`, so `mac_equal(getattr(device, "mac", None), …)` was comparing against `None` on every row. `Device.id` is now compared with `mac_equal` and reported as `mac`; the `unique_id: device.id` line is gone, since that key means the topology identifier and publishing an address under it is what sent a MAC into the reboot URL. Reboot targets resolve through the proxy topology even when the API client answered, via one helper shared by the preview and the write, still opportunistic rather than a gate. Dual-auth tests construct the real `Device(id=<mac>)` rather than a mock that invents a `.mac`.

One narrowing against your wording: you asked me to expose the id as `mac`, and my first pass also normalised its case, which broke an existing test asserting that a device's own `mac` echoes verbatim. `id` is now a fallback used only when the controller supplies no `mac`, which is what you actually asked for and a smaller change.

**That fix then failed live, and the failure was mine.** Deploying it to my controller showed the API client reporting the hub as `1c0b8beef6b5` while the same controller's topology writes it `1c:0b:8b:ee:f6:b5`. `mac_equal` normalised case only, so one hub compared as two devices: the resolution found nothing and fell back to posting the address, and `get_device_configs` answered "not found" for a device `get_device` had just returned. `mac_equal` now compares the twelve hex digits. Values that are not MAC-shaped keep an exact comparison, so `dev-1` and `dev1` remain distinct, and `normalize_mac` is deliberately unchanged — it governs what gets *sent*, where rewriting a separator would alter the request. My fixtures had hidden this by writing the address identically on both sides of every comparison.

**#540 is no longer auto-closed.** `Fixes #540` is now a plain reference, and I've commented on the issue naming the three `apps/api` paths that remain case-sensitive — `Query.client`, `Query.device`, and the per-device SSE filter — with the release-ordering reason they cannot be fixed from inside this PR. The follow-up is a small PR once the `unifi-core` half is published.

**Behavioral tests.** The `inspect.getsource` tests are gone. Each is replaced by one that drives the manager and asserts the request actually issued, including the sibling fields, since a test checking only the transformed value would pass with the rest of the object dropped. The rename regression no longer swallows exceptions: it requires `True` and exactly one `put` to `/rest/device/<id>` with the expected body, and a second case asserts an absent MAC still raises and issues no write. The old version's hole was real — with the write suppressed, it still passed.

**Verification.** `unifi-core` 1,679, `apps/network` 1,028, `apps/access` 211, `apps/api` 1,222 passed; ruff format/check clean at the `uv.lock` pin; `make check-generated` and `check_pin_alignment.py` both exit 0. Exercised on a live controller from a build of this exact branch, layered with #558/#559/#549 so the pending work is tested together: bare-uppercase and colon-separated forms both resolve on the Access API path and the proxy topology path, the network device lookup resolves a bare uppercase MAC and reports the canonical spelling, and an absent MAC still returns not-found.

One coverage gap I'd rather state than paper over: my controller has no OON policies, so the partial-update fix could not be exercised live at all. It rests on the payload-level tests above.

sirkirby (approved):
Approved at exact head `80247f679f059de676ca3a01300d63eedb84fa5d`.

The original blockers are resolved, and I completed the remaining maintainer cleanup on the contributor branch: opaque non-MAC IDs remain exact/case-sensitive, Access device identity fields survive both API and proxy projections, and the affected MCP/API generated contracts are current.

Verification: `make pre-commit` passed; all 14 GitHub checks are green; independent safe live smoke completed with 42 Access checks and 159 Network checks successful and no failures; targeted uppercase Access-device, Network-device, and client lookups all succeeded. The known Access activity-histogram controller error remains an explicitly skipped pre-existing controller issue.

PR #542, opened 2026-08-18
fix: preserve tool annotations through relay discovery

davidkmiller:
## What changed

- Preserve explicitly declared MCP annotation fields in the shared runtime registry for permissioned tools and meta-tools.
- Include those annotations in tool-index responses and carry them through relay registration and catalog-update serialization.
- Treat `*_tool_index` as the canonical catalog whenever a server exposes one, including after a lazy server has loaded some direct tools.
- Overlay matching `tools/list` titles and annotations onto indexed entries for compatibility with older eager servers whose index does not contain that metadata.
- Detect changes across the full relay tool metadata, ignoring catalog order, instead of comparing tool names alone.
- Retry catalog updates that could not be sent and register the latest catalog after a relay reconnect.

## Why

Relay authorization depends on MCP annotations such as `readOnlyHint`. Those hints could be lost before indexed catalogs reached the relay, causing read-only tools with location parameters to be treated as mutations.

The discovery lifecycle also has two important edge cases. A warmed lazy server exposes both meta-tools and a partial set of direct tools, so the presence of a direct tool cannot identify eager mode; the index must remain canonical for complete membership. Conversely, independently deployed older eager servers may expose annotations only through `tools/list`, so matching direct metadata supplements the indexed entries without changing indexed membership.

Finally, periodic refresh must compare and advertise complete tool metadata. The relay now retains the last successfully advertised snapshot, retries failed sends, and uses the current catalog for every reconnect registration.

## Regression coverage

- direct, indexed, permissioned, and meta-tool annotations survive registry and relay serialization
- unannotated tools remain unannotated
- warmed lazy discovery retains the complete indexed catalog
- legacy eager index entries inherit annotations and titles from matching direct tools
- metadata-only catalog changes are sent while order-only changes are ignored
- failed catalog updates retry and reconnect registration uses the latest catalog
- real stdio sessions invoke the actual index in lazy, warmed-lazy, eager, and meta-only modes for Network and Protect

## Validation

- `uv run --package unifi-mcp-shared pytest packages/unifi-mcp-shared/tests -q` — 363 passed
- `uv run --package unifi-mcp-relay pytest packages/unifi-mcp-relay/tests -q` — 122 passed
- `make pre-commit` — passed formatting, generated-artifact checks, lint, all Python and Worker tests, real stdio metadata smoke, generated-drift checks, and Worker typecheck
- real stdio metadata smoke passed Network and Protect in lazy, warmed-lazy, eager, and meta-only modes


sirkirby (approved):
Thanks, David — this is a strong first contribution. The original annotation plumbing was well targeted and came with useful regression coverage.

I reviewed the path end to end and pushed follow-up commit `0a4eb215` to your branch after finding a few relay lifecycle edge cases:

- I kept `*_tool_index` canonical whenever it is available. A lazy server starts exposing direct tools after they are loaded, so direct-tool presence cannot safely identify eager mode; otherwise the next periodic refresh could replace the full catalog with only the warmed subset.
- I overlaid matching `tools/list` metadata onto indexed entries. This preserves annotations and titles when a new relay talks to an independently deployed older eager server whose index does not yet contain them.
- I changed refresh detection from tool-name sets to complete, order-independent `ToolInfo` snapshots, so annotation-only changes now reach the Worker without order-only churn.
- I retained the last successfully advertised snapshot, retry failed updates, and provide the current catalog on every reconnect registration so stale metadata cannot come back after a disconnect.
- I added regression tests for warmed lazy discovery, legacy eager compatibility, metadata-only updates, order stability, failed-send retry, and reconnect registration.
- I extended the real stdio metadata smoke to invoke the actual tool index and warm lazy mode. It now verifies index membership, schemas, and annotations for Network and Protect in lazy, warmed-lazy, eager, and meta-only modes.

Validation I ran on the final head:

- shared suite: 363 passed
- relay suite: 122 passed
- full `make pre-commit`: passed formatting, lint, generated artifacts/drift, every Python and Worker suite, real stdio metadata smoke, and Worker typecheck
- API suite within that gate: 1,222 passed
- two independent post-fix review passes: no blockers
- all 13 fresh GitHub checks: passed

No controller mutation or hardware-dependent path is involved here; the real process-level MCP smoke covers the relevant runtime boundary. I’m approving this version. Thank you for the thoughtful contribution and for making the branch available for maintainer edits.

PR #545, opened 2026-08-20
fix: load .env from the working directory under installed packages

### LINKED ISSUE #544: fix: .env is never loaded under a pip/uvx install (works only from a source checkout)
### Preflight checklist

- [x] I am using the latest released version, or I confirmed the issue still exists on `main`.
- [x] I have removed credentials, tokens, API keys, IP addresses, and other private data from the report.

### Affected component

Network MCP server

### Version

unifi-network-mcp 0.20.9 and 0.27.1 (latest); also present on main@e5a4d96

### Controller hardware / host

UCG Max / UCG Ultra / UCG Fiber

### UniFi OS version

5.1.19

### Network application version

10.5.67

### Protect application version

_No response_

### Access application version

_No response_

### Site setup

Single site, default config

### Install method

uvx (Claude Code plugin marketplace)

### OS where the MCP client runs

macOS 26.5.2

### MCP client

Claude Code 2.1.236 / Claude Desktop (bundled Claude Code 2.1.234)

### AI model used

Claude Opus 5 and Claude Fable 5

### Exact prompt

_No response_

### Tool calls observed

```shell

```

### Raw tool output (sanitized)

```json

```

### Expected behavior

A `.env` in the directory the server is started from is read, so the controller host and credentials can be configured per project. That is what `.env.example` sets up in a checkout, and MCP clients spawn stdio servers with the user's project as the working directory, so it is the natural place for per-project config.

### Actual behavior

Under a pip/uvx install the `.env` is silently ignored and the server exits with "No controller host configured". The bare `load_dotenv()` call relies on python-dotenv walking up from the calling file, so in a source checkout it reaches the repo root, but in an installed package it starts inside site-packages (for uvx, deep in the uv cache) and finds nothing.

### Steps to reproduce

With no `UNIFI_*` variables in the environment:

```
mkdir demo && cd demo
printf 'UNIFI_NETWORK_HOST=demo\n' > .env
uvx unifi-network-mcp==0.20.9
```

The server refuses to start. The same `.env` at the root of a source checkout is picked up (I verified both on main@e5a4d96), so the bug only shows on installed packages.

### Relevant logs or output

```shell
2026-08-19 22:14:15,299 - unifi-network-mcp - ERROR - [unifi-network] No controller host configured — refusing to start.
```

### Additional context

Root cause is the no-arg `load_dotenv()` at `apps/network/src/unifi_network_mcp/bootstrap.py:30`; the access and protect bootstraps have the same pattern. I have a small fix ready (an additional `load_dotenv(Path.cwd() / ".env", override=False)` after the existing call, so real environment variables still win) and will open a PR referencing this issue. It matters for MCP deployments in particular because passing env through client config can be unreliable (Claude Code, for example, does not apply project-scope settings `env` to MCP spawns in its non-interactive modes: anthropics/claude-code#11927). I suspect this may also be behind some "server configured but no tools" reports like #200.

### Code of conduct

gwtaylor:
Fixes #544.

## Problem

The three server bootstraps (network, access, protect) call `load_dotenv()` with no arguments. python-dotenv's `find_dotenv()` walks up from the *calling file's* directory, so this works in a development checkout — walking up from `apps/<app>/src/...` reaches the repo root, where `.env.example` suggests a `.env` should live. Under a pip or uvx install, however, the walk starts inside `site-packages` (for uvx, deep in the uv cache) and silently finds nothing, so an end user's `.env` is never read no matter where they put it.

This bites MCP deployments specifically. MCP clients (Claude Code, Codex, and others) spawn stdio servers with the user's project as the working directory, and passing environment variables through client config can be unreliable — Claude Code, for example, does not apply project-scope settings `env` to MCP spawns in its non-interactive modes (anthropics/claude-code#11927), which is how the Claude Desktop app runs. A per-project `.env` next to the project the server is being used from is the natural escape hatch, and the code already appears to intend to support it — it just never finds the file when installed as a package.

## Fix

After the existing `load_dotenv()` (unchanged, so development checkouts behave exactly as before), also load `.env` from the process working directory:

```python
load_dotenv()
load_dotenv(Path.cwd() / ".env", override=False)
```

`override=False` keeps real environment variables authoritative, so any deployment that already supplies `UNIFI_*` through the environment is unaffected; the cwd `.env` only fills in values that are otherwise absent. Applied identically to the network, access, and protect bootstraps.

## Testing

- Reproduced the bug on installed packages 0.20.9 and 0.27.1 (latest): with no `UNIFI_*` in the environment and a `.env` in the working directory, `uvx unifi-network-mcp` exits with `No controller host configured — refusing to start`. A source checkout of `main` with the same `.env` at the repo root picks it up, confirming the checkout/installed asymmetry.
- With the fix, importing the bootstrap with a `.env` in the working directory picks up the value; with the variable present in the environment, the environment value wins over the `.env` (precedence unchanged); with no `.env` in the working directory, behaviour is identical to before.
- Full test suites pass: `make test` in apps/network (998 passed), apps/protect (503 passed), apps/access (206 passed).

Verified end to end against a live controller: with this fix, `uvx unifi-network-mcp` spawned by an MCP client with only a project-root `.env` reaches "Lazy loader ready" and serves tools.


sirkirby (changes requested):
## Hard blockers

- [ ] Add durable regression coverage for the installed-entrypoint behavior and precedence. I independently reproduced an important test trap: released `0.27.1` already reads the cwd `.env` when the bootstrap is imported through `python -c`, because python-dotenv treats that as interactive, while the actual `uvx unifi-network-mcp` console entry point still ignores the file and exits with code 5. A plain import subprocess therefore passes before this fix. Please cover the real console path, or extract a shared dotenv bootstrap helper and test that the cwd file is loaded, an existing environment value wins, and an absent cwd file changes nothing.

- [ ] Apply the same correction to the relay entry point. `packages/unifi-mcp-relay/src/unifi_mcp_relay/__main__.py` still calls bare `load_dotenv()`, and the root `.env.example` includes `UNIFI_RELAY_*` configuration. I built the relay wheel from this head and ran it with valid relay values only in the cwd `.env`; it ignored them and exited with all three variables missing. Since this is the same installed-package root cause, please include relay coverage as well. A shared helper would avoid maintaining four copies of this startup policy.

The three submitted app changes themselves behave correctly in my clean-wheel checks: Network, Protect, and Access all read the cwd file, explicit environment variables retain precedence, and the Network wheel connected to the live controller using only the cwd `.env`. Local lint/generated checks and all 14 CI jobs are green.

sirkirby:
I pushed `8c6c67b5` to resolve the two review blockers:

- added non-interactive cwd `.env` regression coverage for Network, Protect, Access, and Relay, including environment precedence and the no-file case
- applied the same cwd `.env` loading behavior to the Relay entry point

I verified the Relay test failed before the fix, all 13 focused regression cases pass afterward, `make pre-commit` is green, and clean installed wheels work from a project directory (including the earlier live-controller Network check using only the cwd `.env`). CI is rerunning on the updated head; I’ll do a fresh re-review before merge.

sirkirby (approved):
blocker resolved

PR #549, opened 2026-08-21
fix(access,api): deliver the system-log event fixes to every surface

### LINKED ISSUE #548: Access events are broken end to end: listener never started, wrong websocket keys, unread system-log shape, 5 of 7 topics undocumented, histogram fails at its own default
## Summary

The UniFi Access events surface is broken end to end. Investigating why the recent-events buffer was always empty turned up a chain of defects that each independently prevent an event path from working, across `unifi-core`, `apps/access` and `apps/api`.

Verified against a live Access controller (UniFi OS console, Access application). Happy to split this into separate issues/PRs if you'd prefer smaller units.

## 1. The websocket listener is never started

`apps/access/src/unifi_access_mcp/main.py` carries a `TODO` and logs "Access event websocket listener not yet implemented", at the point where the Protect server calls `start_listening()`. `EventManager.start_listening` exists and is unit-tested, but nothing in the application ever calls it — so the ring buffer can never fill, and `access_recent_events` / the SSE stream always return nothing.

## 2. The websocket handlers are keyed on names the client never dispatches

`EventManager.start_listening` registers handlers keyed `door_open`, `door_close`, `access_granted`, `access_denied`, `door_alarm`. The client dispatches on `WebsocketMessage.event`, whose values are the controller's dotted names — `access.logs.add`, `access.hw.door_bell`, `access.data.device.remote_unlock`, `access.base.info`, and so on. No key ever matches, so every message falls through to the client's "Unhandled websocket message type" branch. The client supports a `"*"` wildcard, which is what this needs.

Consequence: even after fixing (1), the listener connects and buffers nothing.

## 3. The websocket callback reads fields the message does not have

`WebsocketMessage` is a frozen model with `event`, `event_object_id` and `door_id`. `_on_event`'s object branch reads `id`, `type`, `user_id` and `timestamp`, so a buffered row becomes:

```python
{"id": None, "type": "unknown", "door_id": "", "user_id": None, "timestamp": None}
```

`EventBuffer.get_recent(event_type=...)` filters on `type`, so nothing is ever retrievable, and every row is timeless.

## 4. `list_events` only documents two of the seven topics the controller accepts

The `topic` parameter documents `admin` and `admin_activity`. The controller accepts seven, mirroring the Access UI's syslog Category filter:

`unlocks`, `access_denial`, `ring`, `updates`, `critical`, `admin`, `admin_activity`

The door history everyone actually wants — `access.door.unlock` plus door open/close — lives under `unlocks`, and is unreachable as documented. Guessing at plausible names (`door_openings`, `doors`, `activity`) returns `CODE_PARAMS_INVALID ... no such topic`, which names nothing valid, so there is no way to discover the vocabulary from the error.

## 5. `get_event` searches only the admin pair

`get_event` loops over `("admin", "admin_activity")`, so any event id belonging to the other five topics raises not-found.

## 6. The event serializer does not understand what `insights/system_log/search` returns

`list_events` returns raw rows from `insights/system_log/search`. Those rows use `event_type` (not `type`), `published` in epoch milliseconds (not `timestamp`), and nest the actor, door and credential inside a `metadata` object rather than as `*_id` keys. `event_from_controller` reads only the latter set, so every field resolves to `None` and `exclude_none=True` drops it. A caller receives:

```json
{"id": "", "result": "SUCCESS"}
```

No time, no door, no actor, no event type. The rows also carry a human-readable `message` (e.g. `"Access Granted (Face)"`, `"Door status - Opened"`) which is the single most useful field and is discarded entirely — there is no `message` field on the model.

## 7. `metadata` entries are typed, and reading them ungated mis-attributes events

Each metadata entry looks like `{"id": ..., "type": ..., "display_name": ...}`. `actor` is whatever *caused* the event — for an `access.dps.status.update` record that is the door hub, not a person. Reading `actor.id` without checking `type` reports a device as the user who opened the door. Likewise a reader entry carries its own device UUID, which is not the id of any door; feeding it back as `door_id` matches nothing.

## 8. The activity histogram never worked at its own default

`get_activity_summary` hard-codes `interval=3600`. The default `days=7` therefore requests 168 buckets. The endpoint fails with `CODE_SYSTEM_ERROR` above roughly 100 buckets (measured: 100 succeeds, 104+ returns a 500), so the tool fails on its default arguments. The interval needs to be derived from the requested window.

## 9. `apps/api` never normalises the rows at all

Only the MCP tool routes rows through `event_from_controller`. The GraphQL type (`Event.from_manager_output`) and the REST resource route project the raw controller row directly, reading `type` / `timestamp` / `door_id` / `user_id`. So defect (6) is live on GraphQL and REST independently of the model, and fixing the model alone does not reach either surface.

## 10. Both sort keys break cursor pagination on system-log rows

The GraphQL resolver and the REST route both key events on `(raw["timestamp"], raw["id"])`. System-log rows have neither — the time is `published` and the id is the empty string — so every row keys to `(0, "")`. `paginate()` windows with a strict `(ts, id) < (last_ts, last_id)`, and nothing can be strictly less than the key every row shares, so the second page of an events query comes back empty.

On the REST side the bare `0` default also sits next to a string timestamp in the same tuple, which raises `TypeError: '<' not supported between 'str' and 'int'` inside `sorted()` as soon as both shapes appear.

## Environment

- UniFi Access application on a UniFi OS console
- Reproduced against `main`

I have fixes for all of the above with unit tests, verified against a live controller, and will open a PR shortly.

rmaher001:
## What changed

This delivers the Access system-log fixes from #559 through every downstream surface:

- starts the Access websocket listener during MCP server startup
- documents all seven supported system-log topics and the controller's dotted websocket event names
- normalizes Access system-log rows consistently across MCP, REST, GraphQL, and SSE
- exposes the controller-rendered `message` field through REST, GraphQL, and recent-event serialization
- raises the Access and API `unifi-core` floors to `0.4.33`, the published release containing the required normalization and websocket behavior

## Why

The Core fix is now merged and published as `core/v0.4.33`. Without this downstream delivery, the listener remains dead during Access MCP startup and API clients still receive incomplete system-log rows.

Credential projection remains intentionally narrow because the live controller has not established a safe canonical mapping between `authentication`, `nfc_id`, and credential UUIDs. API topic selection remains separate follow-up #560 because it changes the public API contract.

Closes #548.

## Validation

- rebased cleanly onto current `main`
- `make pre-commit` passed
- focused Access startup tests: 4 passed
- focused Access GraphQL fixtures: 23 passed
- Access/API event projection and resource tests: 99 passed
- full pin-alignment checks passed, including 271 API action bindings against published Core 0.4.33
- maintainer live smoke passed for Access safe lifecycle, API actions, API resource parity, and Network/Protect/Access SSE streams
- targeted live startup established a connected Access websocket using the wildcard handler
- targeted live REST/GraphQL projection returned the same 10 IDs with `id`, `type`, `timestamp`, and `message` populated on all 10 rows

<details>
<summary>Maintainer live validation excerpts</summary>

```text
access readonly ok: access_list_events (77ms)
access readonly ok: access_get_event (6415ms)
access lifecycle:create ok: access_create_visitor (101ms)
access lifecycle:delete ok: access_delete_visitor (105ms)
access lifecycle:verify-cleanup-status ok: access_delete_visitor - Developer API DELETE revokes access and retains the visitor as cancelled history

api-actions pass: access_list_doors http=200 success=True live_count=1
api-actions pass: access_list_users http=200 success=True live_count=5
api-actions pass: access_list_visitors http=200 success=True live_count=16

api-resources pass: /v1/sites/default/doors http=200 items=1 parity=ok
api-resources pass: /v1/sites/default/users http=200 items=5 parity=ok
api-resources pass: /v1/sites/default/visitors http=200 items=10 parity=ok

api-streams pass: /v1/streams/network/events events=0 keepalives=1
api-streams pass: /v1/streams/protect/events events=0 keepalives=1
api-streams pass: /v1/streams/access/events events=0 keepalives=1

{'rest_count': 10, 'graphql_count': 10, 'matching_ids': 10, 'populated_rest_fields': {'id': 10, 'type': 10, 'timestamp': 10, 'message': 10}}

{'listener_called': True, 'handlers': ['*'], 'socket_type': 'UnifiAccessWebsocket', 'is_running': True, 'is_connected': True}
```

</details>

### DIFF
--- a/apps/access/pyproject.toml
+++ b/apps/access/pyproject.toml
@@ -8,9 +8,9 @@ dependencies = [
     "unifi-mcp-shared>=0.6.7,<0.7",
     # Access managers in unifi-core import unifi_access_api (PyPI name
     # py-unifi-access); pull it in via the [access] extra rather than
-    # declaring py-unifi-access here directly. Access event identity requires
-    # Core 0.4.31 for separator-insensitive Access device MAC resolution.
-    "unifi-core[access]>=0.4.31,<0.5",
+    # declaring py-unifi-access here directly. Access system-log event
+    # normalization and websocket topics require Core 0.4.33.
+    "unifi-core[access]>=0.4.33,<0.5",
     "mcp[cli]>=1.28.1,<2",
     # Published wheels do not consume uv.lock; keep MCP transitive security
     # floors explicit so upgrades cannot retain vulnerable installed versions.
--- a/apps/access/src/unifi_access_mcp/main.py
+++ b/apps/access/src/unifi_access_mcp/main.py
@@ -69,8 +69,12 @@ async def main_async():
         if ws_enabled:
             try:
                 event_manager.set_server(server)
-                # TODO: Implement websocket event listening for Access
-                logger.info("Access event websocket listener not yet implemented.")
+                # start_listening needs an API client: the Access websocket is
+                # key-authenticated, and a proxy-only session cannot open it.
+                # It logs that condition itself and returns without raising, so
+                # a proxy-only deployment keeps working with REST queries and
+                # an empty recent-events buffer.
+                await event_manager.start_listening()
             except Exception as ws_exc:
                 logger.error(
                     "Failed to start event websocket listener: %s. "
--- a/apps/access/src/unifi_access_mcp/tools/events.py
+++ b/apps/access/src/unifi_access_mcp/tools/events.py
@@ -38,7 +38,12 @@ async def access_list_events(
     topic: Annotated[
         str,
         Field(
-            description="Event topic to query: 'admin' (user/account changes) or 'admin_activity' (admin actions). Default: admin."
+            description=(
+                "Event topic, mirroring the Access UI's syslog Category filter: 'unlocks' "
+                "(door grants and open/close), 'access_denial' (refused attempts), 'ring' "
+                "(doorbell), 'updates', 'critical', 'admin' (user/account changes), "
+                "'admin_activity' (admin actions). No other value is accepted. Default: admin."
+            )
         ),
     ] = "admin",
     start: Annotated[
@@ -122,7 +127,14 @@ async def access_get_event(
 async def access_recent_events(
     event_type: Annotated[
         Optional[str],
-        Field(description=("Filter by event type: door_open, door_close, access_granted, access_denied, door_alarm.")),
+        Field(
+            description=(
+                "Filter by event type, matched exactly against the controller's own dotted "
+                "event names as delivered over the websocket - e.g. 'access.door.unlock', "
+                "'access.dps.status.update', 'access.logs.add', 'access.hw.door_bell', "
+                "'access.data.device.remote_unlock'. Omit to return every buffered event."
+            )
+        ),
     ] = None,
     door_id: Annotated[
         Optional[str],
--- a/apps/access/src/unifi_access_mcp/tools_manifest.json
+++ b/apps/access/src/unifi_access_mcp/tools_manifest.json
@@ -1941,7 +1941,7 @@
               "type": "string"
             },
             "topic": {
-              "description": "Event topic to query: 'admin' (user/account changes) or 'admin_activity' (admin actions). Default: admin.",
+              "description": "Event topic, mirroring the Access UI's syslog Category filter: 'unlocks' (door grants and open/close), 'access_denial' (refused attempts), 'ring' (doorbell), 'updates', 'critical', 'admin' (user/account changes), 'admin_activity' (admin actions). No other value is accepted. Default: admin.",
               "type": "string"
             },
             "user_id": {
@@ -2639,7 +2639,7 @@
               "type": "string"
             },
             "event_type": {
-              "description": "Filter by event type: door_open, door_close, access_granted, access_denied, door_alarm.",
+              "description": "Filter by event type, matched exactly against the controller's own dotted event names as delivered over the websocket - e.g. 'access.door.unlock', 'access.dps.status.update', 'access.logs.add', 'access.hw.door_bell', 'access.data.device.remote_unlock'. Omit to return every buffered event.",
               "type": "string"
             },
             "limit": {
--- a/apps/access/tests/unit/test_websocket_wiring.py
+++ b/apps/access/tests/unit/test_websocket_wiring.py
@@ -1,0 +1,71 @@
+"""Guard the call site: `main_async` must start the event listener.
+
+`EventManager.start_listening` existed, was unit-tested, and was never called
+by the application — `main_async` carried a TODO and a log line claiming the
+feature was unimplemented, where the Protect server calls it. The manager tests
+passed because they invoked the manager directly, so nothing caught that the
+buffer could not fill in production.
+
+These tests drive `main_async` and assert on what it actually awaits. An
+earlier version asserted that the call appeared in `inspect.getsource`, which
+would have passed with the call commented out — the very state it was written
+to catch.
+"""
+
+import asyncio
+from contextlib import contextmanager
+from unittest.mock import AsyncMock, patch
+
+from unifi_access_mcp import main as access_main
+from unifi_access_mcp.runtime import config
+
+
+@contextmanager
+def _startup(*, connected: bool, websocket_enabled: bool, listener: AsyncMock | None = None):
+    """Run `main_async` with its side effects stubbed.
+
+    Only the connection, the listener and the two long-running tails are
+    replaced. The websocket decision itself runs for real, since that branch is
+    what these tests are about.
+    """
+    events = dict(getattr(config.access, "events", {}) or {})
+    events["websocket_enabled"] = websocket_enabled
+    with (
+        patch.object(config.access, "events", events),
+        patch("unifi_mcp_shared.bootstrap.assert_credentials_configured"),
+        patch.object(access_main.connection_manager, "initialize", AsyncMock(return_value=connected)),
+        patch.object(access_main.event_manager, "set_server"),
+        patch.object(access_main.event_manager, "start_listening", listener or AsyncMock()) as started,
+        patch("unifi_mcp_shared.tool_registration.register_tools_for_mode", AsyncMock()),
+        patch("unifi_mcp_shared.transport.resolve_http_config", return_value=(False, "http", "0.0.0.0", 3002)),
+        patch("unifi_mcp_shared.transport.run_transports", AsyncMock()) as transports,
+    ):
+        asyncio.run(access_main.main_async())
+        yield started, transports
+
+
+def test_the_listener_is_started_on_a_successful_websocket_enabled_startup() -> None:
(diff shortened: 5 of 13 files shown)


rmaher001:
Pushed `2872639` after a self-review turned up two ways this depended on unreleased `unifi-core`, plus some defects of my own.

**Floor safety.** `apps/api` pins `unifi-core>=0.4.29,<0.5`. Checked against the published 0.4.29 wheel: `event_sort_key` doesn't exist there, and the `Event` model has no `message` field — so the first raised `ImportError` at import time and the second raised `AttributeError`, i.e. a 500 on every events request. `event_sort_key` now lives in `unifi_api.services.access_event_key` (the API layer is its only consumer), and `message` is read from the row rather than the model. The only remaining new core import is `event_from_controller`, which 0.4.29 does have.

**One thing to flag for sequencing:** `event_from_controller`'s *new* system-log handling naturally only reaches a wheel installed at the 0.4.29 floor once a core release carries it and the floor is raised. Nothing errors in the meantime — those rows just project as they do today. Entirely your call how to sequence that.

Also fixed in this push: the buffer was carrying the whole websocket `model_dump()`, including the `data` sub-payload with actor display names and credential ids, which `access_recent_events` and `GET /access/recent-events` return unprojected — now buffers a flat projection only. High-rate location/remote-view frames delivered by the `"*"` subscription are no longer buffered, since against a 100-entry ring a burst could evict a door unlock seconds after it happened. I also reverted my own `get_event` topic fan-out: system-log rows carry `id: ""`, so a caller-supplied id can't match one however many topics are searched — widening it only made the same 404 more expensive. That underlying limitation is filed separately as #550.

Plus: unwrap `.raw` before normalising; stop the REST sort key falling back to `{}` (which keyed every unknown shape identically); count histogram buckets with ceil, since flooring let `days=673` pass a guard advertising 672 and then request 97 against a cap of 96; and clamp a non-positive `days`.

Re-verified against a live controller after deploying this exact content.

sirkirby (changes requested):
## Hard blockers

- [ ] **Make the cursor identity collision-resistant.** `apps/api/src/unifi_api/services/access_event_key.py` hashes only selected metadata. On the live controller I found two distinct admin rows that differed at `metadata.user.id` but produced the same key. Traversing the 68-row snapshot at `limit=2` returned 67 rows, and the branch-local REST endpoint reproduced the same loss (67 rows across 34 pages). Hash a canonical stable representation of the complete row, or otherwise include every distinguishing field, and add a regression where only `metadata.user.id` differs plus a full cursor traversal assertion.

- [ ] **Sequence and raise the Core dependency floors.** I built the Access and API wheels from this head and installed each with its declared minimum Core. Access + Core 0.4.22 still registered only the five obsolete websocket handler names and projected a system-log row to `{"id": "", "result": "ACCESS"}`. API + Core 0.4.29 still projected `type`, `timestamp`, `door_id`, `user_id`, and `credential_id` as `None`. Please land/release the Core prerequisite first and raise both downstream floors (or use an equivalent split/sequence), with a semantic clean-install floor test.

- [ ] **Finish or narrow the credential projection.** The normalizer reads only `metadata.credential`, but the live `unlocks`/`access_denial` rows used `authentication` and sometimes `nfc_id` (with distinct IDs), never `credential`; `credential_id` remained empty across the sample. Verify which typed metadata ID is the canonical credential, map it safely and test it, or remove the claim from this PR/model scope.

- [ ] **Replace the source-string wiring guard with a behavioral test.** `test_main_starts_the_event_listener` passes if the call is commented out or unreachable. Patch `event_manager.start_listening`/transport setup, invoke `main_async`, and assert the listener is awaited only for a successful, websocket-enabled startup.

## PR body alignment

The body still says all ten defects are fixed and that `get_event` fans out to every topic, but the current head intentionally reverted that and moved the identity work to #550. Please update the body/closing scope so merge does not close #548 with one listed defect still described as fixed.

## What I verified

I ran the full `make pre-commit` gate successfully, branch-local Access safe smoke, all seven topic values, the 7-day histogram, an actual websocket connection, and live REST/GraphQL projections. The overall direction is solid; the blockers above are specific release/runtime gaps that green workspace CI does not catch.

rmaher001:
All four blockers addressed. Every one reproduced first, so the fix is pinned by a test that fails without it.

**Cursor identity collision.** Confirmed and fixed. I reproduced your case as a unit test — two admin rows identical except `metadata.user.id` — and as a page-at-a-time traversal of 68 rows at `limit=2`; both failed on the old key. The subset digest is gone: identity now comes from `event_identity`, so the whole canonical row distinguishes a row, and there is nothing left for me to have guessed exhaustively. A second regression covers a field the module has never heard of. This work is now #558, which stands alone and needs only the published 0.4.30 floor — I verified against the real wheel that `event_identity` is importable there, so a fresh install is safe.

**Core dependency sequencing.** Agreed, and split the way you did #555/#556. #559 is the Core prerequisite alone; this PR is the downstream delivery and is now a draft carrying no floor bump, because the bump can only be written once the version exists. I stopped asserting the gap and reproduced it — installing `unifi-core[access]==0.4.30` in isolation, a system-log row still projects to `{'id': …, 'result': 'ACCESS'}` and `Event.message` does not exist. That output is in the PR body.

**Credential projection.** Withdrawn rather than finished. You are right that the live rows use `authentication` and `nfc_id`; on my controller `credential_id` was empty on every sampled row, so the mapping never fired at all and I had shipped a claim I hadn't observed. I can't establish from the data I have which of those is the canonical credential identity, and mapping either on a guess would put a reader or auth-method id into a field callers read as a credential UUID. So `credential_id` now reads only the top-level field, with a test that fails if someone re-adds the metadata guess, and the docstring says why.

**Listener wiring test.** You're right that it was vacuous — it asserted the call appeared in `inspect.getsource`, which passes with the call commented out. Replaced with tests that patch `start_listening` and the transports, run `main_async`, and assert what is actually awaited: awaited once on a successful websocket-enabled startup; not awaited when the connection fails or the websocket is disabled; and startup still reaches its transports when the listener raises. I checked they fail for the right reason by commenting the call out — two of the four go red.

**Body and closing scope.** Rewritten. The empty-identity defect is credited to #550/#555/#556, not claimed here; my `get_event` fan-out and its no-fan-out test are dropped, since your paged lookup supersedes both. #548 is closed only by this delivery PR, not by #558 or #559.

One thing I found while doing this and did not fold in: neither API surface passes a topic to `list_events`, so REST and GraphQL can only ever return `admin` events and door history is unreachable there. Filed as #560 rather than widened into this PR, since it changes a public contract.

On your verification note — thank you for building the wheels at their declared minimums. That's the check I should have been running myself, and it's the one that caught the two real defects here.

sirkirby (approved):
I rebased this onto current main, raised the downstream Core floors to the published 0.4.33 prerequisite, removed the temporary raw-message fallback, and added the missing SSE message assertion. The full pre-commit and pin-alignment gates pass. Maintainer live validation also passes for Access safe lifecycle, API actions/resources/streams, REST/GraphQL event projection, and real websocket startup on this exact head.

PR #559, opened 2026-08-23
fix(access): read the system-log shape and the websocket the controller sends

rmaher001:
`list_events` queries `insights/system_log/search`, whose rows use `event_type`, `published` in epoch milliseconds, and nest the actor and door inside `metadata`. The serializer read `type`, `timestamp`, `door_id` and `user_id`, so every one of them resolved to `None` and `exclude_none=True` dropped it — a caller got an id and a result and nothing else.

This is the Core half only, so it can be released independently before anything downstream depends on it.

## What changed

- **Read the system-log shape** in `event_from_controller`, and expose the controller's rendered `message`. The websocket/legacy shape still reads exactly as before.
- **Name the seven topics the controller accepts.** The docstring advertised `admin` and `admin_activity`; the vocabulary is the Access UI's syslog Category list, and door history lives under `unlocks` — guessing at `door_openings` or `doors` finds nothing, which is what made that history look unreachable. An unsupported topic is now refused up front with the valid list instead of by the controller's own `no such topic`.
- **Subscribe to the websocket keys the controller actually sends.** Handlers were registered under five friendly names (`door_open`, `access_granted`, …) that never appear in a frame; the client dispatches on `WebsocketMessage.event`, whose values are dotted names like `access.logs.add`. Every message hit the client's "unhandled" branch, so the ring buffer behind `access_recent_events` could not fill.
- **Buffer a flat projection rather than the whole frame.** `model_dump()` carries the `data` sub-payload — actor display names, credential and device identifiers — into a tool that returns buffered rows unprojected. High-rate location and remote-view frames are also no longer buffered: against a 100-entry ring, a burst could evict a door unlock seconds after it happened.
- **Derive the histogram interval from the window.** A hard-coded `interval=3600` asked for 168 buckets at the tool's own 7-day default and the controller fails above ~100, so the tool never worked at its default. Bucket counting uses ceil, since flooring let `days=673` pass a guard advertising 672 and then request 97 against a cap of 96; a non-positive `days` is clamped before the window is built.

`credential_id` is deliberately left reading only a top-level field. The live `unlocks`/`access_denial` rows carry credential context under `authentication` and `nfc_id`, never `credential`, and which of those is the canonical credential identity is not established — so the serializer makes no claim rather than a plausible-looking wrong one.

## Validation

- `unifi-core`: 1,660 passed. `apps/access`: 215 passed. `apps/api`: 1,246 passed. `ruff format --check` and `ruff check` clean at the `uv.lock` pin (0.16.2); `make check-generated` exit 0; `check_pin_alignment.py` exit 0.
- Exercised against a live UniFi Access controller from a deployed build at `06f412b` (this branch layered with #558 and the downstream delivery):
  - system-log rows project a full shape — id, type, ISO timestamp with sub-second precision, door, actor, message, result — where they previously returned only an empty id and a result;
  - all seven topics accepted; `door_openings` refused before any controller request;
  - `get_event` round-trips a listed id, including a four-month-old event surfaced from the `access_denial` topic;
  - `get_activity_summary` returns 84 buckets at its own 7-day default (interval 7200); `days=673` and `days=1000` are refused with the accurate 672-day limit;
  - the listener starts and the websocket connects on boot (log: `[event-mgr] Websocket subscription started.` / `Websocket connection established`). **Not covered live:** the ring buffer filling from a real door event — no door activity occurred in the test window, and I don't actuate hardware for QA. That path is covered by unit tests written against the real `WebsocketMessage` contract.

Relates to #548. Does not close it — the downstream delivery that raises the Core floors follows once this is released.

### DIFF
--- a/packages/unifi-core/src/unifi_core/access/managers/event_manager.py
+++ b/packages/unifi-core/src/unifi_core/access/managers/event_manager.py
@@ -19,6 +19,7 @@ import logging
 import time
 from collections import deque
 from collections.abc import Callable
+from datetime import datetime, timezone
 from typing import Any
 
 from unifi_core.access.models.events import event_identity, with_event_identity
@@ -27,15 +28,6 @@ from unifi_core.exceptions import UniFiConnectionError, UniFiNotFoundError
 logger = logging.getLogger(__name__)
 
 _GET_EVENT_PAGE_SIZE = 100
-_EVENT_SEARCH_TOPICS = (
-    "unlocks",
-    "access_denial",
-    "ring",
-    "updates",
-    "critical",
-    "admin",
-    "admin_activity",
-)
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


sirkirby (approved):
I rebased this onto the current main branch and reviewed the resulting Core-only diff. I verified the installed py-unifi-access websocket contract, the Access layering and privacy boundary, the system-log identity and field mappings, the histogram bucket guard, and the new unit coverage. Locally, the focused event suite (71 tests), all unifi-core tests (1,750), the full make pre-commit gate, and cross-package pin alignment pass. Live validation against an Access controller also passed the complete read-only and safe smoke phases. A privacy-safe branch/main probe confirmed all seven supported log topics, populated type/timestamp/message and typed door/user fields, successful 0/7/672-day histograms capped at 96 buckets, and a running wildcard websocket subscription. I found no blocking issues.

Issue #500, opened 2026-08-04
fix: Network stat endpoints return 401 after ~24h of MCP server uptime (session not re-authenticated)


vanpatrickbevill:
### Preflight checklist

- [x] I searched existing issues and did not find a duplicate.
- [x] I am using the latest released version, or I confirmed the issue still exists on `main`.
- [x] I have removed credentials, tokens, API keys, IP addresses, and other private data from the report.

### Affected component

Network MCP server

### Version

unifi-network-mcp 0.20.9 (uvx @latest)

### Controller hardware / host

UCG Max / UCG Ultra / UCG Fiber

### UniFi OS version

5.1.19

### Network application version

10.5.67

### Protect application version

7.1.87

### Site setup

Single site, default config, zone-based firewall

### Install method

uvx unifi-network-mcp@latest (stdio, Claude Desktop)

### OS where the MCP client runs

macOS (Claude Desktop, MacBook Pro)

### MCP client

Claude Desktop (Cowork / Dispatch)

### AI model used

Claude Fable 5

### Exact prompt

"How is my network health today?" (a routine monitoring request; the agent batched several read-only tools)

### Tool calls observed

```shell
Failing after ~24h uptime: unifi_get_client_details, unifi_get_top_clients, unifi_list_events, unifi_get_alerts, unifi_list_devices, unifi_get_gateway_stats (intermittent), unifi_get_anomalies. Still working in the same session: unifi_get_network_health, unifi_get_traffic_flows, unifi_batch/unifi_batch_status, unifi_get_system_info, unifi_get_port_stats, unifi_create_dns_record, unifi_create_firewall_policy previews.
```

### Raw tool output (sanitized)

```json
Failed to get alerts: /stat/alarm failed (Call https://<controller-ip>/proxy/network/api/s/default/stat/alarm received 401 Unauthorized) after the v2 system-log lookup failed (Call https://<controller-ip>/proxy/network/v2/api/site/default/system-log/critical received 401 Unauthorized).

Failed to list devices: Call https://<controller-ip>/proxy/network/api/s/default/stat/device received 401 Unauthorized

Failed to get client details for <mac>: Call https://<controller-ip>/proxy/network/api/s/default/stat/sta received 401 Unauthorized
```

### Expected behavior

All tools continue to work for the life of the MCP server process, re-authenticating to the controller as needed when the login session/cookie expires.

### Actual behavior

[...]

PR #578, opened 2026-08-25
fix(core): re-authenticate on the handler-refresh path, not just request()

rmaher001:
Relates to #500 — **deliberately non-closing.** That issue is not mine and its reporter should be the one to confirm; I have commented there with the evidence.

#500's symptom is back on a current controller: `unifi_list_devices` returning `401 Unauthorized` after the session ages out. #514 fixed a real bug, but not the one #500 reported.

## The split #500 described is the whole diagnosis

The reporter listed exactly which tools failed and which kept working:

- **failing** — `list_devices`, `get_client_details`, `get_top_clients`
- **working** — `get_network_health`, `get_traffic_flows`, `get_system_info`

That maps cleanly onto two different call paths. The working ones go through `ConnectionManager.request()`, which catches `LoginRequired` and re-authenticates. The failing ones call an aiounifi handler's own `update()`:

```
device_manager.py:41   controller.devices.update()
client_manager.py:45   controller.clients.update()
client_manager.py:78   controller.clients_all.update()
stats_manager.py:328   controller.dpi_apps.update()
stats_manager.py:329   controller.dpi_groups.update()
```

`handler.update()` builds and issues its own request internally — `controller.request()` → `connectivity.request()` → `raise LoginRequired`. It never passes through `ConnectionManager.request()`, so the recovery added in #514 is not on that path at all.

## Why the fix verified clean

#514's regression is `test_expired_session_explicitly_reauthenticates_and_retries_request`. It sets `controller.request` to raise `LoginRequired` once, then drives `manager.request(...)`. That exercises the path that already recovered. Nothing in the suite drove `handler.update()`, so the five call sites above were never covered, and the reported symptom survived a green test suite.

This is the part worth guarding against: the new tests here deliberately drive `handler.update()` rather than `request()`, because that distinction is what the previous round missed.

## Observed live

On a controller running Network 10.5.67, after the gateway took a firmware upgrade and rebooted — which invalidates the session server-side:

```
unifi_list_devices -> "Call https://<controller>/proxy/network/api/s/default/stat/device
                      received 401 Unauthorized"
```

Three consecutive failures at 20ms / 6ms / 13ms, and **no `/api/auth/login` line in the log after the startup login**. A real re-authentication is a network round-trip and logs `Login required detected during request`. Neither appears — no login was attempted. Managers that do use the wrapped `request()` (`oon`, `qos`, `traffic_route`, `content_filter`, `vpn`) recovered normally against the same controller in the same window.

## The change

`ConnectionManager.refresh_handler(name)` wraps `handler.update()` in the same logic `request()` already applies:

- the same generation-locked `_reauthenticate(auth_generation)`, so concurrent refreshes share one login rather than starting a stampede;
- the same terminal treatment of a **second** `LoginRequired` — `_block_automatic_reconnect()` + `_discard_connection()` — so a refreshed-but-rejected session cannot turn every later tool call into another login attempt;
- the handler resolved **by name**, so the retry reads it off `self.controller` again. This mirrors `request()` re-deriving its request method after re-authenticating, and matters if the controller object is ever replaced rather than re-logged-in.

All five call sites converted. I also grepped `apps/protect` and `apps/access` for the same shape — `controller.<handler>.update()` — and there are no other occurrences anywhere in `packages/unifi-core/src`.

## Tests

Five, all driving `handler.update()`:

- re-authenticates and retries once on an expired session
- no spurious login when the session is valid
- re-raises the original error when the login itself fails
- a persistent `LoginRequired` after re-authentication opens the auth circuit rather than looping
- `DeviceManager.get_devices()` recovers end to end — this is `unifi_list_devices`, the tool from #500

Each was confirmed to fail without the fix: removing the re-authentication fails 3 of 5, and removing the circuit guard fails the fourth. Against the code currently on `main` all of them fail.

Six existing test files needed their `mock_connection` double taught the new seam. Those doubles are `MagicMock`, so an awaited call on the connection returns a non-awaitable; each now delegates to the very same `controller.<name>.update()` mock the test already set up, so no existing assertion loses its meaning.

## Verification

`unifi-core` 1,624 passed, `apps/network` 1,007 passed, `apps/api` 1,331 passed. `ruff format --check` and `ruff check` clean at the `uv.lock` pin (0.16.2). `generate_skill_references.py --check` and `generate_api_action_catalog.py --check` both report no drift; `check_pin_alignment.py` exits 0 against the published Core 0.4.30 floor.

Exercised on a live controller from a build of this branch: `unifi_list_devices`, the client lookups and the DPI stats paths all return normally through `refresh_handler`, and an absent device still reports not-found rather than an empty success.

One limit I would rather state than gloss: the re-authentication branch itself is not triggerable on demand against a live controller without invalidating its session, so live QA covers the converted call sites' behavior, and the recovery branch rests on the tests above.

### DIFF
--- a/apps/network/tests/unit/test_client_details.py
+++ b/apps/network/tests/unit/test_client_details.py
@@ -19,6 +19,14 @@ def mock_connection():
     conn._update_cache = MagicMock()
     conn._invalidate_cache = MagicMock()
     conn.ensure_connected = AsyncMock(return_value=True)
+
+    # ConnectionManager.refresh_handler() is how the managers refresh a
+    # handler collection now; delegate to the same ``update()`` these
+    # tests already stub so their setup keeps its meaning.
+    async def _refresh_handler(name, _c=conn):
+        return await getattr(_c.controller, name).update()
+
+    conn.refresh_handler = _refresh_handler
     conn.controller = MagicMock()
     conn.controller.clients = MagicMock()
     conn.controller.clients.update = AsyncMock()
--- a/apps/network/tests/unit/test_client_ip_settings.py
+++ b/apps/network/tests/unit/test_client_ip_settings.py
@@ -28,6 +28,14 @@ class TestClientIPSettings:
         conn.controller.clients_all.update = AsyncMock()
         conn.controller.clients_all.values = MagicMock(return_value=[])
         conn.ensure_connected = AsyncMock(return_value=True)
+
+        # ConnectionManager.refresh_handler() is how the managers refresh a
+        # handler collection now; delegate to the same ``update()`` these
+        # tests already stub so their setup keeps its meaning.
+        async def _refresh_handler(name, _c=conn):
+            return await getattr(_c.controller, name).update()
+
+        conn.refresh_handler = _refresh_handler
         return conn
 
     @pytest.fixture
--- a/apps/network/tests/unit/test_client_lookup.py
+++ b/apps/network/tests/unit/test_client_lookup.py
@@ -28,6 +28,14 @@ class TestGetClientByIP:
         conn.controller.clients_all.update = AsyncMock()
         conn.controller.clients_all.values = MagicMock(return_value=[])
         conn.ensure_connected = AsyncMock(return_value=True)
+
+        # ConnectionManager.refresh_handler() is how the managers refresh a
+        # handler collection now; delegate to the same ``update()`` these
+        # tests already stub so their setup keeps its meaning.
+        async def _refresh_handler(name, _c=conn):
+            return await getattr(_c.controller, name).update()
+
+        conn.refresh_handler = _refresh_handler
         return conn
 
     @pytest.fixture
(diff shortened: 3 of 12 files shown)


schford:
Independent confirmation of the symptom this PR addresses, on `unifi-network-mcp` 0.27.1 against a UCG Fiber (UniFi Network 10.5.67).

After a controller reboot, tools backed by `stat/device` (`unifi_list_devices`, `unifi_get_device_details`) return 401 Unauthorized, while `unifi_get_system_info` and `unifi_get_network_health` keep working — exactly the split described in #500. Hit it twice today alone (an overnight controller restart, then a deliberate reboot during electrical work).

Consistent with this PR's diagnosis: issuing one request()-path call (e.g. `unifi_get_system_info`) re-authenticates the shared session, after which the previously-failing tools immediately recover — which is what you'd expect if the handler `.update()` paths bypass the re-auth wrapper.

sirkirby (approved):
Rebased onto current main and verified at 66c55efc. The named handler seam is the right boundary: it preserves generation-locked re-authentication, re-resolves the handler for the retry, and opens the auth circuit after a second LoginRequired. I also updated the #541 MAC regression fixture for the new async seam. Full pre-commit passes, the five affected live read paths pass, and all 14 fresh checks are green.

Issue #564, opened 2026-08-24
limit=0 is treated as "no limit" across all three event managers, and as "30" on Access list_events


rmaher001:
`limit` is tested for truthiness rather than for `None`, so `limit=0` silently means "no limit" — and on Access `list_events` it means "30".

Found while pre-emptively auditing the event paths for the input-validation class Copilot flagged on #558 (non-finite floats, value-type bypasses). This one is the same shape: a parameter whose zero value is meaningful, guarded by a truthiness test.

## Reproduced live

Against a live controller (Access, Network 10.5.67), `access_list_events` with an explicit `limit: 0`:

```
access_list_events {"topic": "unlocks", "limit": 0}
  -> {"success": true, "data": {..., "count": 30}}
```

Thirty events for a request that asked for none. The documented contract is *"Maximum number of events to return (default 30)"*, so `0` is a caller stating a maximum of zero, not a caller omitting the argument.

## Root cause

Four sites, one mistake — `0` and `None` are not distinguished:

- `packages/unifi-core/src/unifi_core/access/managers/event_manager.py:388` — `page_size = limit or 30`
- `packages/unifi-core/src/unifi_core/access/managers/event_manager.py:81` — `if limit and len(results) >= limit:`
- `packages/unifi-core/src/unifi_core/protect/managers/event_manager.py:138` — `if limit and len(results) >= limit:`
- `packages/unifi-core/src/unifi_core/network/managers/event_manager.py:58` — `if limit and len(out) >= limit:`

The three buffer-truncation sites never break out of the loop when `limit=0`, so the caller gets the **entire** buffer. The Access `page_size` site substitutes the default instead.

The managers type the parameter `limit: int | None = None`, where `None` genuinely means "no limit" — so the intent is there, the test just doesn't implement it. The Access tool signature types it `limit: int = 30`, non-optional, which makes an explicit `0` the only way a caller expresses "none" and the one value that misbehaves.

I verified the Access sites live; the Protect and Network sites are identified by inspection at the file:line above and are the same expression.

## Also worth a look while in here

A negative `limit` is passed through to the controller unvalidated:

```
access_list_events {"topic": "unlocks", "limit": -1}
  -> {"success": false, "error": "Failed to list events: Proxy request failed:
      API code -2 CODE_PARAMS_INVALID
      POST https://<controller-host>/proxy/access/api/v2/insights/system_log/search?page_size=-1&page_num=1&isAccess
      — Invalid parameters."}
```

[...]

PR #614, opened 2026-09-03
fix(core): guard event limit on identity, not truthiness

AmirF194:
Refs #564. This covers the truthiness bug at all four sites the issue names; the issue's separate "also worth a look" note about negative limits is not part of this diff, so I'm not closing the issue with this PR.

## Root cause

Four sites test `limit` for truthiness instead of identity against `None`, so an explicit `limit=0` (a caller stating "return zero events") is read the same as "no limit given":

- `packages/unifi-core/src/unifi_core/access/managers/event_manager.py:81` and `:388` (`EventBuffer.get_recent` and `list_events`'s `page_size = limit or 30`)
- `packages/unifi-core/src/unifi_core/protect/managers/event_manager.py:138`
- `packages/unifi-core/src/unifi_core/network/managers/event_manager.py:59`

The three `EventBuffer.get_recent` sites share one more bug once the truthiness check is fixed: the `limit` guard ran *after* `results.append(event)`, so the very first eligible event was always appended before a `limit=0` guard could ever run. I moved the check to the top of the loop, before an event is filtered or appended, so it can actually stop at zero.

## Fix

Guard on `limit is not None` and check it before processing each event, at all three `EventBuffer.get_recent` sites. For Access's REST `list_events`, `limit` is already a non-optional `int` defaulting to 30. The zero case now returns `[]` locally before any controller request; positive limits are passed through as `page_size = limit`. This also brings the file in line with `network/managers/event_manager.py`'s own `_get_events_v2`, which already treats `limit <= 0` as "return `[]`" on its REST path.

Left the negative-limit case out of scope, as the reporter did: the REST path already fails safe there (the controller rejects a negative `page_size`), and the buffer path's existing behavior for a negative limit is unrelated to this bug.

## Verification

- Added regression tests at all four sites and ran them on Python 3.13: they fail on `main` and pass on this branch. Files: `apps/access/tests/unit/test_events.py` (buffer + REST `page_size`), `apps/protect/tests/unit/test_event_buffer.py`, `packages/unifi-core/tests/network/managers/test_event_manager_buffer.py`.
- Ran the full `access`, `protect`, and `network` app unit suites plus the full `unifi-core` package suite; all pass (2726 tests total).
- `ruff format --check` and `ruff check` are clean on the whole workspace.
- I haven't confirmed the `unifi-core` package suite runs in this PR's checks: `release-core.yml` is the only workflow that runs it, and it's tag-triggered, not PR-triggered, so I ran it by hand instead.

## Maintainer verification

I independently validated exact head `085a26a9` against the reference controllers.

<details>
<summary>Live validation result</summary>

```text
access_list_events(limit=0)
{"success": true, "count": 0, "events_len": 0, "has_error": false}

Full safe smoke: Network passed; Protect passed; Access passed.
Focused regressions: Access 22 passed; Protect 19 passed; Network buffer 7 passed.
Core suite: 1787 passed.
GitHub checks: 14 passed.
```

</details>

### DIFF
--- a/apps/access/tests/unit/test_events.py
+++ b/apps/access/tests/unit/test_events.py
@@ -71,6 +71,13 @@ class TestEventBufferExtended:
         assert events[0]["seq"] == 4
         assert events[-1]["seq"] == 2
 
+    def test_buffer_limit_zero_returns_empty(self):
+        """An explicit limit=0 means zero results, not unlimited."""
+        buf = EventBuffer(max_size=10, ttl_seconds=300)
+        for i in range(5):
+            buf.add({"type": "test", "seq": i})
+        assert buf.get_recent(limit=0) == []
+
     def test_buffer_ttl_expiration(self):
         """Buffer skips events older than TTL."""
         buf = EventBuffer(max_size=10, ttl_seconds=1)
@@ -203,6 +210,16 @@ class TestEventManagerREST:
         assert body["start"] == "2026-03-01"
         assert body["end"] == "2026-03-17"
         assert "page_size=10" in call_args[0][1]
+
+    @pytest.mark.asyncio
+    async def test_list_events_limit_zero_requests_zero_page_size(self, event_mgr_proxy, cm_proxy):
+        """An explicit limit=0 asks the controller for zero events, not the 30 default."""
+        with patch.object(cm_proxy, "proxy_request", new_callable=AsyncMock) as mock_req:
+            mock_req.return_value = {"data": {"events": []}}
+            await event_mgr_proxy.list_events(limit=0)
+
+        call_args = mock_req.call_args
+        assert "page_size=0" in call_args[0][1]
 
     @pytest.mark.asyncio
     async def test_list_events_no_proxy(self, event_mgr_none):
--- a/apps/access/tests/unit/test_events.py
+++ b/apps/access/tests/unit/test_events.py
@@ -212,14 +212,14 @@ class TestEventManagerREST:
         assert "page_size=10" in call_args[0][1]
 
     @pytest.mark.asyncio
-    async def test_list_events_limit_zero_requests_zero_page_size(self, event_mgr_proxy, cm_proxy):
-        """An explicit limit=0 asks the controller for zero events, not the 30 default."""
+    async def test_list_events_limit_zero_returns_empty_without_calling_controller(self, event_mgr_proxy, cm_proxy):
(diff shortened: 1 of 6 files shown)


sirkirby (changes requested):

- [ ] Guarantee `limit=0` locally in the Access manager instead of delegating it as `page_size=0`.

I independently reproduced this against the reference Access controller on exact head `7a049ab1`: `access_list_events(limit=0)` returned a successful response containing 78 events. This controller interprets `page_size=0` as uncapped, so the PR does not satisfy the documented zero-result contract and expands the original symptom from 30 events to the full result set.

Please short-circuit `limit == 0` to return `[]` before `proxy_request`, then update the regression test to assert both the empty result and `mock_req.assert_not_awaited()`. Keep this equality-specific if negative-limit semantics remain intentionally out of scope.

The three in-memory buffer fixes look correct. I also confirmed Ruff/format and generated-artifact checks, the focused Access/Protect/Network tests, and the full 1,787-test Core suite pass on this head.

## Minor Items (nice-to-have)

None.

AmirF194:
You're right, page_size=0 isn't safe to trust across controllers. Pushed 2ac3582:
`list_events` now returns `[]` locally before calling `proxy_request` when
`limit == 0`, and the test asserts the empty result plus
`mock_req.assert_not_awaited()` instead of just checking the query param.

Left negative-limit semantics alone as you suggested, only the equality case.

sirkirby (approved):
Amir addressed the blocking controller-contract issue: Access now short-circuits `limit=0` locally before `proxy_request`, and the regression asserts both an empty result and no controller request.

I independently revalidated exact rebased head `085a26a9`:
- Targeted live Access call: `access_list_events(limit=0)` returned success with count 0 and zero events.
- Full safe live smoke passed across Network, Protect, and Access, including disposable lifecycle cleanup.
- Focused regressions passed: Access 22, Protect 19, Network buffer 7.
- Full Core suite passed: 1,787 tests.
- Format, Ruff lint, generated-artifact drift, and all 14 GitHub checks passed.
- The branch was cleanly rebased onto current `main`.

No remaining findings.

AmirF194:
Appreciate you reproducing it against the live controller both times, that's what actually nailed the fix down. Thanks for merging.

sirkirby:
My pleasure. I'm still testing on my live infra, but I'm setting up a dedicated Cloud Gateway Ultra specifically for testing. My long weekend project :)

Issue #619, opened 2026-09-04
protect_alarm_list_profiles returns [] while the unused AlarmManagerService.list_profiles() returns the profiles


alexpfau:
## Summary

On a console with two arm profiles in daily use, `protect_alarm_list_profiles` returns `{"profiles": [], "count": 0}`. The repo already contains a working reader for the endpoint that does return them — it just has no callers.

## What I see

```
protect_alarm_list_profiles  ->  {"profiles": [], "count": 0}
protect_alarm_get_status     ->  {"armed": false, "status": "disabled", "profile_count": 0}
```

Both profiles exist and are armed daily by webhook automations. `protect_alarm_list_rules` returns rules referencing their ids with recent `stats.executions_24h`.

Testing the three candidate endpoints with an authenticated session:

| Endpoint | Result |
| --- | --- |
| `GET /v1/arm-profiles` (public) | `400 "This operation is not available when global alarm manager is enabled"` |
| `GET /proxy/protect/api/arm/profiles` (private — what this tool calls) | `200` with `[]` |
| `GET /api/v2/alarms/profiles` | `200` with **both profiles** |

The v2 response:

```jsonc
[ {"id": "…", "title": "Away", "state": "disarmed", "state_set_at": "…", "alarm_ids": [...]},
  {"id": "…", "title": "Home", "state": "disarmed", "state_set_at": "…", "alarm_ids": [...]} ]
```

So the tool is faithfully reporting an empty list from an endpoint that returns one. It's looking in the wrong place.

## The reader already exists

```python
# packages/unifi-core/src/unifi_core/protect/managers/alarm_manager_service.py
_API_PATH = "/api/v2/alarms/"

async def list_profiles(self) -> List[Dict[str, Any]]:
    """Return UniFi OS alarm arm profiles (raw; empty when none configured)."""
    data = await self._api_get("profiles")          # -> GET /api/v2/alarms/profiles
    return data if isinstance(data, list) else []
```

A repo-wide search finds no callers. `protect_alarm_list_profiles` calls the legacy `alarm_manager.list_arm_profiles()` instead. It looks like the rules tools were migrated to `AlarmRulesFacade` and the profile/status tools were left on the legacy path.

## This also fixes `protect_alarm_get_status`

Each profile carries its own `state`, and it tracks reality. Arming via webhook and polling:

```
baseline     {'Away': 'disarmed', 'Home': 'disarmed'}
ARM HOME ->  204
[...]

Issue #523, opened 2026-08-11
Protect: expose UP-AirQuality sensor readings (PM2.5/PM10/VOC/CO2/temp/humidity/AQI) in protect_list_sensors


Trebz1:
### Preflight checklist

- [x] I searched existing issues and discussions for similar requests.
- [x] This request fits the project scope described in the README and CONTRIBUTING.md.

### Component

Protect MCP server

### Problem or use case

I have a UniFi Protect Air Quality sensor (UP-AirQuality, wired, adopted on a UDM-PRO running Protect 7.2.101) and I want to read its measurements through the Protect MCP server. Right now that isn't possible.

`protect_list_sensors` returns only identity fields for the device:

```json
{"sensors": [{"id": "<redacted>", "name": "Air Quality-Living Room", "type": "UP-AirQuality"}], "count": 1}
```

No readings, no battery/connection state, no `lastSeen`. The tool description advertises "current readings" for environmental sensors, so this looks like the UP-AirQuality device type simply isn't mapped the way UP-Sense is.

The Network MCP server is no help here either, since it only sees the sensor as a wired client with an IP. Protect is the only place these readings live, so there is currently no path to air quality data through this project.

### Proposed solution

Map the UP-AirQuality device type so its measurements come back alongside the other sensor types. Concretely, `protect_list_sensors` would include the readings the device reports: PM2.5 and PM10 (ug/m3), VOC, CO2 (ppm), temperature, humidity, and the composite AQI score Protect shows in the app. It would also help to return the standard device fields the other sensor types already provide: connection state, `lastSeen`, and per-reading timestamps.

A dedicated `protect_get_sensor_readings(sensor_id)` tool would work just as well if keeping the list tool lightweight matters. Either shape is fine; the important part is that the values are reachable at all.

### Alternatives considered

Reading the values by hand in the UniFi Protect app works, but it can't be automated or scheduled, which is the whole reason for using the MCP server. Going through the Network MCP server isn't viable: it only sees the sensor as a wired client and has no sensor endpoints. `protect_update_sensor_settings` (added in #422) covers the write side for sensor configuration, but there's no matching read path for the measurements themselves.

### Implementation notes

Environment: unifi-protect-mcp 0.7.5 via uvx, UDM-PRO, Protect 7.2.101 / firmware 5.1.29, sensor is a wired UP-AirQuality adopted 2026-08-10. Auth is a local admin account (no MFA), no Protect public API key configured.

[...]

PR #620, opened 2026-09-04
fix(core): add Alarm Manager v2 profile bridge

alexpfau:
## Summary

Refs #619.

This is the Core-first half of the Alarm Manager profile fix. It adds the reusable bridge required by the Protect and API surfaces without publishing downstream packages against an unreleased Core method:

- route profile discovery through Alarm Manager v2, with the legacy endpoint as the narrow compatibility fallback
- normalize v2 profile fields, including state and provenance
- translate legacy profile-selection 404s into the domain error
- preserve unknown controller state values
- cover facade, model, and legacy profile-ID behavior in Core tests

After this merges, maintainers will release `unifi-core==0.4.40` and restore the reviewed Protect/API surface in a follow-up pinned to `unifi-core>=0.4.40`.

## Testing

- `make pre-commit` — passed
- `unifi-core` — 1,799 passed
- full repository suites and generated-artifact drift checks — passed
- contributor live validation: Protect 7.2.105 on UNVR G2 in Global Alarm Manager mode; v2 returned the configured profiles and their current state

The original combined implementation is preserved for the downstream follow-up. No further contributor action is needed.

### DIFF
--- a/packages/unifi-core/src/unifi_core/protect/managers/alarm_facade.py
+++ b/packages/unifi-core/src/unifi_core/protect/managers/alarm_facade.py
@@ -30,6 +30,7 @@ from unifi_core.protect.models.alarm_rules import (
     alarm_rule_to_legacy_create_body,
     alarm_rule_to_v2_body,
 )
+from unifi_core.protect.models.alarms import profile_from_controller
 
 _UUID_RE = re.compile(r"^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$")
 
@@ -66,6 +67,48 @@ class AlarmRulesFacade:
             return rules, True
         raw = await self._legacy.list_rules()
         return [alarm_rule_from_legacy(rule).model_dump(exclude_none=True) for rule in raw], False
+
+    async def list_profiles(self) -> tuple[list[dict[str, Any]], bool]:
+        """Return ``(profiles, complete)``.
+
+        Prefer the v2 Alarm Manager profile API, but fall back to the legacy
+        arm-profile API whenever v2 cannot serve profiles on this console — that
+        is, when it returns an empty list or rejects the request. ``complete`` is
+        True only when v2 actually served the profiles.
+
+        Transient/server failures (5xx, timeouts -> ``NvrError``) are NOT masked:
+        they propagate so a real v2 outage is never silently hidden behind legacy.
+        """
+        permission_error: AlarmManagerPermissionError | None = None
+        try:
+            profiles = await self._service.list_profiles()
+        except AlarmManagerPermissionError as exc:
+            permission_error = exc
+            profiles = None
+        except BadRequest:
+            profiles = None
+        if profiles:
+            shaped = []
+            for profile in profiles:
+                if not isinstance(profile, dict):
+                    continue
+                item = profile_from_controller(
+                    {**profile, "id_family": "alarm_manager_v2", "arm_compatible": False}
+                ).model_dump(exclude_none=True)
+                shaped.append(item)
+            return shaped, True
+        raw = await self._legacy.list_arm_profiles()
+        if raw:
+            return [
+                profile_from_controller({**profile, "id_family": "legacy_arm", "arm_compatible": True}).model_dump(
+                    exclude_none=True
+                )
+                for profile in raw
+                if isinstance(profile, dict)
+            ], False
+        if permission_error is not None:
+            raise permission_error
+        return [], False
 
     async def get_rule(self, rule_id: str) -> tuple[dict[str, Any], bool]:
         """Return ``(rule, complete)`` for one rule by id.
--- a/packages/unifi-core/src/unifi_core/protect/managers/alarm_manager.py
+++ b/packages/unifi-core/src/unifi_core/protect/managers/alarm_manager.py
@@ -50,6 +50,8 @@ import logging
 from datetime import datetime, timezone
 from typing import Any, Dict, List, Optional, Tuple
 
+from uiprotect.exceptions import BadRequest
+
 from unifi_core.exceptions import UniFiNotFoundError
 from unifi_core.protect.managers.connection_manager import ProtectConnectionManager
 
@@ -57,6 +59,13 @@ logger = logging.getLogger(__name__)
 
 # Values of ``nvr.armMode.status`` that mean "not armed".
 _DISARMED_STATUSES = {"disabled", "disarmed", "off", "inactive"}
+_LEGACY_PROFILE_ID_NOT_FOUND_ERROR = (
+    "The legacy Protect arm endpoint did not recognize this profile_id. Alarm Manager "
+    "v2 ids returned by protect_alarm_list_profiles are not accepted by that endpoint; "
+    "use a profile marked arm_compatible=true, or omit profile_id to use the currently "
+    "selected legacy profile. A v2 arming endpoint is not implemented because its "
+    "contract has not been verified."
+)
 
 
 class AlarmManager:
@@ -172,11 +181,16 @@ class AlarmManager:
 
     async def _select_profile(self, profile_id: str) -> None:
         """PATCH ``arm`` to set which profile is active."""
-        await self._cm.client.api_request(
-            "arm",
-            method="patch",
-            json={"armProfileId": profile_id},
-        )
+        try:
+            await self._cm.client.api_request(
+                "arm",
+                method="patch",
+                json={"armProfileId": profile_id},
+            )
+        except BadRequest as exc:
+            if _is_arm_profile_not_found(exc):
+                raise ValueError(_LEGACY_PROFILE_ID_NOT_FOUND_ERROR) from exc
+            raise
 
     # ------------------------------------------------------------------
     # Preview (for confirm=false tool responses)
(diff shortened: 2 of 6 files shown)


alexpfau:
Thanks for the careful review. I agree all four points were valid, and I pushed fixes for them.

1. Action dispatch tuple handling

`protect_alarm_list_profiles` now has an action-dispatch read-result adapter that unwraps the facade's `(profiles, complete)` return value before `AlarmProfileList.from_manager_output()` sees it. That keeps the action endpoint response shaped as `{profiles, count}` instead of falling back to a stringified tuple. I added an action-endpoint regression test for that exact path.

2. Cross-family profile IDs and arming

I did not try to invent or guess a v2 arming endpoint. The legacy arming tool still calls the legacy `PATCH /arm` endpoint, but the model and tool documentation now make profile provenance explicit. v2 profile IDs are tagged as `id_family="alarm_manager_v2"` and `arm_compatible=false`; legacy arm-profile IDs are tagged as `id_family="legacy_arm"` and `arm_compatible=true`.

That field shape is a proposal, not something I am attached to. If you would rather avoid adding public model fields, I am happy to reshape it or drop the fields and rely on documentation or another compatibility marker.

For the arming error itself, I changed the guard to react to the controller's response instead of guessing from ID shape. If the legacy `PATCH /arm` call returns the live-captured arm-profile 404 shape (`entity: armProfile`, `name: NOT_FOUND`, `Entity 'armProfile' not found`), the tool now translates that into an actionable error explaining that Alarm Manager v2 IDs from `protect_alarm_list_profiles` are not accepted by the legacy arm endpoint and that `arm_compatible` marks usable IDs. This avoids false-positive blocking if a working legacy controller ever uses UUID-shaped IDs.

3. Other read surfaces still on legacy profiles

The GraphQL `alarmProfiles` resolver and the REST `/sites/{site_id}/alarm-profiles` endpoint now route through `AlarmRulesFacade.list_profiles()` too, and unwrap the completeness flag. They therefore get the same v2 profile coverage as the MCP/action-catalog path on Global-mode consoles. I added coverage for the Global-mode/v2 profile shape.

4. Permission failures masked as empty profile lists

`AlarmRulesFacade.list_profiles()` now preserves an `AlarmManagerPermissionError` from the v2 endpoint. If the legacy fallback cannot supply profiles, the facade re-raises that permission error instead of returning an empty list. That way a Protect-scoped non-SuperAdmin account sees a credential/remediation failure instead of a misleading `count: 0`.

I also regenerated the affected API/catalog/schema/docs artifacts and added tests around the changed behavior.

---

**Validation.** `ruff check` and `ruff format --check` clean, generated artefacts show no drift, and the suites pass: 430 in `unifi-core`'s protect tests, 507 in `apps/protect`. The 3 failures in `apps/protect/tests/unit/test_bootstrap_dotenv.py` are pre-existing — they fail identically on `main` with this branch not checked out, and look environment-sensitive rather than real.

`make pre-commit` still cannot complete on my machine: `worker-install` runs `npm ci` and gets an `E404` for `wrangler-4.127.1.tgz` from a corporate registry proxy. That is local to me, not the diff — so the worker typecheck step is the one part of the gate I have not been able to exercise.

One note on the 404 translation: the matcher keys off the structured payload (`entity == "armProfile"`) with a string fallback, and the shape it matches was captured from a real controller response rather than inferred — I tested `PATCH /arm` with a v2 profile UUID directly and got `{"error":"Entity 'armProfile' not found","name":"NOT_FOUND","entity":"armProfile","id":"…","idKey":"id"}`. Worth knowing if that shape differs across firmware versions; I have only the one console to check against.

apps/api/src/unifi_api/routes/resources/protect/system.py
```diff
@@ -162,11 +167,11 @@ async def alarm_list_profiles(
             session,
             controller.id,
             "protect",
-            "alarm_manager",
+            "alarm_facade",
         )
         cm = await factory.get_connection_manager(session, controller.id, "protect")
         await _maybe_set_site(cm, site_id)
-        items = await mgr.list_arm_profiles()
+        items, _complete = await mgr.list_profiles()
```
sirkirby: The facade can now re-raise `AlarmManagerPermissionError` when v2 is forbidden and the legacy fallback returns no profiles. This REST route leaves it unhandled, producing HTTP 500 with only `Internal Server Error`; MCP, action dispatch, and GraphQL preserve or map the actionable permission result. Please catch this at the REST boundary, return HTTP 403 with `detail=str(exc)`, and add a regression test for the non-SuperAdmin plus empty-legacy path.

sirkirby (changes requested):
Thanks for the quick, thoughtful response to the Copilot review. I independently checked the latest head and confirmed that the four original findings are addressed. Two hard blockers remain before merge.

- [ ] Raise the `unifi-core` minimum to `>=0.4.39` in both `apps/api/pyproject.toml` and `apps/protect/pyproject.toml`. These surfaces now call `AlarmRulesFacade.list_profiles()`, but their published floors still permit Core 0.4.37/0.4.38, where that method is absent. The current pin-alignment failure confirms the API break; I also reproduced a clean resolution to Core 0.4.38 with `has_list_profiles=false`, and the Protect tool has the same requirement even though that half is not caught by the failing job.
- [ ] Preserve `AlarmManagerPermissionError` at the REST boundary. When v2 returns the actionable SuperAdmin error and legacy has no profiles, `/v1/sites/{site_id}/alarm-profiles` currently lets the exception escape. I exercised that route with `raise_app_exceptions=false` and received HTTP 500 with only `Internal Server Error`. Please map it to HTTP 403 with the actionable detail and add a REST regression test.

## Validation

I ran the full local pre-commit gate on the preceding head; the latest follow-up only changed `alarm_manager.py` and its tests. At exact head `965242ae`, the affected 25 tests, Ruff, generated-artifact checks, lock validation, and diff checks pass. Branch-local reference-controller smoke for `protect_alarm_list_profiles` also passes, although that controller has zero profiles, so your Global-mode evidence remains the non-empty topology check.

This review is intentionally scoped to #619/#620. The sensor discussion in #523 should remain a separate follow-up.

alexpfau:
Thanks — both blockers are fixed in `ed5be3f`.

**Core floor.** Raised to `>=0.4.39,<0.5` in both `apps/api/pyproject.toml` and `apps/protect/pyproject.toml`. You're right that Protect had the same requirement despite the job not catching it — `protect_alarm_list_profiles` calls the same facade method, so its `>=0.4.35` floor was equally wrong. I'd been treating the sequencing as an open question, since the published ceiling is 0.4.38 and I could see from history that floor bumps normally land as separate `chore(deps)` PRs after a Core release; thanks for settling it.

**REST permission mapping.** `/v1/sites/{site_id}/alarm-profiles` now catches `AlarmManagerPermissionError` and returns HTTP 403 with `detail=str(exc)`, matching the pattern the neighbouring routes already use for `UniFiNotFoundError`. Added `test_alarm_list_profiles_permission_error_maps_to_403`, which stubs the facade to raise and asserts both the 403 and that the remediation text survives into `detail`.

I checked that test actually catches the regression rather than passing vacuously: with the route's `try/except` reverted it fails with the raw `AlarmManagerPermissionError` propagating, and passes once restored.

**Validation.** Ruff lint and format clean, action-catalog generation reports no drift, `uv lock --check` still resolves after the pyproject edits, 430 tests in `unifi-core`'s protect suite, 1361 in `apps/api`.

Two failures I want to name rather than leave for you to trip over, both of which reproduce on `main` with this branch not checked out:

- `apps/api/tests/test_action_catalog_packaging.py::test_built_wheel_contains_and_loads_catalog_without_sibling_apps`
- the three `apps/protect/tests/unit/test_bootstrap_dotenv.py` parametrisations

Both build or install into a fresh environment, and my `npm`/registry setup is behind a corporate proxy that also blocks `make pre-commit` at `worker-install` (`E404` for `wrangler-4.127.1.tgz`). I believe they're local, but I can't fully rule them out from here — worth a glance if CI disagrees.

Noted on scope: keeping #523 as a separate follow-up. And agreed that your reference controller having zero profiles means the non-empty Global-mode topology is only covered by my console — happy to run any further sequence against it if that would help.

alexpfau:
Both blockers are implemented in `ed5be3f`, but the floor bump has run into the other half of the pin-alignment gate, and I don't think there's a value that satisfies both sides in a single PR.

**The two states I've now measured:**

| `unifi-core` floor | Pin-alignment result |
|---|---|
| `>=0.4.37` / `>=0.4.35` (before) | ❌ `[FAIL] API catalog against published Core floor` → `protect_alarm_list_profiles: missing alarm_facade.list_profiles` |
| `>=0.4.39` (after) | ❌ `ERROR: Could not find a version that satisfies the requirement unifi-core<0.5,>=0.4.39` — highest published is `0.4.38` |

The exact-Core-floor contract needs a published Core that contains `AlarmRulesFacade.list_profiles()`; the PyPI resolution step needs the floor to name a version that exists. No published version satisfies both, so the gate blocks every floor value while the method is unreleased.

The gate's own output points at the resolution:

> **Premature pin bump**: a downstream now requires an upstream version that has not been published to PyPI yet. Split the change: merge and release the upstream first, then open a follow-up PR that bumps the downstream pin and adds the imports.

…which matches `monorepo-release-pipeline` Procedure D, Step 0 ("runs on every PR and cannot be skipped").

**Happy to split it whichever way suits your release flow.** The shape I'd suggest, unless you'd rather do it differently:

1. **This PR, reduced to Core only** — `AlarmRulesFacade.list_profiles()`, the `AlarmProfile` model additions, the legacy arm-profile 404 translation, and their tests. No `pyproject.toml` floor changes, no `action_catalog.json` rebinding, no GraphQL/REST routing, so both halves of the gate stay green.
2. **Follow-up PR after Core 0.4.39 is on PyPI** — raise the floors in `apps/api` and `apps/protect`, rebind the action catalog, route GraphQL/REST through the facade, and carry the REST 403 mapping plus its regression test.

That does temporarily re-open your review points (1) and (3), since the tuple adapter and the other read surfaces belong with the downstream half — I'd carry both into the follow-up rather than drop them.

Alternatively, if you'd prefer to cut Core 0.4.39 first and have me rebase this as-is, that works too and keeps everything in one review. Just say which and I'll restructure — no strong preference from my side, you know the release cadence.

One note in case it's useful: Procedure D, Step 0 already documents the inverse gap ("a PR that adds code using a new cross-package API without updating the floor bound passes CI"). This is the mirror image — updating the floor for an unreleased upstream fails the gate — and both arise from the same single-PR-touching-both-halves shape.

sirkirby:
Agreed — let's use the Core-first split. I'll take it from here to avoid another round trip: I'm going to trim this branch to the Core implementation and tests, rebase it onto current `main`, run the gates, and get Core merged and released as 0.4.40. Once that wheel is on PyPI, I'll restore the reviewed Protect/API pieces in a follow-up with `>=0.4.40` floors and regenerated artifacts. No further action needed from you.

sirkirby (approved):
Core-first split verified at 505537be. The full local pre-commit gate and all 14 GitHub checks pass; the prior downstream floor and REST-boundary blockers are intentionally removed from this package-boundary PR and will be handled after Core 0.4.40 is published.

sirkirby:
Core-first split is complete: #620 merged, `unifi-core==0.4.40` is published and live-smoke verified, and the Protect/API half is now in #624 with `>=0.4.40` floors and regenerated artifacts.

PR #630, opened 2026-09-04
fix(network): validate destination/source port shape in firewall policies

### LINKED ISSUE #608: create_firewall_policy: destination port field is undocumented and preview accepts an invalid shape
## Summary

`unifi_create_firewall_policy` cannot create a policy that matches a destination port using the
documented input shape. The docstring never describes port targeting, and the field the controller
actually requires is `port` (a string), not `ports` (an array). Making this worse, the preview path
returns success for the wrong shape, so the failure only surfaces on the write.

## Environment

- `unifi-network-mcp` 0.29.3, installed via the `unifi-network` Claude Code plugin 0.29.3
- UniFi Network application 10.5.67
- UniFi Dream Machine Pro SE, UniFi OS firmware 5.1.31.34074
- Zone-based (V2) firewall

Not tested against Network 10.6.x, which reached the release channel as 10.6.101 on 2026-08-27, so
this may be specific to 10.5.x if the destination port shape changed.

## What the docstring says

The tool description documents targeting for IPs, networks and "any in zone", and lists the optional
keys:

> For IP targeting: `matching_target='IP'`, `matching_target_type='SPECIFIC'`, `ips=[...]`.
> For network targeting: `matching_target='NETWORK'`, `matching_target_type='OBJECT'`, `network_ids=[...]`.
> For any in zone: `matching_target='ANY'`.
> Optional: `enabled`, `description`, `protocol`, `connection_state_type`, `connection_states`, `ip_version`, `schedule`, `logging`.

Port matching is not mentioned anywhere, even though `port_matching_type` appears in the output of
`unifi_get_firewall_policy_details` for existing policies.

## Reproduction

Goal: allow one host on one zone to reach one host on another zone, on TCP 445 only.

Given `ips` is the documented plural form for addresses, `ports` is the natural guess for ports:

```json
{
  "name": "test",
  "action": "ALLOW",
  "enabled": true,
  "protocol": "tcp",
  "ip_version": "IPV4",
  "source": {
    "zone_id": "<untrusted zone id>",
    "matching_target": "IP",
    "matching_target_type": "SPECIFIC",
    "ips": ["10.0.28.10"],
    "port_matching_type": "ANY"
  },
  "destination": {
    "zone_id": "<internal zone id>",
    "matching_target": "IP",
    "matching_target_type": "SPECIFIC",
    "ips": ["10.0.15.10"],
    "port_matching_type": "SPECIFIC",
    "ports": ["445"]
  }
}
```

### Step 1, preview (`confirm` omitted)

Returns `success: true` with `requires_confirmation: true`, and echoes `"ports": ["445"]` back inside
`preview.will_create.destination` unchanged.

### Step 2, same payload with `confirm: true`

```json
{
  "success": false,
  "error": "Failed to create firewall policy 'test': {'code': 'api.err.MissingFirewallDestinationPort', 'details': {}, 'errorCode': 400, 'message': 'Missing firewall destination port'}"
}
```

### Step 3, the shape that works

Replacing `"ports": ["445"]` with `"port": "445"` succeeds, and the created policy reads back with:

```json
"destination": {
  "ips": ["10.0.15.10"],
  "matching_target": "IP",
  "matching_target_type": "SPECIFIC",
  "port": "445",
  "port_matching_type": "SPECIFIC",
  "zone_id": "..."
}
```

So addresses are a plural array and the port is a singular string, which is an easy asymmetry to
get wrong when the docstring documents one and not the other.

## Suggested fixes

1. **Document port targeting in the docstring**, naming `port_matching_type` and the `port` string,
   and calling out that it is singular while `ips` is plural.
2. **Validate in the preview path.** Right now `confirm: false` echoes whatever it is given, which
   means the preview provides no signal about whether the write will succeed. Rejecting unknown keys
   inside `source` and `destination`, or mapping a `ports` array onto `port`, would catch this before
   the write.
3. Optionally **accept `ports` as an alias** and join it, since a port group or range is expressible
   as a single string anyway.

AmirF194:
Firewall policy creation now rejects plural `ports` and requires a non-empty string for `port_matching_type='SPECIFIC'` before preview or controller execution. Both source and destination are covered. The MCP wrapper and API argument translator use the shared Core validator, so API callers receive the same fail-fast behavior.

The Core prerequisite landed in #662 and is published as `unifi-core==0.4.45`. This PR raises Network/API dependency floors accordingly and documents the accepted port shape in generated tool descriptions.

Fixes #608.

Validation on final head `7b61c380c00669d7d25e1712d8c3307e7a5dc75b`:

- All 14 fresh CI checks pass, including published-package pin alignment.
- Rebased Network firewall tests: 147 passed. API dispatcher tests: 327 passed. Generated artifacts match.
- Full `make pre-commit` passed before the rebase; the rebase only removed the separately merged Core prerequisite.
- Maintainer reference-hardware testing rejected 48 malformed-port cases across MCP/API, source/destination, and preview/execute. Two disabled-policy create/update/readback/delete cycles verified port preservation and left zero orphaned policies.
- Network safe smoke: 160 successful checks, no failures, three disposable resources cleaned and gateway settings restored. Published Core 0.4.45 also passed independent Network/Protect/Access installed-wheel smoke and a targeted disabled-policy lifecycle.
- Copilot's validation and comment-drift feedback is addressed. Its request for live-controller verification is covered above.

<details>
<summary>Targeted maintainer hardware probe output</summary>

```text
mcp: malformed ports rejected in preview and execution
api: malformed ports rejected in preview and execution
mcp: disabled policy created, ports preserved after update, and deleted
api: disabled policy created, ports preserved after update, and deleted
{"invalid_cases_rejected": 48, "cycles": ["mcp: create/update/readback passed", "api: create/update/readback passed"], "orphans": 0}
```

The valid test policies were disabled, used explicit string ports and `create_allow_respond=false`, and were deleted after readback verification.

</details>

### DIFF
--- a/apps/api/pyproject.toml
+++ b/apps/api/pyproject.toml
@@ -18,6 +18,7 @@ dependencies = [
     # require Core 0.4.39. UTC device timestamps in action read views require
     # Core 0.4.43.
     # Firewall index rejection and uncapped client lookup require Core 0.4.44.
+    # Shared firewall create-port validation requires Core 0.4.45.
     "unifi-core[network,protect,access]>=0.4.45,<0.5",
     # The API server is a deployable Network client too. Pin the authentication
     # transport exactly so a released API version has a reproducible login stack
--- a/apps/api/pyproject.toml
+++ b/apps/api/pyproject.toml
@@ -18,7 +18,7 @@ dependencies = [
     # require Core 0.4.39. UTC device timestamps in action read views require
     # Core 0.4.43.
     # Firewall index rejection and uncapped client lookup require Core 0.4.44.
-    "unifi-core[network,protect,access]>=0.4.44,<0.5",
+    "unifi-core[network,protect,access]>=0.4.45,<0.5",
     # The API server is a deployable Network client too. Pin the authentication
     # transport exactly so a released API version has a reproducible login stack
     # even when installed from PyPI without the workspace uv.lock.
--- a/apps/api/src/unifi_api/action_catalog.json
+++ b/apps/api/src/unifi_api/action_catalog.json
@@ -2526,7 +2526,7 @@
       "input_schema": {
         "properties": {
           "policy_data": {
-            "description": "V2 zone-based firewall policy dict. Required: name, action (ALLOW/BLOCK/REJECT), source (object with zone_id, matching_target), destination (same structure). For IP targeting: matching_target='IP', matching_target_type='SPECIFIC', ips=[...]. For network targeting: matching_target='NETWORK', matching_target_type='OBJECT', network_ids=[...]. For any in zone: matching_target='ANY'. Optional: enabled, description, protocol, connection_state_type, connection_states, ip_version, schedule, logging.",
+            "description": "V2 zone-based firewall policy dict. Required: name, action (ALLOW/BLOCK/REJECT), source (object with zone_id, matching_target), destination (same structure). For IP targeting: matching_target='IP', matching_target_type='SPECIFIC', ips=[...]. For network targeting: matching_target='NETWORK', matching_target_type='OBJECT', network_ids=[...]. For any in zone: matching_target='ANY'. For port targeting: port_matching_type='SPECIFIC', port='445' (a single string, not a 'ports' array). Optional: enabled, description, protocol, connection_state_type, connection_states, ip_version, schedule, logging.",
             "type": "object"
           }
         },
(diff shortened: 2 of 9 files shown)


AmirF194:
Good catch. Fixed: port now has to be a non-empty string, not just truthy, so ["445"] or the int 445 get the same SPECIFIC error as a missing port instead of reaching the controller. Added a test covering both malformed shapes. Pushed 761238b.

sirkirby (approved):
Thanks Amir. I verified your port-shape fix, completed the shared Core/API validation wiring, and addressed the Copilot feedback. Core 0.4.45 is now published, and all 14 checks pass on 7b61c380, including pin alignment. I also independently verified malformed-port rejection and disabled-policy create/update/readback/delete cycles on the test controller, with no leftover policies.

AmirF194:
Good to hear the live-controller round trip held up, that's more than I could verify locally. Makes sense on the pin failure too, it was waiting on core 0.4.45 to publish.

PR #637, opened 2026-09-04
fix(core): render device last_seen as UTC-aware ISO time

tmowbrey:
## Summary

`shape_device_list()` in `unifi-core` was the one remaining bare `datetime.fromtimestamp()` in the network read views (`read_views.py` ~line 293). It rendered `last_seen` in the MCP server's local timezone with no offset, so the same device reported different times depending on where the server ran, and the value could not be compared with the tz-aware timestamps every other view emits. This passes `tz=timezone.utc`, matching the other nine call sites; `last_seen` now ends in `+00:00`.

## Type of change

- [x] `fix:` bug fix

## Scope

- [x] Shared packages

## Checklist

- [ ] I updated generated manifests with `make manifest` when changing MCP tools. (No tool signature changed; manifest unaffected.)
- [ ] I updated documentation for user-visible changes. (No doc describes the `last_seen` format.)

## Testing

Ran the equivalent of `make pre-commit` for the Python side (`ruff format`, `ruff check`, core + network suites); the worker npm install was skipped on this host, and no worker code is touched.

```
TZ=America/Denver uv run --all-extras pytest packages/unifi-core/tests/network/test_read_views.py   # new test fails before the fix, passes after
uv run --all-extras pytest packages/unifi-core/tests   -> 1849 passed
uv run pytest apps/network/tests                       -> 1085 passed
uv run ruff format packages/unifi-core && uv run ruff check packages/unifi-core   -> clean
```

The new test pins the exact UTC string for a fixed epoch, so it only passes with an explicit UTC tz (it fails under a non-UTC `TZ` before the fix).

## Notes for reviewers

User-visible change: `last_seen` gains a `+00:00` suffix. Callers parsing it with `datetime.fromisoformat` are unaffected; callers string-comparing against a naive value would see the difference.


sirkirby (commented):
Tyler, I reviewed `96fb6258` and found no functional issue in the UTC conversion. I reproduced the host-timezone difference with a fixed epoch, and all 13 focused read-view tests passed. The read-only live check on this head returned 47 device timestamps, all timezone-aware UTC. The API action path also uses the corrected projection.

The change was included in the tested current-main combination. Releasing the corrected Core package remains a separate delivery step. This is a review comment; I have not submitted a formal approval.

sirkirby (approved):
Tyler, I approve this revision. I verified this revision in the combined candidate and installed wheels. On the live controller, all 47 device timestamps were UTC-aware; the focused epoch/timezone regression and full repository gate passed. The independent combined review found no issue. The coordinated release will publish Core first and raise affected consumer floors.

PR #639, opened 2026-09-04
fix(network): populate current values in the auto-backup update preview

tmowbrey:
## Summary

`unifi_update_autobackup_settings` built its preview with `current_state={}`, so every field came back `current: null` even when the controller had a value. The preview confirmed which fields were addressed and nothing else: it could not diff, and an operator reading it could not see they were about to replace a working schedule. `unifi_update_snmp_settings` in the same module already reads the live record before previewing; this does the same via `system_manager.get_autobackup_settings()`, and returns the standard `{success: false, error}` shape if that read fails instead of previewing against nothing.

## Type of change

- [x] `fix:` bug fix

## Scope

- [x] Network MCP server

## Checklist

- [x] I updated generated manifests with `make manifest` when changing MCP tools. (Regenerated; no diff, the signature is unchanged.)
- [ ] I updated documentation for user-visible changes. (No doc describes the preview's `current` block.)

## Testing

Python side of `make pre-commit` (`ruff format`, `make manifest`, `make server-manifest`, `ruff check`, network suite); the worker npm install was skipped on this host, and no worker code is touched.

```
uv run pytest apps/network/tests/unit/test_backup_tools.py   # new preview test fails before (current is None), passes after
uv run pytest apps/network/tests                             -> 1086 passed
uv run ruff format / ruff check apps/network                 -> clean
```

## Notes for reviewers

The preview now makes one extra read (`rest/setting` → `super_mgmt`) before returning, same cost as the SNMP preview. Client-side range checks for bounded fields (`autobackup_max_files` has a controller ceiling the preview still cannot see) are a separate change I have not made here.

### DIFF
--- a/apps/api/src/unifi_api/services/dispatch_overrides.py
+++ b/apps/api/src/unifi_api/services/dispatch_overrides.py
@@ -91,6 +91,8 @@ DISPATCH_OVERRIDES: dict[str, tuple[str, str]] = {
     # Gateway/SNMP settings updates pre-fetch current state for previews.
     "unifi_update_gateway_settings": ("gateway_settings_manager", "update_gateway_settings"),
     "unifi_update_snmp_settings": ("system_manager", "update_settings"),
+    # Auto-backup update pre-fetches get_autobackup_settings for the preview.
+    "unifi_update_autobackup_settings": ("system_manager", "update_autobackup_settings"),
     # Firewall: tool layer pre-fetches list to find policy by id.
     "unifi_toggle_firewall_policy": ("firewall_manager", "toggle_firewall_policy"),
     "unifi_get_firewall_policy_details": ("firewall_manager", "get_firewall_policy_by_id"),
--- a/apps/network/src/unifi_network_mcp/tools/system.py
+++ b/apps/network/src/unifi_network_mcp/tools/system.py
@@ -359,11 +359,16 @@ async def update_autobackup_settings(
         return {"success": False, "error": "No valid fields to update after validation."}
 
     if not confirm:
+        try:
+            current = await system_manager.get_autobackup_settings()
+        except Exception as e:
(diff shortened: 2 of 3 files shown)


sirkirby (commented):
Tyler, I rechecked the rebased head and found no blocking issue. The preview now uses existing auto-backup values, and a failed preview read makes no mutation call. I also verified that production API dispatch selects `update_autobackup_settings` and preserves `False` values. The focused tests and catalog check pass on this head.

The getter/explicit dispatch-binding pair fits the existing design. Final CI and delivery checks still apply; I did not exercise a live auto-backup mutation.

Reviewed commit: `2186b3262baea1ec24b25de9c94b5d4a001f2b94`.

sirkirby (commented):
Tyler, I accept this as a preview correctness fix. Showing existing controller values is necessary for an operator to assess a schedule change; it uses the existing getter and does not introduce another scheduling service.

I’ve now completed that targeted smoke on Network 10.6.102: the preview required confirmation, its current value matched the real getter, and its proposed value matched the requested opposite boolean. Settings were unchanged afterward, with two GETs and no mutation requests. That closes the live read/preview gate; I do not need you to repeat it or perform a confirmed update for this preview-only change. Final delivery checks remain separate.

Reviewed/live-tested commit: `2186b3262baea1ec24b25de9c94b5d4a001f2b94`.

sirkirby (approved):
Tyler, I approve this revision. I verified the installed preview against Network 10.6.102: current and proposed settings were populated correctly, confirmation was required, settings remained unchanged, and the trace contained two GETs and zero mutations. The production API dispatch binding also passed the combined probe, selecting the update method with the correct settings payload. The full repository gate and installed Network safe smoke passed. I found no actionable issue.

PR #640, opened 2026-09-04
fix(network): reject index in unifi_update_firewall_policy, point at reorder

tmowbrey:
## Summary

The V2 firewall-policy write DTO has no `index` field. `unifi_update_firewall_policy` accepted `index` anyway (both its description and the `update_data` field description listed it), the controller accepted the request and ignored the field, and the tool then reported `Controller accepted the request but did not apply changes to: index` after the write. This rejects `index` before validation with an error that names `unifi_reorder_firewall_policies`, and removes `index` from both descriptions. Order changes were never possible through this tool; the change is in when the caller finds out.

## Type of change

- [x] `fix:` bug fix

## Scope

- [x] Network MCP server

## Checklist

- [x] I updated documentation for user-visible changes. (Tool description + manifest.)

## Testing

Python side of `make pre-commit` (`ruff format`, `make manifest`, `make server-manifest`, `ruff check`, network suite); the worker npm install was skipped on this host, and no worker code is touched.

```
uv run pytest apps/network/tests/unit/test_firewall_tools.py   # new test fails before (falls through to the not-found path), passes after
uv run pytest apps/network/tests                               -> 1086 passed
uv run ruff format / ruff check apps/network                   -> clean
```

## Notes for reviewers

Deliberately tool-level rather than marking `index` read-only on `FirewallRule`: `MUTABLE_FIELDS` is shared with `apps/api` (`dispatch_overrides.py`), and a model change there would turn this into a silent drop plus "effectively empty" error rather than the explicit pointer. If you would rather fix it at the model boundary for both surfaces, say so and I will redo it that way. `create_firewall_policy` still allows `index`; I have not verified whether the controller honours it on create, so I left it alone.

### DIFF
--- a/apps/api/src/unifi_api/action_catalog.json
+++ b/apps/api/src/unifi_api/action_catalog.json
@@ -6243,7 +6243,7 @@
             "type": "string"
           },
           "update_data": {
-            "description": "Dictionary of V2 zone-based fields to update: name, action (ALLOW/BLOCK/REJECT), enabled, source, destination, protocol, ip_version, index, logging, connection_state_type, connection_states, schedule.",
+            "description": "Dictionary of V2 zone-based fields to update: name, action (ALLOW/BLOCK/REJECT), enabled, source, destination, protocol, ip_version, logging, connection_state_type, connection_states, schedule. index is not accepted here; use unifi_reorder_firewall_policies to change order.",
             "type": "object"
           }
         },
--- a/apps/api/tests/test_actions_dispatcher.py
+++ b/apps/api/tests/test_actions_dispatcher.py
@@ -2376,6 +2376,58 @@ async def test_dispatch_translates_update_firewall_policy_update_data_to_updates
     domain_manager.update_firewall_policy.assert_awaited_once_with(policy_id="p1", updates={"enabled": False})
 
 
+@pytest.mark.asyncio
+@pytest.mark.parametrize("confirm", [False, True])
+async def test_dispatch_update_firewall_policy_rejects_index_before_preview_and_manager(confirm: bool) -> None:
+    """unifi_update_firewall_policy: ``index`` is rejected on the API path for both
+    confirmation states, before a preview is built and before any manager call.
+
+    The MCP wrapper guards this; the API translator must share the same
+    rejection so callers cannot reach the ignored-index partial-update path."""
+    entry = ToolEntry(
+        name="unifi_update_firewall_policy",
+        product="network",
+        category="firewall_policies",
+        manager="",
+        method="",
+    )
+    registry = _registry_with(entry)
+
+    domain_manager = MagicMock()
+    domain_manager.get_firewall_policies = AsyncMock(return_value=[])
+    domain_manager.update_firewall_policy = AsyncMock(return_value=True)
+
+    conn_manager = MagicMock()
+    conn_manager.site = "default"
+    conn_manager.set_site = AsyncMock()
+
+    factory = MagicMock()
+    factory.get_domain_manager = AsyncMock(return_value=domain_manager)
+    factory.get_connection_manager = AsyncMock(return_value=conn_manager)
+
+    with pytest.raises(ValueError, match="unifi_reorder_firewall_policies"):
+        await dispatch_action(
+            registry=registry,
+            factory=factory,
+            session=MagicMock(),
+            tool_name="unifi_update_firewall_policy",
+            controller_id="cid",
+            controller_products=["network"],
+            site="default",
+            args={"policy_id": "p1", "update_data": {"index": 2000, "enabled": True}},
+            confirm=confirm,
+            dispatch_table={
+                "unifi_update_firewall_policy": DispatchEntry(
+                    manager_attr="firewall_manager", method="update_firewall_policy"
+                ),
+            },
+        )
+
+    factory.get_domain_manager.assert_not_called()
+    domain_manager.get_firewall_policies.assert_not_awaited()
+    domain_manager.update_firewall_policy.assert_not_awaited()
+
+
 # ---------------------------------------------------------------------------
 # Network — toggle_port_forward: port_forward_id → rule_id rename
 # ---------------------------------------------------------------------------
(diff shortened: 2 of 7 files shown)


sirkirby (changes requested):
Tyler, I’m requesting one correction: apply the `index` rejection to API calls before both preview and execution.

The [new MCP guard](https://github.com/sirkirby/unifi-mcp/blob/b9f3fb72591768d4830f615abd48a48fe198e237/apps/network/src/unifi_network_mcp/tools/firewall.py#L513-L522) works, but the [API translator](https://github.com/sirkirby/unifi-mcp/blob/b9f3fb72591768d4830f615abd48a48fe198e237/apps/api/src/unifi_api/services/dispatch_overrides.py#L678-L685) bypasses it. Using the production catalog and dispatcher, I reproduced an API preview accepting `{"index": 2000, "enabled": true}` and confirmed execution forwarding both fields to the manager. That leaves API callers exposed to the ignored-index/partial-update behavior this change addresses, despite the updated catalog saying index is not accepted.

Please share the rejection across MCP and API, with tests for both confirmation states asserting no mutation call. The catalog regeneration is clean; the remaining issue is behavior. Recheck the corrected head and its CI before delivery, preserving the ordering API’s separate ID family.

Reviewed commit: `b9f3fb72591768d4830f615abd48a48fe198e237`.

sirkirby (commented):
Tyler, I accept rejecting `index` as a correction to the existing update contract: a caller should not get a partly applied update when the controller ignores ordering. Reordering remains a separate tool family with its own IDs.

After the MCP/API validation gap is fixed, please include targeted controller evidence for the revised candidate: use an owned disposable policy, read its current state, submit an update containing `index` plus another field through both preview and confirmed entry points, and show rejection before a mutation request, unchanged controller read-back and cleanup. A trace asserting zero mutation requests plus sanitized equality results is sufficient; no production policy change or cross-family ID translation is needed. Include the tested SHA, Network version and MCP/API entry points. The existing fixture tests and green CI do not establish the controller-side unchanged-state check. We can supply the maintainer run once the corrected candidate is ready.

Acceptance follow-up at `b9f3fb72591768d4830f615abd48a48fe198e237`.

tmowbrey:
Corrected head: `6f6f075bcf936a164862f9b69c796067e796b82e`.

**Finding → change**
- `apps/api/src/unifi_api/services/dispatch_overrides.py:678-685` bypassed the MCP guard. The rejection now lives in `normalize_policy_update` (`packages/unifi-core/src/unifi_core/network/models/firewall.py`), the one boundary both the MCP tool and the API translator already call, so it runs before preview and before execution on both paths. The MCP-only block at `apps/network/src/unifi_network_mcp/tools/firewall.py:513-522` is removed; that tool now gets the same error through its existing `ValueError` handling.
- Tests, written red-first: `packages/unifi-core/tests/network/models/test_firewall.py` (`index` with and without another field raises naming `unifi_reorder_firewall_policies`); `apps/network/tests/unit/test_firewall_tools.py::test_v2_update_rejects_index_before_any_manager_call[False|True]` asserts no `get_firewall_policies` / `update_firewall_policy` await; `apps/api/tests/test_actions_dispatcher.py::test_dispatch_update_firewall_policy_rejects_index_before_preview_and_manager[False|True]` drives `dispatch_action` with the production catalog and asserts `ValueError` before any manager is resolved. Suites: Core 1875, Network 1085, API dispatcher 299, catalog `--check` clean.
- Ordering stays on `unifi_get_firewall_policy_ordering` / `unifi_reorder_firewall_policies` with their own ID family; nothing there changed.
- Note: the behaviour change is in Core. Happy to follow the #657 pattern (Core release, then floor bump) if you prefer that over shipping it in this PR.

**Controller evidence** — Network 10.6.102, tested SHA above. Disposable DISABLED BLOCK policy (same zone both sides) created and deleted in the run. Request trace wraps `ConnectionManager.request` and counts HTTP methods; no policy identifiers or settings recorded.

| Step | Entry point | Result |
|---|---|---|
| Read before | manager, fresh (cache invalidated) | found, disabled |
| `{"index": 2000, "enabled": true}` preview | MCP `unifi_update_firewall_policy` (`confirm=False`) | success=false, error names reorder tool, **0 requests** |
| same, confirmed | MCP (`confirm=True`) | success=false, error names reorder tool, **0 requests** |
| same, preview | API `dispatch_action` + production `ManifestRegistry` (`confirm=False`) | `ValueError` naming reorder tool, no manager resolved, **0 requests** |
| same, confirmed | API (`confirm=True`) | `ValueError` naming reorder tool, no manager resolved, **0 requests** |
| Read after the four rejections | manager, fresh | raw document equal to the pre-read: **true**; still disabled: true |
| Positive control: `{"name": …}` confirmed | API (`confirm=True`) | manager returned true, **1 PUT** seen by the trace, name changed, index unchanged, still disabled |
| Cleanup | MCP `unifi_delete_firewall_policy` (`confirm=True`) | delete ok (1 request), absent on fresh read: true |

sirkirby (commented):
Tyler, I re-reviewed `6f6f075b`. Moving index rejection into `normalize_policy_update` resolves the MCP/API gap: it is rejected before preview and execution manager resolution on both surfaces. The previous MCP-only guard is removed. Independent focused suites passed: 59 Core, 67 Network and 301 API tests.

Your exact-head disposable-policy record supplies the requested zero-request rejections, unchanged raw read-back, positive control and cleanup. I accept that targeted contributor evidence; no repeat discovery is needed.

No new source blocker found. This is an advance candidate for maintainer integration, with two remaining delivery steps: update from its 10-commit-old base and preserve regenerated artifacts, then publish the shared Core behavior and align the consuming floors before releasing apps. The trial merge onto current main is clean. I will handle that coordinated integration rather than request another mechanical contributor revision.

sirkirby (approved):
Tyler, I approve the corrected head `c917cb84`. The shared index rejection is verified through installed MCP/API preview and confirm paths with zero manager access. Your exact-head disposable-policy unchanged-state/cleanup record remains the targeted controller evidence. The rebase tree matches the reviewed merge preview. Core publication and Network/API minimums will guarantee this behavior in released app installs.

PR #642, opened 2026-09-05
fix(network): use the per-MAC client lookup from Core past the /rest/user cap

### LINKED ISSUE #632: fix: get_client_details / rename_client say "not found" for clients past the 3,000-row /rest/user cap
### Affected component

Network MCP server

### What happened

`unifi_get_client_details` and `unifi_rename_client` report `client '<mac>' not found` for a client the controller demonstrably has a record for. The tool resolves the MAC by scanning `get_clients()` (`/stat/sta`) and `get_all_clients()` (`/rest/user`), and the controller caps `/rest/user` at exactly 3,000 rows. A client outside that window is reported as *absent* rather than *not enumerable*.

Measured on this site: `/rest/user` returns 3,000 rows, `/stat/alluser` returns 55,990.

### Reproduction

```
unifi_get_client_details(mac_address="<mac>")
  -> { "success": false, "error": "Failed to get client details for <mac>: client '<mac>' not found" }

GET /api/s/default/stat/user/<mac>                ->  HTTP 200, data[] length 1, hostname present
GET /api/s/default/stat/user/00:00:00:00:00:01    ->  HTTP 400, data[] length 0     <- control
```

The control matters: the per-MAC endpoint discriminates real MACs from fake ones, so the 200-with-one-record is a genuine hit, and the tool's "not found" is a false negative.

### Why this is worse than a papercut

It is a false negative dressed as a fact. An agent reading "not found" will conclude the client does not exist and act on that (skip a rename, declare a firewall rule's source dead, and so on). A tool that says "not found" when it means "not in my truncated list" manufactures that error.

### Where

`packages/unifi-core/src/unifi_core/network/managers/client_manager.py` `ClientManager.get_client_details()` (~line 119 at `5302363`): both branches iterate a list endpoint; neither consults `stat/user/<mac>`, and the `UniFiNotFoundError` carries no truncation hint.

### Proposed fix

Resolve single-MAC operations via `GET stat/user/<mac>` (authoritative, O(1), no cap) before or instead of scanning the lists. Failing that, distinguish the two cases in the error text: *"not found in the first 3000 user-table rows; the list is truncated"*.

Workaround today: read `_id` from `stat/user/<mac>` directly, then `PUT rest/user/{_id}`.

I am willing to open the PR for the `stat/user/<mac>` path if you confirm the direction (it touches `unifi-core`, so filing first per CONTRIBUTING).

### Environment

### Preflight

tmowbrey:
## Summary

Closes #632. App half of the fix; the Core half is #659 and this branch is rebased on it (merge #659 first, the diff here then shrinks to the app commit).

`get_client_details` scanned `/stat/sta` and `/rest/user`, and the controller caps `/rest/user` at 3,000 rows, so a client past that window was reported as `client '<mac>' not found` although the controller has its record. #659 makes `ClientManager` consult the authoritative per-MAC `GET /stat/user/<mac>` when the scan misses, and raises `UniFiOperationError` ("existence could not be determined") when that lookup fails for any reason other than the controller's own `api.err.UnknownUser` / a 404 answer.

This PR wires the apps to it:
- `unifi_get_top_clients` enriches names with `existence_only=True`, so an entry past the cap does not cost a per-MAC request.
- The API client route maps `UniFiOperationError` to 502 instead of an unhandled 500, and keeps 404 for a real not-found.
- The MCP tool reports an undetermined lookup as an error rather than a not-found (regression only; the tool's error path already did this).

Dependency floor: `unifi-core[network]>=0.4.44` is committed locally and will be pushed once PyPI serves the Core release that carries `existence_only`, per the review.

## Type of change

- [x] `fix:` bug fix

## Scope

- [x] Network MCP server
- [x] API server

## Checklist

- [x] I updated generated manifests with `make manifest` when changing MCP tools. (No tool signature or description changed; `make check-generated` passes with no diff.)
- [x] I updated documentation for user-visible changes. (No user-facing doc describes the lookup path.)

## Testing

```
uv run --package unifi-core --all-extras pytest packages/unifi-core/tests -n auto   # 1961 passed
uv run --package unifi-network-mcp pytest apps/network/tests/unit -n auto           # 1085 passed
uv run --package unifi-api-server pytest apps/api/tests -n auto                     # 1372 passed
uv run --package unifi-protect-mcp pytest apps/protect/tests -n auto                #  513 passed
uv run ruff format --check . && uv run ruff check . && make check-generated && make docs-test
```

New tests here: `apps/network/tests/unit/test_top_clients_existence_only.py`, `apps/network/tests/unit/test_clients_filtering.py` (undetermined lookup → error), `apps/api/tests/routes/resources/test_network_resources.py` (404 vs 502). The per-MAC behaviour suite and the status-parsing regressions live with Core in #659.

Site evidence on the affected site is in the comment below, tied to the head SHA.

## Notes for reviewers

- Measured on the reporting site: `/rest/user` returns exactly 3,000 rows, `/stat/alluser` 55,990; 55 online clients are absent from `/rest/user`, so the merge path is needed, not only the both-miss path.
- No REST/GraphQL surface exposes `existence_only`; the API always performs the full lookup.

### Deferred items

- `get_client_by_ip` still scans the capped lists; an IP lookup past the cap misses.
- `apps/network/src/unifi_network_mcp/tools/stats.py` (`get_client_stats`, `get_top_clients`) logs the client identifier and exception text at ERROR, a pre-existing exception to the Logging rule; worth its own fix.
- The diagnostics channel (`UNIFI_MCP_DIAGNOSTICS`, opt-in) still records the raw request path and error text, as it does for every endpoint.

### DIFF
--- a/apps/api/src/unifi_api/routes/resources/network/clients.py
+++ b/apps/api/src/unifi_api/routes/resources/network/clients.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 from fastapi import APIRouter, Depends, HTTPException, Query, Request
-from unifi_core.exceptions import UniFiNotFoundError
+from unifi_core.exceptions import UniFiNotFoundError, UniFiOperationError
 
 from unifi_api.auth.middleware import require_scope
 from unifi_api.auth.scopes import Scope
@@ -109,6 +109,9 @@ async def get_client(
             client = await mgr.get_client_details(mac)
     except UniFiNotFoundError as exc:
         raise HTTPException(status_code=404, detail=str(exc))
+    except UniFiOperationError as exc:
+        # The per-MAC lookup failed, so existence is undetermined: not a 404.
+        raise HTTPException(status_code=502, detail=str(exc))
     if client is None:
         raise HTTPException(status_code=404, detail="client not found")
 
--- a/apps/api/tests/routes/resources/test_network_resources.py
+++ b/apps/api/tests/routes/resources/test_network_resources.py
@@ -142,6 +142,39 @@ async def test_list_clients_capability_mismatch(tmp_path, monkeypatch) -> None:
 
 
 @pytest.mark.asyncio
+async def test_get_client_detail_maps_manager_errors(tmp_path, monkeypatch) -> None:
+    """Not-found is 404; an undetermined lookup (the per-MAC endpoint failed)
+    is a 502, never a 404 and never an unhandled 500."""
+    from unifi_core.exceptions import UniFiNotFoundError, UniFiOperationError
+    from unifi_core.network.managers.client_manager import ClientManager
+
+    monkeypatch.setenv("UNIFI_API_DB_KEY", "k")
+    app, key, cid = await _bootstrap(tmp_path)
+    _stub_connection(app, cid)
+
+    async def fake_details(self, mac):
+        if mac == "aa:bb:cc:dd:ee:01":
+            raise UniFiNotFoundError("client", mac)
+        raise UniFiOperationError("existence could not be determined")
+
+    monkeypatch.setattr(ClientManager, "get_client_details", fake_details)
+
+    async with AsyncClient(transport=ASGITransport(app=app), base_url="http://test") as c:
+        missing = await c.get(
+            f"/v1/sites/default/clients/aa:bb:cc:dd:ee:01?controller={cid}",
+            headers={"Authorization": f"Bearer {key}"},
+        )
+        undetermined = await c.get(
+            f"/v1/sites/default/clients/aa:bb:cc:dd:ee:02?controller={cid}",
+            headers={"Authorization": f"Bearer {key}"},
+        )
+
+    assert missing.status_code == 404
+    assert undetermined.status_code == 502
+    assert "could not be determined" in undetermined.json()["detail"]
+
+
+@pytest.mark.asyncio
 async def test_get_client_detail_happy_and_404(tmp_path, monkeypatch) -> None:
     monkeypatch.setenv("UNIFI_API_DB_KEY", "k")
     app, key, cid = await _bootstrap(tmp_path)
--- a/apps/network/src/unifi_network_mcp/tools/stats.py
+++ b/apps/network/src/unifi_network_mcp/tools/stats.py
@@ -241,7 +241,7 @@ async def get_top_clients(
         for entry in top_client_stats:
             mac = entry.get("mac")
             if mac:
-                details = await client_manager.get_client_details(mac)
+                details = await client_manager.get_client_details(mac, existence_only=True)
                 if details:
                     raw = details.raw if hasattr(details, "raw") else details
                     if not entry.get("name") and not entry.get("hostname"):
--- a/apps/network/tests/unit/test_clients_filtering.py
+++ b/apps/network/tests/unit/test_clients_filtering.py
@@ -251,6 +251,24 @@ async def test_get_client_details_not_found():
 
 
 @pytest.mark.asyncio
+async def test_get_client_details_reports_an_undetermined_lookup_as_an_error():
+    from unifi_core.exceptions import UniFiOperationError
+
+    with patch("unifi_network_mcp.tools.clients.client_manager") as mock_cm:
+        mock_cm.get_client_details = AsyncMock(side_effect=UniFiOperationError("existence could not be determined"))
+        mock_cm._connection = _mock_conn()
+
+        from unifi_network_mcp.tools.clients import get_client_details
+
+        result = await get_client_details("aa:bb:cc:99:99:99")
+
+    assert result == {
+        "success": False,
+        "error": "Failed to get client details for aa:bb:cc:99:99:99: existence could not be determined",
+    }
+
+
+@pytest.mark.asyncio
 async def test_get_client_details_summary_status_offline_without_signal():
     # Offline/historical clients lack `is_online` and active-connection counters;
     # status must be derived as Offline, not defaulted to Online.
--- a/apps/network/tests/unit/test_top_clients_existence_only.py
+++ b/apps/network/tests/unit/test_top_clients_existence_only.py
@@ -1,0 +1,31 @@
+"""``unifi_get_top_clients`` enriches names through ``get_client_details``.
+
+Every entry it enriches costs a client lookup; a client past the /rest/user
+row cap would otherwise cost a per-MAC controller request per entry, and the
+name enrichment only needs to know the client exists (Core ``existence_only``).
+"""
+
+from types import SimpleNamespace
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import pytest
+
+
+@pytest.mark.asyncio
+async def test_top_clients_enrichment_asks_for_existence_only():
+    with (
+        patch("unifi_network_mcp.tools.stats.stats_manager") as mock_sm,
+        patch("unifi_network_mcp.tools.stats.client_manager") as mock_cm,
+    ):
+        mock_sm._connection = MagicMock(site="default")
+        mock_sm.get_top_clients = AsyncMock(return_value=[{"mac": "aa:bb:cc:dd:ee:ff", "rx_bytes": 1, "tx_bytes": 2}])
+        mock_cm.get_client_details = AsyncMock(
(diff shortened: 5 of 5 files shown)


sirkirby (changes requested):
Tyler, I reviewed `c928e1441dc795927d61da63eab809f597763127` and am requesting changes for two P2 issues:

- **Match the actual HTTP status.** The [fallback check](https://github.com/sirkirby/unifi-mcp/blob/c928e1441dc795927d61da63eab809f597763127/packages/unifi-core/src/unifi_core/network/managers/client_manager.py#L188) matches `404` anywhere in a `ResponseError`. I reproduced an HTTP 429 from `controller404.example` becoming `UniFiNotFoundError`, which turns an undetermined lookup into API 404. The SDK includes URL/body text in this exception. Please read a structured status where available, or match its status-bearing segment, and add regressions for non-404 responses with those digits in the URL or body.
- **Raise the Network Core dependency floor.** The [new call](https://github.com/sirkirby/unifi-mcp/blob/c928e1441dc795927d61da63eab809f597763127/apps/network/src/unifi_network_mcp/tools/stats.py#L244) passes `existence_only=True`, while the [declared floor](https://github.com/sirkirby/unifi-mcp/blob/c928e1441dc795927d61da63eab809f597763127/apps/network/pyproject.toml#L18) still permits Core 0.4.41, whose method lacks that parameter. I verified the old signature rejects the call with `TypeError`; top-client results containing a MAC can therefore fail in a supported installation. Please publish the Core change first, require its version in Network, and verify a fresh installed package.

The 157 focused tests and changed-file lint checks passed. The dependency check above establishes the incompatible signature; it is separate from a fresh-wheel install test.

sirkirby (commented):
Tyler, I want to make the feature decision and remaining controller check explicit. I accept the targeted lookup repair: [#632](https://github.com/sirkirby/unifi-mcp/issues/632) establishes a real false absence, and the per-MAC fallback avoids fetching the whole user history or adding a request to every successful cached lookup.

Your reported 3,000/55,990 counts and known/unknown endpoint controls are useful hardware evidence. What I still need is the repaired tool result on the affected site, tied to the candidate SHA. Please reuse the existing discovery evidence and add a sanitized, read-only comparison for one known target outside the capped user list: list presence/absence, direct per-MAC hit, MCP details success with the correct controller ID, and live-field preservation if that client is online. Include the unknown control and controller/version/topology; counts and assertions are enough, without MACs or hostnames.

The 429/transport classifications belong in fault-injected tests, not deliberate controller failures. No production rename or IP change is needed for this acceptance check. This adds a feature-evidence gate to my earlier technical review; it does not claim the latest pushed head has been re-reviewed.

Acceptance follow-up at `a3cc53f699ca83c5d5a2fcbce8506999b6ebcb77`.

tmowbrey:
Both findings addressed. Tested SHA `40f465d0c6137beb94bc4774f78ab25fa2bb354f` (this branch, rebased on upstream `main` `bc09e3c` + the Core commit `cf269df`).

**Split.** The Core change is now its own PR, #659 (`client_manager.py`, `stats_manager.py`, `connection_manager.py`, `mac.py`, Core tests). This PR carries only the app half: `tools/stats.py` `existence_only=True`, the API client route mapping `UniFiOperationError` → 502, and their tests. The `unifi-core[network]>=0.4.44` floor bump is committed locally and will be pushed once PyPI serves the Core release that carries `existence_only`, so the head stays installable against the current floor until then.

**Status match** (`packages/unifi-core/src/unifi_core/network/managers/client_manager.py:188` → now via `connection_manager.py` `response_status()`): aiounifi's `ResponseError` carries no status attribute; its message is `Call <url> received <status>[: <body>]`, so the parser anchors on `Call <no-whitespace-url> received <3 digits>` and reads that segment only. Regressions written first, in `packages/unifi-core/tests/network/managers/test_client_manager_stat_user.py`: 429 from `controller404.example` → `UniFiOperationError`; 429 whose body mentions 404 → `UniFiOperationError`; 404 from `controller429.example` → not-found (list verdict); `ResponseError` with no status segment → undetermined; transport error with 404 in the URL → undetermined with cause; plus a `response_status` table. The auth circuit's `"received 429" in message` check uses the same parser now.

**Floor** (`apps/network/pyproject.toml:18`): staged as above; not pushed until the Core version exists.

**Site evidence**, read-only, on the affected site, Network Application **10.6.102** (self-hosted, non-UniFi-OS path), site `default`, aiounifi 95, entry points `unifi_get_client_details` (MCP tool function) and `ClientManager.get_client_details` (what rename/set-IP use), each on a cold client cache:

| Check | Result |
|---|---|
| `/stat/sta` rows / `/rest/user` rows / `/stat/alluser` rows | 67 / 3,000 / 55,990 |
| online clients absent from `/rest/user` | 55 |
| target: online, absent from `/rest/user`, present in `/stat/alluser` | true / true / true |
| direct `GET /stat/user/<mac>`: rows, has `_id` | 1 / true |
| MCP details: success | true |
| MCP details: `_id` equals the per-MAC record's `_id` | true |
| MCP details: `mac` equals the target | true |
| MCP details: live fields preserved (every `/stat/sta` field of the record the merge consumed, 66 compared; `first_seen`/`last_seen` are the ISO renderings from #637) | true |
| manager details: success, `_id` equals the per-MAC record's | true / true |
| unknown control: manager raises `UniFiNotFoundError`; MCP `success=false` with a not-found error | true / true |

No production rename or IP change was made. 429/transport classifications are the fault-injected tests above.

Suites on this head: Core 1961, Network unit 1085, API 1372, Protect 513 passed; `ruff format --check`, `ruff check`, `make check-generated`, `make docs-test` clean.

sirkirby (commented):
Tyler, I re-reviewed `40f465d0`, including its exact #659 prerequisite. The lookup/status classification correction and your SHA-tied beyond-cap live record address the prior behavioral requests. Independent focused checks passed: 110 on the Core prerequisite, 43 Network checks and 13 API client-resource checks. I found no new source blocker in the app delta.

The existing Core-floor request remains intentionally unresolved in this pushed head: Network still permits Core 0.4.43, whose `get_client_details` cannot accept the new `existence_only=True` call. The staged floor update is the right next step after the actual version is published. I will coordinate that foundation/floor sequence, then validate the installed combination. No need to repeat the same controller discovery just to replace your evidence.

sirkirby (approved):
Approving. The Core-floor hold that kept this open is discharged: `core/v0.4.47` is published and `main` now pins `unifi-core[network]>=0.4.47`, so the `existence_only=True` call this branch makes is guaranteed to exist rather than merely likely.

Rebased onto current `main`, dropping `cf269dfb` — that is #659, merged as `9e0b1b6`. The diff is now the app delta only: 5 files, +87.

## What this actually delivers

Worth being precise, because the headline symptom in #632 was already fixed upstream. Core 0.4.44 made `ClientManager` consult the authoritative per-MAC `GET /stat/user/<mac>`, so `get_client_details` stopped reporting a beyond-cap client as absent. What remained, and what this PR carries, is the part that surface parity depends on:

- **`apps/api` maps `UniFiOperationError` to 502.** Without it the API route falls through to 404, so a lookup that could not determine existence is reported as *definitively absent* — the same false-negative-dressed-as-fact the issue objected to, just relocated from the manager to the HTTP surface. This is the meaningful half.
- **`get_top_clients` passes `existence_only=True`**, so an enrichment loop over the top-N stops doing a full per-client fetch when all it needs is a presence check.

## Live verification, and one honest gap

Against the reference controller:

```
real MAC, normal lookup            -> found, identity fields present
real MAC, existence_only=True      -> truthy
fabricated 00:00:00:00:00:01       -> UniFiNotFoundError (definitively absent)
```

The discriminator the issue leaned on is intact: a real MAC resolves and a fabricated one raises rather than being reported as found, so the per-MAC path is not simply answering "yes" to everything.

**I could not reproduce the beyond-cap condition itself.** This controller holds 682 clients in both `/rest/user` and `/stat/alluser`, so the 3,000-row window is never reached and there is no client outside it to probe:

```
/rest/user rows   : 682
/stat/alluser rows: 682
clients beyond the window: 0
```

That is an environment gap, not a code doubt — the behaviour is covered by the Core suite and by your SHA-tied beyond-cap record on a site that does exceed the cap. Recording it so the evidence trail is honest about which half came from where.

## Gates

Repo-wide ruff clean, no generated-artifact drift, no f-string loggers, no issue references. Network 1150, API 1426, and the 17 tests this branch adds all pass. No floor change is needed in the diff — `main` already carries it.

Closes #632.

Issue #635, opened 2026-09-04
fix: Network tool-surface papercuts — MAC param spellings, firewall-group shapes, update_wlan description, site settings, empty event buffer


tmowbrey:
### Affected component

Network MCP server

### What happened

A batch of small tool-surface inconsistencies, each of which cost one failed call in a multi-week engagement driving this server from an agent. Individually trivial; collectively the largest single source of wasted calls. Filed together so they can be triaged as one and picked off in small PRs. All re-verified at `main` @ `5302363`.

#### 1. Four spellings for "the MAC" across tool signatures

Counted over tool signatures at `5302363`: `mac_address` ×18, `device_mac` ×14, `client_mac` ×3, `mac` ×1. An agent that learns one spelling from one tool guesses wrong on the next. Suggest accepting the others as aliases and documenting one canonical name, with a deprecation note.

Related: `unifi_list_clients` accepts `fields` (comma-separated string; a list raises a pydantic error), `unifi_list_devices` rejects `fields` entirely.

#### 2. `create_firewall_group` and `update_firewall_group` take different argument shapes

`unifi_update_firewall_group` takes a `group_data` wrapper; `unifi_create_firewall_group` takes flat `name` / `group_type` / `group_members` and errors on `group_data` (`unknown arguments {group_data}`). `firewall.py` ~1057 vs ~1118. Accept both shapes on create, or align update to flat args.

#### 3. `unifi_update_wlan` description advertises ~30 flat fields the schema rejects

The description lists `proxy_arp`, `l2_isolation`, `wpa3_support`, `mac_filter_list`, and so on; the schema accepts only `[confirm, update_data, wlan_id]`. `unifi_update_wlan(wlan_id=…, proxy_arp=false)` → `unknown arguments {proxy_arp}`; the same value inside `update_data` works. Higher cost than the others because it fails on a write to live wireless config after the operator has already approved the action. Align the description with the schema, or accept the documented flat fields and marshal them into `update_data`.

#### 4. `unifi_get_site_settings` under-returns vs its description

Description promises country code, timezone, connectivity monitoring. `SiteSettings` returns `site_id` / `name` / `role` / `country`. Timezone is not retrievable through the MCP at all, which matters because every naive timestamp question ends up being a timezone question.

#### 5. `unifi_recent_events` returns an empty buffer with no hint

[...]

PR #643, opened 2026-09-05
fix(network): say update_wlan and update_network fields live inside update_data

tmowbrey:
## Summary

Part of #635 (item 3).

`unifi_update_wlan`'s description lists ~30 WLAN fields (`proxy_arp`, `l2_isolation`, `wpa3_support`, `mac_filter_list`, ...) but the schema accepts `wlan_id`, `update_data` and `confirm`. An agent that reads the list as parameters calls `unifi_update_wlan(wlan_id=..., proxy_arp=false)` and gets `unknown arguments {proxy_arp}` after the operator already approved the write. `unifi_update_network` in the same module has the identical shape.

Both descriptions now say the fields are keys inside the `update_data` dict, with an example, and that unknown keys are rejected (which `validate_update` already does). The `update_data` parameter description points at the list instead of "See tool description". First sentences are unchanged, so the generated skill reference does not move. Schemas are unchanged; the manifest and action catalog carry only the text.

## Type of change

- [x] `fix:` bug fix

## Scope

- [x] Network MCP server
- [x] API server (regenerated `action_catalog.json`, text only)

## Checklist

- [x] I updated documentation for user-visible changes. (The description is the documentation; `apps/network/docs/tools.md` blurbs are unaffected.)

## Testing

```
uv run pytest apps/network/tests/unit/test_update_data_descriptions.py   # 8 passed
uv run --package unifi-network-mcp pytest apps/network/tests/unit -n auto  # 1080 passed
make check-generated && make catalog-test
```

The new manifest-driven test asserts, for both tools: the schema has exactly the three arguments; the description says the fields live inside `update_data`; every field name the description documents is accepted by the domain validator (`MUTABLE_FIELDS` plus controller aliases), so a future field typo or removal fails the test; the `update_data` parameter text points at the list.

## Notes for reviewers

- AGENTS.md's update-tool anchor `apps/network/src/unifi_network_mcp/tools/wlans.py:update_wlan` points at a file that does not exist; the tool lives in `tools/network.py`. Left for you since AGENTS.md is maintainer-owned.

### Deferred items

- `unifi_update_firewall_policy` (`tools/firewall.py`) lists its accepted keys in the tool description without saying they are `update_data` keys; milder, different validator, separate PR.
- About a dozen other update tools (`dns`, `qos`, `acl`, `port_forwards`, `switch`, `client_groups`, `oon`, `content_filtering`, `dynamic_dns`, and Access `policies`) still describe `update_data` as "Dictionary of fields to update. See tool description for all supported fields."; same treatment, one follow-up.
- Items 1, 4 and 5 of #635 are in flight as separate PRs; item 2 (firewall-group argument shapes) waits for your call on the direction.

### Considered & declined

- Accepting the documented fields as flat parameters and marshalling them into `update_data`: ~30 new schema properties per tool, catalog and REST churn, and it would diverge from `unifi_create_wlan` / `unifi_create_network`, which take a dict for the same reason (a large, controller-version-dependent field set validated in `unifi_core`).


sirkirby (commented):
Tyler, I reviewed `0e696eb6` and found no actionable issue. The wording correctly places the fields inside `update_data`, and both validators support the unknown-field rejection claim. All eight focused description/schema tests and the API catalog drift check passed; the changes preserve the accepted schemas and mutation behavior.

Loaded/lazy discovery of these two descriptions remains a final verification step. I’m leaving this as a review comment without a formal approval.

sirkirby (commented):
Tyler, I support the description-only scope here. The failed call in #635 is a concrete usability problem, and explaining `update_data` fixes the contract the client sees without adding another accepted argument shape.

To make the remaining acceptance check specific, please attach a sanitized discovery result from a supported installed MCP client: both loaded `tools/list` and lazy `unifi_tool_index` with schemas should show the nesting instruction for `unifi_update_wlan` and `unifi_update_network`, with only the resource ID, `update_data`, and `confirm` as arguments. Include the client/app versions and tested commit. No controller write is needed to validate this text change.

I see the current head is `7a0f39ac`; my earlier technical results were on `0e696eb6`, so this scope/evidence note does not claim the new revision has already been revalidated.

Acceptance follow-up at `7a0f39ac5e8902956d621d195c943b53cdd576e8`.

sirkirby (approved):
Tyler, I approve this revision. I verified the update_data guidance through an actual MCP client session against the installed candidate: eager discovery, lazy index discovery, and lazy load followed by tools/list all expose the correct wrapper, required arguments and examples. Tool contracts and permissions are unchanged. The full combined repository gate passed; I found no actionable issue.

PR #644, opened 2026-09-05
fix(network): return timezone, connectivity monitor and NTP from unifi_get_site_settings

tmowbrey:
## Summary

Part of #635 (item 4).

`unifi_get_site_settings` promised "country code, timezone, connectivity monitoring" and returned `site_id`, `name`, `role`, `country`. Timezone was not retrievable through the MCP at all, and every naive-timestamp question ends up being a timezone question. The manager already fetches every section of `GET /get/setting`; the model dropped them.

`SiteSettings` gains read-only `timezone` (locale section), `connectivity_enabled` and `connectivity_uplink_type` (connectivity section), `ntp_servers` (the non-blank `ntp_server_1..4` slots, in order) and `ntp_setting_preference` (ntp section). Key names were verified on Network 10.6.102. The model is the allowlist: the connectivity section also carries the mesh PSK, which never reaches the result. A missing section leaves its fields `None`; an NTP section with no servers gives `[]`. The tool description now names what it returns, and the tool return goes through `redact_sensitive_fields` like its siblings.

`SITESETTINGS_MUTABLE_FIELDS` stays empty, so the cross-layer symmetry gate is unaffected and the Strawberry type is unchanged in this PR (see notes).

## Type of change

- [x] `fix:` bug fix

## Scope

- [x] Network MCP server
- [x] Shared packages

## Checklist

- [x] I updated documentation for user-visible changes. (`apps/network/docs/tools.md` blurb, regenerated skill reference.)

## Testing

```
uv run --package unifi-core --all-extras pytest packages/unifi-core/tests -n auto   # 1863 passed
uv run --package unifi-network-mcp pytest apps/network/tests/unit -n auto          # 1073 passed
uv run --package unifi-api-server pytest apps/api/tests -n auto                    # 1370 passed
```

New tests: `packages/unifi-core/tests/network/models/test_system.py::TestSiteSettingsExtendedSections` (sections mapped; the model is the allowlist, mesh PSK absent; missing sections leave `None`; blank NTP slots skipped, order kept; an NTP section with no servers is `[]`; blank strings read as absent; flag coercion for bool, 0/1, "true"/"false", unrecognised → `None`; new fields are read-only), `apps/network/tests/unit/test_system_tools.py` (tool returns the extended keys).

## Notes for reviewers

- The GraphQL/REST `SiteSettings` type (`apps/api/src/unifi_api/graphql/types/network/system.py`) re-implements the section walk and still projects the original four fields, so the same tool returns nine fields over MCP and four over the API. No gate catches that drift. The clean follow-up is to have it delegate to `site_settings_from_controller` the way the events and firewall types delegate to their `unifi_core` converters; I can open that as a separate PR if you want the API layer to follow.

### Deferred items

- API `SiteSettings` parity, as above.

### Considered & declined

- Making any of these fields mutable: they are read-only site state here; a writable timezone or NTP tool would follow the update-tool golden path as its own feature.


sirkirby (commented):
Tyler, I found no blocking issue in the MCP projection. The model/tool tests pass, and a read-only controller check confirmed the returned timezone matches the controller, all three settings sections are present, and the NTP servers reach the result. The allowlisted projection also avoids returning unrelated raw settings.

I’m treating the deferred REST/GraphQL fields as a separate API follow-up. Before release, publish the containing Core version and align consumer dependency floors so installed users receive the corrected projection; final CI and integration checks still apply.

Reviewed commit: `af32cd97f7bfbf5e99d05ec38f359e099321ab97`.

sirkirby (commented):
Tyler, I accept this read-only MCP projection as the fix for the site-settings mismatch. Timezone, connectivity and NTP are useful controller context. API read parity remains a separate follow-up; writable settings are outside this change.

I completed the missing targeted read smoke on `53f6eeec` against Network 10.6.102. The timezone, connectivity enabled/uplink values, all four NTP servers in order and NTP preference matched their controller sections. The output contained only the nine public fields, so the mesh PSK and unrelated settings were absent. There were zero controller mutations. This closes the field-population smoke gap; no repeated controller run is needed for these observations. Core delivery and final integration checks remain separate.

Acceptance follow-up at `53f6eeecd065ecbd2956684753eff0d82b26eb37`.

sirkirby (approved):
Tyler, I approve this revision. I compared the installed candidate with live Network 10.6.102: timezone, connectivity fields, four ordered NTP servers and preference matched the controller. Only the nine intended public fields were returned; this check made no mutations. False values and private-field exclusion also passed the focused probe. Full repository and installed Network safe smoke passed. This accepts the existing MCP settings contract; the separate REST/GraphQL projection remains unchanged.

PR #645, opened 2026-09-05
feat(network): expose SNMPv3 in unifi_get/update_snmp_settings

### LINKED ISSUE #633: feat: expose SNMPv3 (enabledV3) in unifi_get/update_snmp_settings — enabled=false today leaves v3 listening
### Component

Network MCP server

### Problem or use case

The controller's `snmp` site setting carries two independent flags: `enabled` (SNMP v1/2c, community string) and `enabledV3` (SNMPv3, user + password). `unifi_update_snmp_settings` exposes only `enabled` and `community`, and `unifi_get_snmp_settings` returns `{enabled, community, port: null, version: null}`. So:

- an operator who calls `unifi_update_snmp_settings(enabled=false)` and reads the tool's own output will reasonably believe SNMP is off. SNMPv3 is still listening.
- nothing in the tool surface reveals that a second SNMP service exists.

Reproduce: `unifi_get_snmp_settings` on a site with v3 configured, then compare `GET /api/s/default/rest/setting` → `.data[] | select(.key=="snmp")`. `enabledV3` is in the controller record and absent from the tool output. Still true at `main` @ `5302363`: `apps/network/src/unifi_network_mcp/tools/system.py` has no `enabledV3` anywhere.

Note `version: null` in the getter is not the v1/2c-vs-v3 indicator it looks like; the controller does not populate it. The two booleans are the real state.

### Proposed solution

- Getter: surface `enabledV3` (as `enabled_v3`) and the v3 `username`.
- Setter: accept `enabled_v3`, `username`, and a write-only v3 password (redacted in previews, same as `community`).
- At minimum, document that the current tool governs v1/2c only; the naming implies it governs SNMP.

### Alternatives considered

A separate `unifi_*_snmpv3_settings` pair. Fewer surprises for existing callers, but two tools for one controller record.

### Implementation notes

`system.py` snmp tools + `system_manager`; the preview already passes `current_state=current` for snmp, so the diff will be visible. New fields need `make manifest`.

### Contribution

- [x] I am willing to help implement or test this change.

### Code of conduct

### Environment

### Preflight

tmowbrey:
## Summary

Closes #633.

The controller's `snmp` site setting carries two independent services: `enabled` (v1/v2c, `community`) and `enabledV3` (SNMPv3, `username`, `x_password`). The tools exposed only the first, so an operator who called `unifi_update_snmp_settings(enabled=false)` and read the result believed SNMP was off while v3 kept listening, and nothing in the tool surface said a second service existed. Key names verified on Network 10.6.102; there are no auth/priv protocol fields (v3 is fixed SHA/AES-128), and the controller echoes `x_password` on GET.

- `SnmpSettings` gains mutable `enabled_v3`, `username`, `x_password`; `snmp_to_controller_update` now spells `enabled_v3` as `enabledV3` for the controller.
- `unifi_get_snmp_settings` returns the v3 state; `x_password` is redacted like `community`.
- `unifi_update_snmp_settings` gains `enabled_v3`, `username`, `x_password`; `enabled` becomes optional so a v3-only change need not restate v1 state. Only the fields passed are sent. The preview and the confirm echo use the model's key names so current and proposed line up.
- Error paths for both tools report the controller's `api.err.*` code or the exception class, never the message, and `SystemManager.update_settings` logs the same way: the settings document carries a password that a controller validation error can quote.
- API: the GraphQL/REST `SnmpSettings` type exposes the three fields (`xPassword` redacted) and reads the non-secret fields from the core converter; the action translator forwards them and rejects an all-omitted update, matching the MCP tool.
- Docs: tool blurbs, the three "what gets redacted" lists (root README, network README, plugin skill), regenerated manifest, action catalog, skill reference, GraphQL SDL and reference.

## Type of change

- [x] `feat:` new feature

## Scope

- [x] Network MCP server
- [x] API server
- [x] Shared packages
- [x] Documentation

## Checklist

## Testing

```
uv run --package unifi-core --all-extras pytest packages/unifi-core/tests -n auto   # 1854 passed
uv run --package unifi-network-mcp pytest apps/network/tests/unit -n auto          # 1087 passed
uv run --package unifi-api-server pytest apps/api/tests -n auto                    # 1378 passed (SDL/openapi/docs drift gates included)
```

New tests: model (v3 keys read, aliases written, `False` kept, read-only dropped), redaction vocabulary (`x_password` sensitive; `username`, `enabledV3` not), tool (getter surfaces v3 and redacts; policy-off returns raw; preview shows current v3 state and redacts both secrets; confirm sends controller keys and echoes model keys; v3-only, `enabled=False` alone, mixed v1+v3; empty rejected; controller rejection reported; controller error code surfaced; preview/confirm/getter failures leak nothing to the response or the log), manager (failure logs class only), API (serializer, GraphQL `enabledV3 username xPassword`, REST route, policy-off on both, translator forwards and rejects empty, action endpoint accepts v3 args without `enabled`).

## Notes for reviewers

- `enabled` leaving `required` loosens the action catalog's input schema for `unifi_update_snmp_settings`; the translator's empty-update guard keeps the two surfaces consistent.
- Partial writes to `/set/setting/snmp` preserve the fields not sent: the pre-existing tool already omitted `community` on an enable/disable, and a measured partial write on 10.6 changed one field and left the v3 credentials in place. If you know a controller version that replaces the document instead, say so and I will switch the manager to fetch-merge-put.
- The description now says the community string and v3 password are "returned by the controller and always shown redacted by this tool" rather than write-only, which is what the controller does.
- `.agents/skills/claude-plugin-config-transport/SKILL.md` lists the redacted fields too; left for you since `.agents/` is maintainer-owned.

### Deferred items

- The GraphQL type still defaults `community` to `""` before redaction, so an unconfigured v1/v2c community reads as redacted rather than null (pre-existing).

### Considered & declined

- A separate `unifi_*_snmpv3_settings` pair: two tools for one controller record, and the issue's own alternative; the flat params on the existing pair keep the round-trip (getter keys are setter params).
- Moving the update tool to an `update_data` dict: it would break existing callers, the API translator and the catalog for five scalar fields.


sirkirby (changes requested):
Tyler, the SNMPv3 shape and independent enable flags look right, and the focused tests and generated artifacts pass. I’m requesting a correction to the credential error path before this ships.

I exercised the real ConnectionManager and SystemManager with a synthetic controller exception containing the submitted `x_password`. The MCP response was sanitized, but [ConnectionManager logs the raw exception first](https://github.com/sirkirby/unifi-mcp/blob/1453cb5cb4a98585f552264b1c3464fee77d1bd3/packages/unifi-core/src/unifi_core/network/managers/connection_manager.py#L777-L778). The original exception also survives the manager and reaches the API route’s [audit `detail=str(e)` sink](https://github.com/sirkirby/unifi-mcp/blob/1453cb5cb4a98585f552264b1c3464fee77d1bd3/apps/api/src/unifi_api/routes/actions.py#L287-L297). That can copy a password into logs or audit storage when a controller error quotes the request. The new class-only manager log does not close that chain.

Please sanitize before logging or propagating these credential-bearing failures, covering reads/previews as well as writes, and add regressions through the real manager/transport boundary that check both logs and audit detail. This overlaps the management-credential error path in #648 and should be fixed consistently.

The read-only SNMP check passed. Partial-write preservation still needs verification, followed by Core-first publication, consumer floor alignment and the corrected-head checks.

Reviewed commit: `1453cb5cb4a98585f552264b1c3464fee77d1bd3`.

sirkirby (commented):
Tyler, I accept extending the existing SNMP pair to expose the two independent services. #633 establishes the misleading disable/read workflow, and one pair for one controller record is a better fit than separate v3 tools. This feature decision is separate from the credential error-path review.

Your Network 10.6.102 partial-write observation is useful; I need its preservation and restoration assertions made explicit. After safe error handling is established, please provide a smoke on an owned test site, or equivalent sanitized evidence from your existing run: identify the tested revision/tool or API entry point, change one v3 field, read it back, verify all omitted v1/v3 settings and credentials stayed unchanged, then restore and verify the original state. Publish equality/pass-fail booleans only—no credentials, hashes or raw settings. Use disposable credentials if testing credential replacement, and do not change production monitoring credentials for this review. Our successful getter/redaction check does not prove partial-write preservation.

Acceptance follow-up at `206d085728791889d7c2674c16aa1d0c309a7efd`.

tmowbrey:
Tested head: `39060ee` (rebased onto #660 `9ff0624`, the shared credential error-path correction; `bc09e3c` = upstream `main`).

**Findings addressed**

- Credential error path (`connection_manager.py:777-778`, `system_manager.py:327-332`, `routes/actions.py:287-297`): fixed in #660, which this branch now sits on. This branch's own class-only manager log was dropped in favour of #660's in-place scrub; the `_error_detail` code-or-class reporting in `tools/system.py` stays. `apps/network/tests/unit/test_system_manager.py` now asserts the scrubbed message (`x_password=***REDACTED***`) instead of class-only.
- Regenerated after the rebase: manifest, action catalog, skill reference, `make check-generated` clean.

**Partial-write preservation smoke** — owned controller, Network Application 10.6.102, entry point `unifi_update_snmp_settings` (MCP tool → real `SystemManager` → real `ConnectionManager`), raw read-back via `SystemManager.get_settings("snmp")`. The site had both `enabled` and `enabledV3` false with a stored username/community/x_password, so the one v3 field changed was `username` (to a throwaway value, then restored); the stored password and community were never sent or changed. Booleans only:

| step | result |
|---|---|
| preview (`confirm=false`): `requires_confirmation` true, no controller write | pass |
| write one v3 field (`username`), `success` | pass |
| read-back: `username` equals the submitted value | pass |
| every omitted field unchanged (`_id`, `key`, `site_id`, `enabled`, `enabledV3`, `community`, `x_password`) | pass (0 changed) |
| `x_password` unchanged / `community` unchanged / `enabled` unchanged / `enabledV3` unchanged | pass / pass / pass / pass |
| restore original `username`, `success` | pass |
| final record equals baseline (all 8 keys) | pass (0 differing) |
| stored password or community in any tool response | false |
| stored password, community, or login password in 22 captured DEBUG log lines | false |

Suites on this head: Core 1916, Network unit 1101, API 1374 passed; ruff clean.

Floor alignment: the Core model/manager changes ship in this PR; the Network/API floor bump follows the #657 pattern once you cut Core, or I can split the Core half out if you prefer that for this batch (same question asked once in #658).

sirkirby (commented):
Tyler, I reviewed `39060eeb` and its unique changes above #660. The SNMP extension fits the existing getter/update pair. Partial updates preserve omitted fields, explicit false is retained, and MCP/API use the shared translation. I independently passed 518 focused Core/Network/API tests, including the API artifact drift gates. No new blocking functional defect found in this feature's own delta.

Your exact-head username-only write, omitted-field comparison and full restore record supplies the requested controller preservation evidence. No repeat credential rotation is needed for reassessment.

This remains held on the escaped-credential defect now formally reported on #660, which is an exact ancestor, plus coordinated Core publication and Network/API floors. Ordinary successful smoke does not exercise that error path. After correction, verify the installed combined tool/API contract.

Minor wording correction during integration: the getter says secrets are always redacted, but response policy permits raw values; use redacted by default and regenerate the affected descriptions.

sirkirby (approved):
Approving. I rebased this onto merged `main` myself rather than send it back — with #660 squash-merged, the shared half here was conflicting against its own merged content.

## Rebase

`9ff0624d` is dropped: it is #660, now on `main` as `3e6aface`. Replaying only the SNMPv3 commit with `--onto` left one real conflict, in `packages/unifi-core/tests/test_redaction.py`, where this branch appends the v3 sensitivity test at the same point `main` now carries the encoded-representation block. Both sides are kept. The diff falls from 29 files / +1249 to 22 / +529, which is this PR's actual scope.

Core 0.4.46 is published, so the credential-scrub dependency is real rather than pending.

## The claim I most wanted to check

The tool description promises "Pass only the fields you want to change — current values are automatically preserved", and `SystemManager.update_settings` does *not* fetch-merge-put: it reads the section only to lift `_id`, then PUTs what it was given. So preservation rests entirely on controller behaviour, and if it were wrong, a caller setting `x_password` alone would silently wipe the community string.

I verified it independently rather than take the run in your comment. On the reference controller I captured the full `snmp` record, wrote **only** `enabled` at its existing value — a semantic no-op — and re-read:

```
LOST:    []
CHANGED: []
```

`community` survived a write that never mentioned it. The controller merges. Your reading is right, and no fetch-merge-put change is needed here.

One limit worth recording: that site has `enabledV3: false` with no stored `username` or `x_password`, so I confirmed preservation of the v1/v2c siblings, not a populated v3 credential round-trip. Your 10.6.102 measurement covers the half mine cannot.

## Redaction

Exercised through the real tool against the controller, comparing against the raw record:

```
snmp_settings: {"enabled": true, "community": "***REDACTED***", "enabled_v3": false,
                "username": null, "x_password": null, ...}
  real community value leaked : False
  preview leaks real community: False
```

Both the getter and the preview's `current`/`proposed` are clean. `_error_detail` is the right instinct — reporting `api.err.*` or the exception class and never the message, because the settings document carries the secret a validation error would quote. It complements #660's scrub rather than duplicating it.

## Gates

Ruff format and lint clean repo-wide; no generated-artifact drift; no f-string loggers; no issue references reintroduced (the repo was swept to zero earlier today, and this branch keeps it there). All five new model fields sit under `# --- mutable ---` with `None` defaults, so `SNMPSETTINGS_MUTABLE_FIELDS` picks them up and Gate 4 is satisfied. Docs, skill reference, manifest, action catalog and SDL all move together.

Suites: core 2228, Network 1144, API 1420.

## Release note for whoever ships this

This changes `packages/unifi-core`, and the failure mode if the floor is missed is **silent**, not an import error. `snmp_to_controller_update` already exists in 0.4.46, so nothing crashes — but the three new fields do not, so `SNMPSETTINGS_MUTABLE_FIELDS` filters them out and the tool accepts `enabled_v3` and quietly drops it. Verified against the published 0.4.46 wheel:

```
0.4.46 snmp_to_controller_update exists: True
0.4.46 SnmpSettings has enabled_v3:      False
```

So the sequence after merge is `core/v0.4.47` first, then the Network and API floors to `>=0.4.47`, then those two releases. Not optional.

Issue #216, opened 2026-05-08
feat(traffic-rules): expose UniFi Traffic Rules + DNS-redirect via MCP


sirkirby:
## Background

PR #215 (closing #210) surfaced a real capability gap in the project's MCP tool surface: **UniFi's port-based filtering and DNS-redirect features are not exposed through MCP**. This is distinct from "traffic routes" (which is policy-based VPN routing — already covered via `unifi_list_traffic_routes` / `unifi_update_traffic_route`).

The gap shows up most clearly in the firewall-auditor skill's EGR-02 benchmark (DNS forced through approved resolvers). V2 zone-based firewall policies cannot match on port, so the benchmark's full security intent ("block UDP/TCP port 53 to non-approved resolvers") cannot be implemented at the firewall layer alone. PR #215 had to split EGR-02 into:

- **Part 1 (programmatic)**: a firewall ALLOW rule for approved resolver IPs (handles the legitimate-DNS path).
- **Part 2 (manual UniFi UI)**: traffic rule OR DNS-redirect to actually block port-53 to non-approved resolvers.

EGR-02 is just the trigger — any benchmark or user request that involves port-based filtering, application-category filtering, time-based traffic shaping, or forced DNS resolution will hit the same wall.

## What this issue tracks

Add MCP tool coverage for the two UniFi features that fill the gap:

### A. Traffic Rules (UniFi UI: Settings → Security → Traffic Rules)

UniFi traffic rules support:
- Port-based filtering (TCP/UDP, single ports + ranges)
- Application/category-based filtering (DPI categories — distinct from V2 firewall's IP/network targeting)
- Time-based scheduling
- Source matching by network, IP/IP-group, MAC, country, and \"All Devices\"
- Action: block / allow

This is the primary missing piece. EGR-02 needs port-53 blocking; many other use cases need port/app filtering.

### B. DNS Filtering / DNS Redirect (UniFi UI: Settings → Security → DNS Filtering, plus per-network DNS overrides)

Two related capabilities:
1. **DNS Filtering** (the controller's built-in): Family / Work / Adult content category presets applied per-network.
2. **DNS Redirect**: forcing client DNS queries to a specific resolver IP regardless of the client's configured DNS.

The DNS Records tooling we already have (`unifi_create_dns_record` / `unifi_update_dns_record`) is for hostname overrides, not client-DNS interception — different feature.

## Why MCP coverage matters

[...]

PR #646, opened 2026-09-05
feat(network): validate port and client targeting on firewall policies

tmowbrey:
Zone-based firewall policies can target clients and specific ports or port groups, but the tools lacked consistent validation and read-back of those selectors. This change validates activating enums and required selectors, canonicalizes client MACs, and includes active targeting fields and inversion flags in policy summaries.

Partial updates share Core preparation across MCP and API execution: merge against stored state, retire obsolete selectors, preserve untouched fields, and reject invalid targeting before PUT. MCP previews use the same preparation. API previews reject contradictions visible in the supplied update; validation requiring stored policy state occurs at execution. Policy ordering remains a separate API family.

Independent live testing found two additional port-inversion defects and this revision fixes both: summaries retain inversion for OBJECT port groups, and switching to ANY automatically clears stored port inversion. Explicit ANY plus inversion is rejected locally, matching the controller's observed rejection.

The unproven DNS deployment recipes are excluded. The remaining docs describe selector syntax and read-back behavior; this PR does not claim proof of end-to-end DNS enforcement. It addresses the firewall selector portion of #216, not the full issue.

Validation:

- Focused Core selector/read-view tests, including inversion transition regressions.
- Independently installed Core and Network wheels exercised a disposable disabled policy and port group against the controller: create and read CLIENT/SPECIFIC selectors, transition to OBJECT with inversion, preserve the untouched source, retire client/port selectors by switching to ANY, and delete both test resources. All mutations followed preview then confirmation. Existing policies matched the initial full read-back after cleanup.
- The original inversion failure was reproduced, corrected, and the same lifecycle passed afterward. This proves storage/transition behavior; it does not test packet enforcement.
- Full `make pre-commit` and all 14 GitHub checks passed on 2f57eef9d2b0be360298b695ed1c206ecd2a1cd7.

Release Core first, then raise Network/API Core floors before app publication. The installed smoke used a locally built Core wheel, not an already published Core version.

### DIFF
--- a/apps/api/src/unifi_api/action_catalog.json
+++ b/apps/api/src/unifi_api/action_catalog.json
@@ -2526,7 +2526,7 @@
       "input_schema": {
         "properties": {
           "policy_data": {
-            "description": "V2 zone-based firewall policy dict. Required: name, action (ALLOW/BLOCK/REJECT), source (object with zone_id, matching_target), destination (same structure). For IP targeting: matching_target='IP', matching_target_type='SPECIFIC', ips=[...]. For network targeting: matching_target='NETWORK', matching_target_type='OBJECT', network_ids=[...]. For any in zone: matching_target='ANY'. For port targeting: port_matching_type='SPECIFIC', port='445' (a single string, not a 'ports' array). Optional: enabled, description, protocol, connection_state_type, connection_states, ip_version, schedule, logging.",
+            "description": "V2 zone-based firewall policy dict. Required: name, action (ALLOW/BLOCK/REJECT), source (object with zone_id, matching_target), destination (same structure). For IP targeting: matching_target='IP', matching_target_type='SPECIFIC', ips=[...]. For network targeting: matching_target='NETWORK', matching_target_type='OBJECT', network_ids=[...]. For any in zone: matching_target='ANY'. For clients: matching_target='CLIENT', client_macs=[...]. Ports on either side: port_matching_type='SPECIFIC' + port='53,853' (a single string, not a 'ports' array) or 'OBJECT' + port_group_id. Optional: enabled, description, protocol, connection_state_type, connection_states, ip_version, schedule, logging.",
             "type": "object"
           }
         },
@@ -6265,7 +6265,7 @@
             "type": "string"
           },
           "update_data": {
-            "description": "Dictionary of V2 zone-based fields to update: name, action (ALLOW/BLOCK/REJECT), enabled, source, destination, protocol, ip_version, logging, connection_state_type, connection_states, schedule. index is not accepted here; use unifi_reorder_firewall_policies to change order.",
+            "description": "Dictionary of V2 zone-based fields to update: name, action (ALLOW/BLOCK/REJECT), enabled, source, destination, protocol, ip_version, logging, connection_state_type, connection_states, schedule. Partial source/destination dicts are merged with the current policy; port_matching_type='SPECIFIC' needs port, 'OBJECT' needs port_group_id, and switching to another port_matching_type (or a side off CLIENT) retires the stored port/port_group_id/client_macs automatically. index is not accepted here; use unifi_reorder_firewall_policies to change order.",
             "type": "object"
           }
         },
--- a/apps/api/src/unifi_api/services/dispatch_overrides.py
+++ b/apps/api/src/unifi_api/services/dispatch_overrides.py
@@ -678,11 +678,24 @@ def _translate_list_clients(args: dict[str, Any]) -> tuple[tuple[Any, ...], dict
 
 
 def _translate_create_firewall_policy(args: dict[str, Any]) -> tuple[tuple[Any, ...], dict[str, Any]]:
-    """Apply shared port validation before the API preview/execute split."""
-    from unifi_core.network.models.firewall import validate_policy_port_targeting
+    """Apply shared port and selector validation before the API preview/execute split.
 
-    policy_data = args["policy_data"]
+    Both checks also run in ``FirewallManager.create_firewall_policy``; running them
+    here as well is what gives the API preview the same verdict as its execute, since
+    neither needs stored controller state.
+    """
+    from unifi_core.network.models.firewall import (
+        normalize_policy_endpoint_enums,
+        validate_policy_port_targeting,
+        validate_policy_selectors,
+        validate_zone_targeting,
+    )
+
+    policy_data = normalize_policy_endpoint_enums(args["policy_data"])
+    if zone_error := validate_zone_targeting(policy_data):
+        raise ValueError(zone_error)
     validate_policy_port_targeting(policy_data)
+    validate_policy_selectors(policy_data)
     return (), {"policy_data": policy_data}
 
 
--- a/apps/api/tests/test_actions_dispatcher.py
+++ b/apps/api/tests/test_actions_dispatcher.py
@@ -59,6 +59,105 @@ async def test_create_firewall_policy_rejects_invalid_ports_before_preview_or_ma
 @pytest.mark.asyncio
 @pytest.mark.parametrize("confirm", [False, True])
 @pytest.mark.parametrize("direction", ["source", "destination"])
+@pytest.mark.parametrize(
+    ("endpoint", "expected"),
+    [
+        ({"matching_target": "CLIENT"}, "client_macs"),
+        ({"matching_target": "CLIENT", "client_macs": ["nope"]}, "client_macs"),
+        ({"port_matching_type": "OBJECT"}, "port_group_id"),
+        ({"port_matching_type": "ANY", "port": "53"}, "port_matching_type"),
+        ({"matching_target": "ANY", "client_macs": ["aa:bb:cc:dd:ee:ff"]}, "matching_target"),
+    ],
+)
+async def test_create_firewall_policy_rejects_inactive_selectors_before_preview_or_manager(
+    confirm, direction, endpoint, expected
+):
+    """The Core create path rejects these at execution; the API preview must refuse them too,
+    since no stored state is needed to judge them."""
+    factory = MagicMock()
+    factory.get_domain_manager = AsyncMock()
+    data = {"name": "Inactive selector", "action": "ALLOW", direction: endpoint}
+    with pytest.raises(ValueError, match=expected):
+        await dispatch_action(
+            registry=PRODUCTION_REGISTRY,
+            factory=factory,
+            session=MagicMock(),
+            tool_name="unifi_create_firewall_policy",
+            controller_id="cid",
+            controller_products=["network"],
+            site="default",
+            args={"policy_data": data},
+            confirm=confirm,
+        )
+    factory.get_domain_manager.assert_not_awaited()
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("confirm", [False, True])
+@pytest.mark.parametrize(
+    ("endpoint", "expected"),
+    [
+        ({"matching_target": "any", "port_matching_type": "specific"}, "port"),
+        ({"matching_target": "any", "port_matching_type": "object"}, "port_group_id"),
+        ({"matching_target": "client", "client_macs": []}, "client_macs"),
+    ],
+)
+async def test_create_firewall_policy_validates_lower_case_endpoint_enums(confirm, endpoint, expected):
+    """The MCP tool upper-cases endpoint enums before validating. Until the API create
+    translator did the same, a lower-case enum turned every one of these checks into a
+    no-op on this surface only."""
+    factory = MagicMock()
+    factory.get_domain_manager = AsyncMock()
+    data = {"name": "Lower case", "action": "ALLOW", "source": {"zone_id": "z1", **endpoint}}
+    with pytest.raises(ValueError, match=expected):
+        await dispatch_action(
+            registry=PRODUCTION_REGISTRY,
+            factory=factory,
+            session=MagicMock(),
+            tool_name="unifi_create_firewall_policy",
+            controller_id="cid",
+            controller_products=["network"],
+            site="default",
+            args={"policy_data": data},
+            confirm=confirm,
+        )
+    factory.get_domain_manager.assert_not_awaited()
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("confirm", [False, True])
+@pytest.mark.parametrize(
+    ("endpoint", "expected"),
+    [
+        ({"matching_target": "IP"}, "matching_target_type is required"),
+        ({"matching_target": "NETWORK", "matching_target_type": "OBJECT"}, "network_ids array is required"),
+        ({"matching_target": "IP", "matching_target_type": "OBJECT"}, "ip_group_id is required"),
+    ],
+)
+async def test_create_firewall_policy_rejects_an_incomplete_target(confirm, endpoint, expected):
+    """The completeness rules used to live in the MCP tool, so this surface accepted a
+    target with nothing to match — a BLOCK policy that stops blocking."""
+    factory = MagicMock()
+    factory.get_domain_manager = AsyncMock()
+    data = {"name": "Incomplete", "action": "ALLOW", "source": {"zone_id": "z1", **endpoint}}
+    with pytest.raises(ValueError, match=expected):
+        await dispatch_action(
+            registry=PRODUCTION_REGISTRY,
+            factory=factory,
+            session=MagicMock(),
+            tool_name="unifi_create_firewall_policy",
+            controller_id="cid",
+            controller_products=["network"],
+            site="default",
+            args={"policy_data": data},
+            confirm=confirm,
+        )
+    factory.get_domain_manager.assert_not_awaited()
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("confirm", [False, True])
+@pytest.mark.parametrize("direction", ["source", "destination"])
 async def test_create_firewall_policy_valid_port_preserves_payload(confirm, direction):
     import copy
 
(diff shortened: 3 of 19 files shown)


sirkirby (changes requested):
Tyler, I’m requesting that the new selector validation and retirement behavior also apply to API calls. The [MCP wrapper performs both steps](https://github.com/sirkirby/unifi-mcp/blob/a7c314eb36372c22bba14263a2efa6e140c370d7/apps/network/src/unifi_network_mcp/tools/firewall.py#L501-L505), while the [API translator goes straight through normalization to the manager](https://github.com/sirkirby/unifi-mcp/blob/a7c314eb36372c22bba14263a2efa6e140c370d7/apps/api/src/unifi_api/services/dispatch_overrides.py#L678-L685). The [generated API catalog now promises automatic retirement](https://github.com/sirkirby/unifi-mcp/blob/a7c314eb36372c22bba14263a2efa6e140c370d7/apps/api/src/unifi_api/action_catalog.json#L6246).

With production dispatch and a real FirewallManager, I reproduced two differences: updating `port` to `"53"` on an existing `port_matching_type: "ANY"` is rejected by MCP but submitted successfully by API; switching an existing `SPECIFIC`/`53` selector to `ANY` through API retains the stale port. These probes used a fake controller connection and establish the dispatch gap, not hardware enforcement.

Please put merged-state validation and retirement on a path shared by both surfaces, with API regressions for invalid selector-only updates and activation changes. The focused suites and generated checks pass; they currently miss these cases. Preserve #640’s index rejection when integrating, then verify controller preservation/removal semantics and the Core-first dependency floors before release.

Reviewed commit: `a7c314eb36372c22bba14263a2efa6e140c370d7`.

sirkirby (commented):
Tyler, I accept the observed ZBF selector work as the right scope for this part of #216. Your discovery supports extending the existing firewall tools; a new Traffic Rules domain is unnecessary for port filtering. DNS redirect remains separate NAT work, and I’m deferring any claim that the DNS enforcement recipes are validated until there is traffic evidence.

Your disabled-policy create/port-update/delete smoke is useful real-controller evidence. Please extend it on an owned disposable policy to cover selector retirement (SPECIFIC→ANY, SPECIFIC↔OBJECT, CLIENT→ANY where supported), actual controller read-back, preservation of omitted fields, and cleanup of the policy and any disposable group. Exercise the corrected MCP/API paths and record revision, Network version and pass/fail comparisons; a summary that hides an inactive selector does not prove removal. No private rule records or identifiers are needed.

For a validated DNS recipe, I also need an isolated-client smoke showing approved DNS succeeds, non-approved DNS is blocked, resolver upstream remains available, and cleanup restores the test setup, including IPv6 if claimed. If that isolated test is unavailable, we can keep the configuration feature and leave the recipe’s enforcement verification explicitly deferred. Do not test this by disrupting production DNS.

Acceptance follow-up at `04e99a0e146a4d6cdcab5e7e4d75596839a029f7`.

tmowbrey:
Corrected head: `bce9da9ee57203bedbde9dddb26d7de8cb22320d`. Stacked on #640 (its commits are included, rebased onto `bc09e3c`); #640's own head is unchanged at `6f6f075`.

**Finding → change**
- Shared path: `prepare_policy_update` in `packages/unifi-core/src/unifi_core/network/models/firewall.py` retires deactivated selectors (`retire_stale_selectors`) and validates each updated side as merged (`policy_update_targeting_error`), raising `ValueError`. `FirewallManager.update_firewall_policy` calls it before building the PUT body, so MCP, API and direct Core callers get the same behaviour; the MCP wrapper (`apps/network/src/unifi_network_mcp/tools/firewall.py`) calls the same function for its preview instead of its own two steps.
- `normalize_policy_update` now also rejects a selector paired with a non-activating enum in the same update (needs no stored state, so it guards API previews too). `_selector_activator_errors` is the one table walk both checks use.
- #640's `index` rejection is preserved on that boundary.
- The catalog's retirement wording now holds for the API path: the manager does the retirement.
- Regressions (all red before the change): Core composition cases; `apps/network/tests/unit/test_firewall_manager.py::TestUpdateFirewallPolicySelectorPath` (activation change drops the stored `port` from the PUT body; `port` on a stored `ANY` raises before any PUT); `apps/api/tests/test_actions_dispatcher.py` through the production catalog with a real `FirewallManager` over a fake connection: selector-only update on an inactive enum rejected with no PUT, SPECIFIC→ANY sends no stale port, `{ANY, port}` contradiction rejected on preview and confirm before any manager is resolved. Suites: Core 1957, Network 1120, API dispatcher 303, catalog 18.
- One design limit to state plainly: an API *preview* of a selector-only update on a stored inactive enum still returns a preview, because API previews are built without a manager; execution rejects it. The stateless contradiction case is rejected at preview.

**Controller evidence** — Network 10.6.102, tested SHA above. Owned disposable DISABLED BLOCK policy and a disposable port group, both created and deleted in the run. Every read-back is the raw controller document from a fresh GET (cache invalidated), not the list summary. Request trace counts HTTP methods on `ConnectionManager.request`.

| Transition | Surface | Read-back |
|---|---|---|
| SPECIFIC/53 → ANY | MCP confirmed | 1 PUT; `port_matching_type` ANY, `port` key absent; other fields unchanged |
| `port` on stored ANY | API confirmed | rejected by the manager, 0 requests, document equal |
| `port` on stored ANY | MCP preview | rejected |
| `{port_matching_type: ANY, port}` | API preview + confirmed | rejected before manager resolution, 0 requests |
| ANY → SPECIFIC/853 | API confirmed | 1 PUT; SPECIFIC/853; untouched keys unchanged |
| SPECIFIC → OBJECT (disposable group) | MCP confirmed | 1 PUT; OBJECT + group id, `port` key absent |
| OBJECT → SPECIFIC/53 | API confirmed | 1 PUT; SPECIFIC/53, `port_group_id` key absent |
| source ANY → CLIENT (synthetic locally-administered MAC) | MCP confirmed | controller stored CLIENT with `client_macs` |
| source CLIENT → ANY | API confirmed | 1 PUT; ANY, `client_macs` key absent; other fields unchanged |
| `{"index": 2000, "enabled": true}` | API + MCP confirmed | rejected, 0 requests, document equal |
| Cleanup | MCP | policy and group deleted, both absent on fresh reads |

DNS recipe: the isolated-client smoke is still owed; until it is posted, treat the recipe's enforcement claim as deferred, per your acceptance note.

sirkirby (commented):
Tyler, I re-reviewed `bce9da9e`. The original shared-path finding is resolved: `FirewallManager` now applies selector retirement and merged-state validation before PUT, with the MCP preview using the same preparation. Shared normalization rejects stateless contradictions on API preview/confirm, and #640's index guard remains intact. Independent focused checks passed: 133 Core/read-view, 143 Network and 382 API/catalog/symmetry tests.

Your disabled-policy/port-group transition record is adequate targeted evidence for selector storage, omitted-field preservation and cleanup. The API's state-free preview limitation is explicitly bounded: selector-only invalidity requiring stored state is rejected at execution; this does not establish full preview parity.

One acceptance gate remains: the DNS recipes still ship unqualified enforcement/safety claims, but your live record does not test traffic and your comment correctly says that proof is owed. Before this whole PR advances, supply the isolated-client DNS test with resolver upstream continuity, or defer/split those recipe changes while retaining the validated selector work. Do not change enabled production policies merely to satisfy this check. Also retain the coordinated Core/app-floor gate; workspace tests do not close it.

tmowbrey:
Tested SHA: `4085f06d` (rebased onto `1ed216cc`; the five previous commits are gone).

## The scope changed under this PR

#662 and #630 merged after your last review, so `validate_policy_port_targeting` now lives in Core with three callers, and `_validate_zone_targeting` lives in `apps/network/.../tools/firewall.py`. The version you reviewed replaced both with its own `validate_policy_targeting`. This revision **builds on them instead** and is narrowed to what main does not already have.

| | main today | this PR |
|---|---|---|
| SPECIFIC port shape on create | `validate_policy_port_targeting` | unchanged, reused |
| zone / IP / NETWORK requirements | `_validate_zone_targeting` (MCP create) | unchanged, reused |
| CLIENT targeting (`client_macs`) | — | added |
| port `OBJECT` / `port_group_id` | — | added |
| selector paired with a non-activating enum | — | added |
| any validation on update | none | added, merged-state |
| zone / IP / NETWORK on update and API create | create + MCP only | now all three surfaces |
| selector retirement on update | — | added |

`validate_policy_targeting` is gone. The new Core surface is `validate_policy_selectors`, `retire_stale_selectors`, `policy_update_targeting_error` and `prepare_policy_update`, plus `SELECTOR_ACTIVATORS` / `RETIRABLE_SELECTORS` / `activator_is_known`.

## Against your remaining gate

> the DNS recipes still ship unqualified enforcement/safety claims … supply the isolated-client DNS test …, or defer/split those recipe changes

Split. The recipes are lifted onto `docs/firewall-dns-egress-recipes`, based on this branch, and are not in this PR:

- `policy-templates.yaml` / `policy-templates.md` — Templates 7 and 8 (`dns-egress-lock-external`, `dns-egress-lock-local`)
- `firewall-manager/SKILL.md` — the two index rows
- `firewall-schema.md` — the "Worked Example — DNS Egress Lock (EGR-02)" section
- `firewall-auditor/.../security-benchmarks.md` — the EGR-02 procedure and recipe rewrite

I did **not** run an isolated-client DNS smoke. Production DNS was not touched.

What stayed here is the validated selector work: the port-selector schema table, the `matching_target` enum list, and the "Client (MAC) Targeting" section. Two sentences in `security-benchmarks.md` had become false and I corrected them rather than leaving the docs contradicting the code — please pull them if you would rather they travelled with the recipes:

- SEG-03's "V2 firewall policies cannot match by client MAC" → CLIENT targeting, with `unifi_create_acl_rule` still named for switch-level enforcement (`security-benchmarks.md:157`).
- EGR-02's "V2 zone-based firewall policies do not match on port, so … isn't expressible at this layer" → they do; and the sentence now says explicitly that a full recipe is **not published here** because it has not been verified against live traffic (`security-benchmarks.md:270`). No enforcement claim is made.

> retain the coordinated Core/app-floor gate; workspace tests do not close it

No new Core or Shared symbol is imported by an app. Everything added is in `unifi-core`, and the two app-side call sites (`tools/firewall.py`, `dispatch_overrides.py`) already import from `unifi_core.network.models.firewall`. Core and the apps must still be released together for the behaviour to appear, and the floor stays yours to set — I have not touched a `pyproject.toml`.

## The zone/IP/NETWORK rules now apply everywhere

An earlier cut of this revision left them where you put them, in `_validate_zone_targeting`, called only by `unifi_create_firewall_policy`. That meant `unifi_update_firewall_policy` accepted `{"source": {"matching_target": "IP"}}` with no `ips` — an IP-targeted side with nothing to match, which is a BLOCK policy that stops blocking — and the API create path accepted it too. Main accepts both as well, so it was not a regression, but this is the PR that adds update validation and it was adding it for CLIENT and ports while stepping over the one next to them.

The rules moved to `unifi_core` as `zone_target_error` / `validate_zone_targeting`. **`_validate_zone_targeting` keeps its name and its single call site** — it is now a delegate — so the create tool reads as before. The Core function is called from the merged-update path, the API create translator and `FirewallManager.create_firewall_policy`, so all three surfaces reach the same verdict.

Suppression treats the `matching_target` family as one owner: a gap the controller already had does not block an unrelated update, but an update that writes any of `matching_target`, `matching_target_type`, `ips`, `ip_group_id` or `network_ids` owns the result. That distinction has its own tests — the first set I wrote did not actually reach it.

## Findings fixed in this revision

Two review rounds, six reviewers each. P0/P1 fixed and re-reviewed; the second round was run against the fix pass specifically.

**Privacy (P1).** `_client_macs_error` interpolated the caller's MAC list with `%r`. On the API surface that message is not returned to the caller — it lands in `AuditLog.detail` via `routes/actions.py`, which scrubs secret-*keyed* values only, and a MAC is not one. Fixed at the root rather than at the sinks:
- `models/firewall.py` — the message names the failing index (`source.client_macs[0] is not a MAC address; …`); no validation message in this change interpolates any caller value. `normalize_policy_enums`'s `Invalid action '<value>'` is included in that (`firewall.py:818`, pre-existing).
- `managers/firewall_manager.py` — the create path no longer dumps the payload at DEBUG, no longer logs the raw controller response, and no longer quotes it into the `RuntimeError` it raises.
- `tools/firewall.py` — the post-update mismatch warning logs the field name, not the expected/actual values.
- `managers/connection_manager.py` + `mac.py` — **the leak survived all of the above.** `_scrub_error` rewrites a controller rejection to remove credentials, but the rejection echoes the record it refused, and a CLIENT-targeted policy puts MACs in that record. New `mask_exception_macs` masks MAC tokens in an exception's args, message attributes and `__notes__`, following the cause chain, and `_scrub_error` calls it. This brings the re-raised exception in line with the log line, which `_sanitize_text` already masked.

**Correctness (P1).**
- `firewall_manager.py:410` — the retirement null-strip dropped **every** `None`-valued key from both merged endpoints on every update. `{"source": {"zone_id": null}}` silently lost the policy's zone binding instead of being rejected by the controller. Now it removes only the keys this update retired.
- `models/firewall.py::policy_update_targeting_error` — suppressed a pre-existing error by **message-string equality**, and two of the three selector messages are value-independent, so writing a new bad value into an already-bad selector was accepted: stored `port_matching_type: ANY` + `port: "445"`, update `{"port": "80"}` → accepted and PUT. Suppression is now per selector key, and excused only when the update touched none of the keys that error is about.
- Activation was compared case-sensitively while the MCP tool normalized first and the API create translator did not, so the two surfaces reached opposite verdicts on the same policy (`{"matching_target": "client", "client_macs": [...]}` accepted by MCP, rejected by the API). `_activator_matches` is now case- and space-insensitive, and `normalize_policy_endpoint_enums` runs in the API create translator, in `FirewallManager.create_firewall_policy` and in `prepare_policy_update` — `validate_policy_port_targeting` compares against the controller's exact spelling, so normalization is what keeps the two halves agreeing.
- `tools/firewall.py::_selector_applied` — the post-PUT verification compared a retired `port: None` against the controller's re-read. A controller that answers with the key absent or an empty default made a successful retirement report `"did not apply changes to: destination"`. A retired key is now satisfied by absent-or-empty; a surviving value is still reported.

**Reporting (P2).**
- `read_views.shape_firewall_policy_list` — a selector is hidden only when its enum value is one this project knows and is not the activating one; an unrecognised target (App, Web, Region, …) keeps its selectors, because reporting a narrow rule as zone-wide is the worse error on the surface the firewall auditor reads. The IP and NETWORK selectors (`ips`, `ip_group_id`, `network_ids`, `match_opposite_*`) are gated the same way for display.
- `models/firewall.py` — a leftover selector is now named as the thing to remove ("`destination.port_group_id` is ignored while `destination.port_matching_type` is 'SPECIFIC'. Remove it, or set …"). The old wording told a caller who had just chosen SPECIFIC to switch back to OBJECT.
- `firewall_manager.py` — a `prepare_policy_update` rejection is caller input, not a controller failure; it no longer emits an operator-facing ERROR with a traceback. `create_firewall_policy` validates outside its `try` for the same reason.
- `_normalize_endpoint_macs` sends `canonical_mac`'s colon form: validation accepts the dashed and bare-hex spellings, and the controller reports colons.

## Entry points exercised

`unifi_create_firewall_policy` and `unifi_update_firewall_policy` at `confirm=False` and `confirm=True`; `unifi_list_firewall_policies` at `summary=true` and `summary=false`; the API dispatcher for both mutations at both confirm values; `FirewallManager.create_firewall_policy` / `update_firewall_policy` directly. No live-controller run against this SHA — the disabled-policy create/port-update/delete record stands against `bce9da9e`, and the shapes it covered are unchanged, but a fresh run is owed.

## Checks

| | |
|---|---|
| `packages/unifi-core/tests` | **2369 passed** |
| `apps/network/tests` | **1194 passed** |
| `apps/api/tests` | **1463 passed** |
| `apps/protect/tests` / `apps/access/tests` | **514** / **223 passed** |
| `make shared-test` | **477 passed** |
| `make check-generated` | **pass** (no drift) |
| `ruff format` / `ruff check` | **pass** |

Run separately — `apps/` and `packages/` paths collide in one pytest invocation.

Every fix above is pinned by a test that fails when the production check is removed; I ran that mutation for each one rather than assuming it.

### Deferred items

- **Port-string format.** The earlier revision validated `port` as a comma list of 1-65535 with `low-high` ranges. Dropped: `validate_policy_port_targeting` owns the SPECIFIC port shape now, and two functions validating one field to different depths on different paths is the divergence this revision exists to avoid. `"53, 853"` (with a space) and `"70000"` still reach the controller.
- **`ips` / `ip_group_id` / `network_ids` are not activation-checked or retired.** They follow the same contract on the controller. `firewall-schema.md` now says so explicitly instead of claiming the tools reject every ignored pairing. Extending `SELECTOR_ACTIVATORS` to them needs a two-key activator (`matching_target` + `matching_target_type`) and changes retirement behaviour on a widely used path; it should be its own PR.
- **API update preview.** `dispatch_action` builds the preview before resolving a manager, so `prepare_policy_update` runs at execution only: the API preview does not show a retirement and does not carry the stored-state verdict. This is the bound you described as explicit; there is now a test pinning it so it cannot change unnoticed.
- **`unifi_core/diagnostics.py:211-225`** logs the whole request payload at INFO when `UNIFI_MCP_DIAGNOSTICS=true`, with no MAC masking, so CLIENT targeting routes MACs there. Opt-in and off by default, and masking policy for diagnostics as a whole is yours to set — not changed here. Reproduction on request.
- `unifi_list_firewall_policies` still returns without the `redact_sensitive_fields` wrapper its siblings use (pre-existing; no firewall field is in the secret vocabulary).
- `description` is accepted on create but dropped on update by `to_controller_update` (pre-existing).

### Considered & declined

- **Replacing `_validate_zone_targeting` / `validate_policy_port_targeting`.** The reason for the whole revision; see the table above.
- **Refactoring the create tool's `_allowed_keys` frozenset to `MUTABLE_FIELDS | {"description"}`.** In the earlier revision. It is a pure refactor with no test behind the equivalence, and this PR is narrower without it.
- **Accepting a read-modify-write that echoes a stale selector.** `normalize_policy_update` rejects `{"source": {…, "port_matching_type": "ANY", "port": "53"}}` — the document the caller submitted contains the contradiction — while the same intent as a partial dict (`{"source": {"zone_id": "z9"}}`) is accepted. Asymmetric, and deliberately so: the stateless boundary is the only check the API preview gets, and the message is actionable either way. Pinned by a test so it stays a choice. Say the word and I will make it stateful-only.
- **Showing a selector whose activating enum key is absent.** A missing `port_matching_type` means the controller's default (`ANY`), so the selector is inert and stays hidden; an enum present with an unrecognised value is shown. One reviewer argued absent should also be shown. Every controller observed emits the key.

sirkirby (approved):
Tyler, I re-reviewed the selector implementation and independently exercised it against the controller with a disposable disabled policy and port group. The DNS deployment recipes are excluded, satisfying the remaining recipe hold without changing enabled policies.

The live run found two additional problems: OBJECT port inversion disappeared from summaries, and switching an inverted match to ANY failed with FirewallPolicyMatchOppositePortsUnsupportedForAnyPort. I fixed both on your branch and added regressions. The repeated lifecycle passed create/read, SPECIFIC-to-OBJECT transition, inversion read-back, omitted-field preservation, selector retirement, and deletion. Cleanup was verified and the existing policies matched their initial full read-back.

Full `make pre-commit` passed on 2f57eef9d2b0be360298b695ed1c206ecd2a1cd7, and all 14 current GitHub checks are green. This evidence covers selector storage and transitions, not packet-level DNS enforcement. I am merging this into the Core release batch; Network/API publication remains gated on the released Core floor and installed checks.

PR #647, opened 2026-09-05
fix(network): start the event websocket listener and report its state

tmowbrey:
Network's websocket event buffer was never started by the MCP server, and the listener subscribed only after awaiting the receive loop. The server now starts one runtime event manager, subscribes before receiving, and runs the socket in an owned background task with bounded reconnect backoff and session reauthentication.

The buffered-event and subscription tools report listener state, attachment, buffer occupancy/capacity, and a diagnostic when the stream is unavailable. Attachment requires a frame received on the current socket; resetting reconnect backoff also requires a stable attachment. Operators can disable the listener with `UNIFI_NETWORK_WEBSOCKET_ENABLED`. Buffer size and TTL are configurable and documented.

The API manager factory now owns listener startup and disposal, including managers created after credential rotation. Discarding a manager terminates its SSE streams so clients can reconnect; old callbacks and old stream cleanup cannot affect the replacement manager. All three MCP servers now close their controller sessions after normal shutdown or startup/transport failures. Failed or cancelled Protect initialization also closes each partially constructed SDK client before retrying. Healing a blocked API connection evicts only the managers attached to that connection, preserving healthy sibling streams. Websocket diagnostics expose exception classes/statuses rather than raw exception details.

Addresses the event-listener portion of #635. Other portions remain separate.

Validation:

- Focused Network lifecycle/tool tests: 83 passed; focused API factory/SSE/lifespan tests: 38 passed.
- Integrated with the published startup security fix; Shared >=0.6.17 remains required. Full `make pre-commit` passed; the final API suite passed all 1,470 tests, and all 14 GitHub checks passed on c567a6010ae42988213baf5c001595ec33dc275d.
- Independently built Core and app wheels, installed outside the checkout with published Shared 0.6.17, and exercised the actual MCP stdio startup against the controller. Enabled: listening and attached, normal controller read succeeded. Disabled: neither listening nor attached, normal controller read succeeded. Both shutdowns produced no unclosed-session warning.
- No event arrived during the live observation window. That run proves real attachment and cleanup; event ingestion, reconnect timing, cancellation, and credential-rotation stream delivery are covered by deterministic tests.

Installed-wheel safe matrix: Network 162 successful checks, Protect 59, Access 44; zero failures, all disposable resources cleaned up. Approval-required operations were not executed.

Publication requires releasing the new Core 0.4.49 code first and raising Network/API Core dependency floors before app releases. The source-wheel check used the Core wheel built from this branch; it is not evidence that the published Core package already contains these changes.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -111,6 +111,13 @@ UNIFI_MCP_LOG_LEVEL=INFO
 # ACCESS_EVENT_BUFFER_TTL=300
 # ACCESS_WEBSOCKET_ENABLED=true
 
+# ---------------------------------------------------------------------------
+# Network event streaming (feeds unifi_recent_events)
+# ---------------------------------------------------------------------------
+# UNIFI_NETWORK_EVENT_BUFFER_SIZE=100
+# UNIFI_NETWORK_EVENT_BUFFER_TTL=300
+# UNIFI_NETWORK_WEBSOCKET_ENABLED=true
+
 # ── Cloud relay (optional) ───────────────────────
 # UNIFI_RELAY_URL=https://your-worker.workers.dev
 # UNIFI_RELAY_TOKEN=your-relay-token
--- a/apps/access/src/unifi_access_mcp/main.py
+++ b/apps/access/src/unifi_access_mcp/main.py
@@ -58,72 +58,75 @@ async def main_async():
     check_unknown_policy_env_vars("access", logger, policy_gates(), ACCESS_CATEGORY_MAP)
     assert_credentials_configured(config, plugin_name="unifi-access", env_prefix="ACCESS", logger=logger)
 
-    # Initialize the global Access connection
-    logger.info("Initializing global Access connection from main_async...")
-    if not await connection_manager.initialize():
-        logger.error("Failed to connect to UniFi Access. Tool functionality may be impaired.")
-    else:
-        logger.info("Global Access connection initialized successfully from main_async.")
-
-        # Start the websocket event listener if enabled and connection succeeded
-        ws_enabled_raw = config.access.events.get("websocket_enabled", True) if hasattr(config, "access") else True
-        ws_enabled = parse_config_bool(ws_enabled_raw)
-        if ws_enabled:
-            try:
-                event_manager.set_server(server)
-                # start_listening needs an API client: the Access websocket is
-                # key-authenticated, and a proxy-only session cannot open it.
-                # It logs that condition itself and returns without raising, so
-                # a proxy-only deployment keeps working with REST queries and
-                # an empty recent-events buffer.
-                await event_manager.start_listening()
-            except Exception as ws_exc:
-                logger.error(
-                    "Failed to start event websocket listener: %s. "
-                    "Real-time events will be unavailable; REST queries still work.",
-                    ws_exc,
-                    exc_info=True,
-                )
-        else:
-            logger.info("Access event websocket disabled via config.")
-
-    # ---- Register MCP resources ----
     try:
-        import unifi_access_mcp.resources.events  # noqa: F401
+        # Initialize the global Access connection
+        logger.info("Initializing global Access connection from main_async...")
+        if not await connection_manager.initialize():
+            logger.error("Failed to connect to UniFi Access. Tool functionality may be impaired.")
+        else:
+            logger.info("Global Access connection initialized successfully from main_async.")
 
-        logger.info("MCP resources registered (events).")
-    except Exception as res_exc:
-        logger.error("Failed to register MCP resources: %s", res_exc, exc_info=True)
+            # Start the websocket event listener if enabled and connection succeeded
+            ws_enabled_raw = config.access.events.get("websocket_enabled", True) if hasattr(config, "access") else True
+            ws_enabled = parse_config_bool(ws_enabled_raw)
+            if ws_enabled:
+                try:
+                    event_manager.set_server(server)
+                    # start_listening needs an API client: the Access websocket is
+                    # key-authenticated, and a proxy-only session cannot open it.
+                    # It logs that condition itself and returns without raising, so
+                    # a proxy-only deployment keeps working with REST queries and
+                    # an empty recent-events buffer.
+                    await event_manager.start_listening()
+                except Exception as ws_exc:
+                    logger.error(
+                        "Failed to start event websocket listener: %s. "
+                        "Real-time events will be unavailable; REST queries still work.",
+                        ws_exc,
+                        exc_info=True,
+                    )
+            else:
+                logger.info("Access event websocket disabled via config.")
 
-    # ---- Register tools ----
-    await register_tools_for_mode(
-        mode=UNIFI_TOOL_REGISTRATION_MODE,
-        server=server,
-        original_tool_decorator=_original_tool_decorator,
-        tool_index_handler=tool_index_handler,
-        start_async_tool=start_async_tool,
-        get_job_status=get_job_status,
-        register_tool=register_tool,
-        tool_module_map=TOOL_MODULE_MAP,
-        setup_lazy_loading=setup_lazy_loading,
-        base_package="unifi_access_mcp.tools",
-        config=config,
-        logger=logger,
-        support_bundle_handler=support_bundle_service.generate,
-        prefix="access",
-        server_label="UniFi Access",
-    )
+        # ---- Register MCP resources ----
+        try:
+            import unifi_access_mcp.resources.events  # noqa: F401
 
-    # ---- Start transports ----
-    http_enabled, http_transport, host, port = resolve_http_config(config.server, default_port=3002, logger=logger)
-    await run_transports(
-        server=server,
-        http_enabled=http_enabled,
-        host=host,
-        port=port,
-        http_transport=http_transport,
-        logger=logger,
-    )
+            logger.info("MCP resources registered (events).")
+        except Exception as res_exc:
+            logger.error("Failed to register MCP resources: %s", res_exc, exc_info=True)
+
+        # ---- Register tools ----
+        await register_tools_for_mode(
+            mode=UNIFI_TOOL_REGISTRATION_MODE,
+            server=server,
+            original_tool_decorator=_original_tool_decorator,
+            tool_index_handler=tool_index_handler,
+            start_async_tool=start_async_tool,
+            get_job_status=get_job_status,
+            register_tool=register_tool,
+            tool_module_map=TOOL_MODULE_MAP,
+            setup_lazy_loading=setup_lazy_loading,
+            base_package="unifi_access_mcp.tools",
+            config=config,
+            logger=logger,
+            support_bundle_handler=support_bundle_service.generate,
+            prefix="access",
+            server_label="UniFi Access",
+        )
+
+        # ---- Start transports ----
+        http_enabled, http_transport, host, port = resolve_http_config(config.server, default_port=3002, logger=logger)
+        await run_transports(
+            server=server,
+            http_enabled=http_enabled,
+            host=host,
+            port=port,
+            http_transport=http_transport,
+            logger=logger,
+        )
+    finally:
+        await connection_manager.close()
 
 
 def main():
--- a/apps/access/tests/unit/test_server_cleanup.py
+++ b/apps/access/tests/unit/test_server_cleanup.py
@@ -1,0 +1,33 @@
+"""Server teardown releases the controller after normal and failed startup."""
+
+from unittest.mock import AsyncMock
+
+import pytest
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("failure", [None, "initialize", "registration", "transport"])
+async def test_server_always_closes_its_connection(monkeypatch, failure):
+    from unifi_access_mcp import main
+
+    initialize = AsyncMock(return_value=False)
+    registration = AsyncMock()
+    transport = AsyncMock()
+    close = AsyncMock()
+    stages = {"initialize": initialize, "registration": registration, "transport": transport}
+    if failure:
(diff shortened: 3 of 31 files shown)


sirkirby (changes requested):
Tyler, I reviewed `00fc1f2651acc4f820fbcd3a16c93eeb26c54c5f` and am requesting changes for three P2 lifecycle issues:

- **Report a completed websocket attachment.** [Attachment state](https://github.com/sirkirby/unifi-mcp/blob/00fc1f2651acc4f820fbcd3a16c93eeb26c54c5f/packages/unifi-core/src/unifi_core/network/managers/event_manager.py#L149-L156) is inferred from a running task and no last error. With the real aiounifi 95 websocket path held at its handshake, I observed `attached=True` before the socket opened. It also stays true during backoff after a normal closure. Please derive attachment from confirmed socket-open state and clear it on closure/backoff.
- **Allow reconnect after the auth cooldown expires.** The [retry guard](https://github.com/sirkirby/unifi-mcp/blob/00fc1f2651acc4f820fbcd3a16c93eeb26c54c5f/packages/unifi-core/src/unifi_core/network/managers/event_manager.py#L268-L270) uses `reconnect_blocked`, which remains latched until a successful connection. I expired the real ConnectionManager circuit and observed zero listener login attempts, although a direct `ensure_connected()` call could retry. Please let ConnectionManager enforce its time-aware circuit, or expose an active-cooldown check.
- **Unsubscribe when the stopping caller is cancelled.** In [shutdown](https://github.com/sirkirby/unifi-mcp/blob/00fc1f2651acc4f820fbcd3a16c93eeb26c54c5f/packages/unifi-core/src/unifi_core/network/managers/event_manager.py#L187-L207), caller cancellation is re-raised before subscription cleanup. I reproduced a retained subscription after the manager reported it had stopped. Please put cleanup in `finally` while preserving cancellation, and assert both propagation and unsubscribe.

The 109 focused tests passed; these separate, network-free probes expose paths their mocks/assertions miss. I have not established live event receipt and recovery for this revision.

sirkirby (commented):
Tyler, I accept repairing the existing recent-event feature: a bounded in-process buffer with accurate attachment state and an operator disable switch fits the project. I am not accepting an expansion into persistent monitoring, alerting or automated remediation.

Your earlier websocket-path and stale-cookie observations help establish the endpoint, but sync frames alone do not prove that this branch delivers client events through MCP. After the lifecycle fixes are re-reviewed, I need one targeted controller run on the candidate SHA: start an isolated MCP process, connect/disconnect an operator-owned test client, and show a timestamped EVT_* in the tool's buffer. Then close only that process's websocket, show the not-attached state, automatic recovery and receipt of a second event without duplicate subscriptions. Also record disabled-mode behavior and clean shutdown. If the API lifecycle remains part of this change, include equivalent receipt/recovery evidence through that surface.

Please record controller/SDK versions and transport/registration mode with sanitized outcomes. I will coordinate the maintainer run if you lack hardware. We should test auth cooldown with fault injection, without triggering controller lockouts or rotating production credentials. My earlier passing tests are separate from this live feature proof.

Acceptance follow-up at `4572bbdc58ba51508e2a3c8b2258a115e1af83ee`.

tmowbrey:
All three lifecycle findings addressed, then the live cycle on both surfaces. Tested SHA `9cc95672b02e92069fc7ad08acc8b49e22917953` (fix commit `c825ce6` + merge of upstream `main` `bc09e3c`).

**Fixes** (`packages/unifi-core/src/unifi_core/network/managers/event_manager.py`, `connection_manager.py`), each written as the network-free probe first (`apps/network/tests/unit/test_event_manager.py::TestWebsocketHealth`):

- **Attachment** (`event_manager.py` L149-156 → `attached`): aiounifi exposes no open/close callback but stamps `ws_message_received` on every frame, and the controller sends sync frames continuously. `attached` is now true only once that stamp has changed since the current `start_websocket` attempt began; a stamp left by an earlier socket does not count; it is cleared in a `finally` around `start_websocket` (returned or raised), so it stays false for the whole backoff. Probes: `test_attached_needs_a_confirmed_open_socket` (handshake held → `attached=False`, frame → `True`, closure → `False`), `test_a_frame_from_a_previous_socket_does_not_count`.
- **Cool-down** (`event_manager.py` L268-270): the loop now consults a new time-aware `ConnectionManager.reconnect_cooldown_active` (`_reconnect_block_active() is not None`) instead of the latched `reconnect_blocked`, so an expired cool-down gets its half-open login attempt. The pre-check is kept rather than dropped because `initialize()` logs the block at ERROR on every call. Probes: `test_loop_retries_once_the_auth_cooldown_has_expired` (latched flag + expired timer → `ensure_connected` awaited), `test_reconnect_cooldown_active_half_opens_on_the_timer` in Core (block → active; past `_reconnect_block_until` → inactive while `reconnect_blocked` stays true). Auth cool-down was exercised by this fault injection only; no controller lockout was triggered.
- **Cancelled stopper** (`event_manager.py` L187-207): `_unsubscribe()` (and the attachment reset) run in `finally`; a cancellation aimed at the caller still propagates. Probe: `test_stop_unsubscribes_even_when_the_caller_is_cancelled` asserts `CancelledError` propagation, `unsub` called once, `is_listening=False`.

Suites: Core 1896, Network unit 1131, API 1374, Protect 513 passed; ruff, `make check-generated`, `make docs-test` clean; no generated drift (no tool surface changed).

**Live cycle, MCP surface.** Isolated process: `unifi-network-mcp` from this checkout, stdio transport, registration mode lazy (the plugin default; calls through `unifi_execute`), Network Application **10.6.102** (self-hosted path), aiounifi 95, mcp 2.1.1. Test client: an operator-owned phone joining/leaving Wi-Fi. Socket drop: `ss -K` on that process's TCP connections to the controller only (nothing else on the host was touched); the first drop reconnected inside the 2 s poll interval, so the recorded drop is a 12 s outage during which every reconnect of that process was destroyed. Buffer polled every 2 s through `unifi_recent_events`; state lines from the server's own diagnostics log.

| Step | Result |
|---|---|
| start → `listening=true`, `attached=true`, `last_error=null` | true (10 s after start) |
| `EVT_WU_Disconnected` then `EVT_WU_Connected` for the test client in the buffer, timestamped (`time` set) | true / true |
| drop that process's websocket → `attached=false`, `hint` present, buffer intact | true (2 s after the drop; buffer stayed at 9) |
| during backoff: `last_error="ConnectionError (closed by the controller before it was stable)"` | true |
| automatic recovery → `attached=true`, `last_error=null` | true (16 s after the drop began, 5 s after the outage ended) |
| second test-client event received after recovery (`EVT_WU_Disconnected`) | true |
| duplicate subscriptions: duplicate event ids in the final buffer / buffer growth per event | 0 / +1 |
| disabled mode (`UNIFI_NETWORK_WEBSOCKET_ENABLED=false`, separate process): `listening=false`, `attached=false`, hint present, no listener started | true |
| clean shutdown (stdin closed): "websocket listener stopped" logged, process gone | true / true |

Final buffer: 10 events (LAN and wireless client events from the site, all with `time`), 0 duplicate ids.

**Live cycle, API surface** (the API lifecycle stays in the PR). Isolated process: `python -m unifi_api serve` from this checkout on a throwaway state DB, uvicorn 0.52.1 / fastapi 0.141.1, our controller registered before startup so the lifespan's eager `start_listening` picks it up. `GET /v1/streams/network/events` tailed for the whole run; `POST /v1/actions/unifi_recent_events` polled every 2 s.

| Step | Result |
|---|---|
| lifespan: "websocket listener started", SSE stream 200 | true / true |
| `EVT_WU_Disconnected` + `EVT_WU_Connected` for the test client on the SSE stream, and in the buffer action (count 2) | true / true |
| drop that process's websocket (12 s outage): listener logs closure and three backed-off attach failures (+1, +2, +4 s), then re-attaches | true |
| second pair of test-client events on the stream after recovery; buffer count 4 | true |
| duplicate SSE frames by event time | 0 |
| SIGTERM: "websocket listener stopped" logged | true |

The API's `unifi_recent_events` action returns the event list only (it binds straight to `get_recent_from_buffer`), so `attached`/`last_error` are not visible on that surface; the not-attached window is evidenced there by the listener log lines. If you want the MCP tool's listener-state fields on the action response too, I can add that as a small parity follow-up.

Scripts and sanitized outputs stay in my session scratchpad; no MAC, hostname or credential was written to any of the above.

sirkirby (changes requested):
Tyler, I re-reviewed `9cc95672`. The three earlier Network lifecycle corrections are present: attachment waits for a new frame, expired auth cooldown can retry, and a cancelled stopper releases its subscription. Your MCP/API reconnect and cleanup record addresses that targeted controller evidence. Independent focused Network/Core/API tests passed (83/9/22).

I am requesting a P2 correction in `apps/api/src/unifi_api/services/managers.py:530-532`. `_heal_blocked_connections` identifies only the blocked connections that a probe can heal, but removes every domain manager for the controller and now calls `_stop_domain_managers` on all of them. Recovering a Network connection therefore unsubscribes an otherwise healthy Protect event manager on the same console, even though its Protect connection is preserved.

I reproduced the consequence through the production Protect EventManager, SubscriberPool and SSE generator with synthetic SDK events: an existing stream receives an event before Network-only healing, then only keepalives afterward; the healthy Protect websocket has zero callbacks. Rebuilding the domain manager does not start its listener. This is a network-free regression probe, not an induced controller lockout. A control using byte-identical base factory/stream/Protect code retains the callback and delivers the next event. The new stop call terminates the previously live sibling subscription; broad cache eviction itself predates this change.

Please restrict healing-time domain eviction/stopping to managers bound to the connections being healed, preserving healthy products and Network sites. Add a regression with an active healthy sibling stream that still delivers afterward, while the healed connection's old listener is stopped. Full controller invalidation/shutdown should still stop every affected listener.

The Core/app floors also remain an acknowledged delivery gate. The new state properties and returning start-listening contract must be present in the installed Core before app delivery; keep that coordinated with the foundation release.

tmowbrey:
The P2 is fixed in `7482d83a`, and there is a formatting commit behind it that I should flag first.

**`dd0fafe7` is not mine to claim.** The pre-push lint gate rejected the branch on `apps/protect/src/unifi_protect_mcp/main.py:72`, a long line from your `a921f2bd`. The commit is `ruff format`'s own output on that one line, nothing else. Fold it into yours or drop it as you prefer.

## The fix

`_heal_blocked_connections` now evicts only the domain managers built on the connections actually being healed, rather than every manager sharing the controller id.

The projection is named rather than derived inline. `_connection_key_for` maps a domain key `(controller, product, attr, site)` to the connection key `(controller, product, site)` it was built on, and the sweep tests membership of that against the set of healable connection keys. That subsumes the `k[0] == controller_id` guard the previous line needed: because the controller id travels inside the projected key, a same-product, same-site scope on a *different* controller can no longer be swept, which the old predicate would have allowed if the guard were ever dropped.

`invalidate_controller` and shutdown are untouched and still stop every listener for the controller — the point you made about full invalidation.

## Regressions

Two, in `apps/api/tests/test_managers_factory.py`.

`test_probe_heal_keeps_a_healthy_sibling_stream_delivering` is the one you asked for: a `SubscriberPool` stream is attached to the Protect manager and receives an event, the Network connection is then latched and healed, and the stream delivers the next event afterwards. It also asserts the Protect subscription was never torn down, while the healed Network listener's `stop_listening` was awaited exactly once.

`test_probe_heal_leaves_a_healthy_products_manager_cached` adds the part the stream cannot observe: the Protect manager is still the *cached* instance, so it was not evicted and silently rebuilt.

I proved the red state before fixing, and the failure was the consequence you described — the healthy Protect listener losing its subscription. Then I mutation-tested the predicate three ways: reverting to the controller-wide sweep, making the projection return the attribute instead of the site, and emptying the healable set. All three fail the new tests, and the second and third also fail your existing `test_probe_heal_stops_listening_event_managers`, so the tests pin both directions — too broad and too narrow.

Suites, run separately: API 1470, Core 2393, Network 1250, Protect 521, all passing. `ruff check` clean, `make check-generated` reports no drift.

## Not in this commit

Five things I found while reconciling my local work against your two commits. I have deliberately left all of them out to keep this reviewable — say which you want and I will do them as separate commits.

1. `_stop_domain_managers` catches `except Exception`, which does not catch `CancelledError`. A cancellation while stopping one listener skips every listener after it in the list, and each one left running re-logs-in on its own timer.
2. `_heal_blocked_connections` closes the popped connections in a plain loop. A cancellation between stopping a listener and closing its session leaves an authenticated session that nothing owns.
3. The re-login after a rejected handshake has no rate cap. With backoff at 1, 2, 4, 8, 16, 32s, attempts land at t=0, 1, 3, 7, 15, 31 — six logins inside the first 31 seconds. The controller's limiter answers 429 at roughly six a minute and that latches the *shared* reconnect circuit, which denies every tool, not just the listener.
4. `main.py` starts the listener only inside the `else` branch of a successful `initialize()`. If the boot connect fails, the listener never starts for the life of the process and `unifi_recent_events` stays empty with `listening=false`, even after the controller comes back.
5. `int(cfg.get("buffer_size", 100))` and the TTL beside it are unguarded, so a non-numeric or zero value either raises at construction or yields a buffer that silently drops everything.

Items 3 and 4 are controller-behaviour claims. I reasoned them from the code and the limiter's documented rate, not from a live run, and after the two defects you caught on #646 I am not going to assert them as verified without hardware saying so.

## Separately

I opened #673 for something this PR makes reachable: `aiounifi` logs raw event payloads on its own logger — `models/event.py:167` emits the whole event dict at WARNING for any key that release does not recognise, which is the normal case when the controller is newer than the pinned library. Client MACs, hostnames and IPs land in the server log. Nothing could trigger it before this PR because nothing opened a websocket. It needs a decision rather than a patch from me, since the metadata whitelist from #671 cannot reach a logger we do not own.

PR #648, opened 2026-09-05
feat(network): add unifi_get_mgmt_settings and unifi_update_mgmt_settings

### LINKED ISSUE #634: feat: add unifi_get/update_mgmt_settings — the device-SSH management plane is invisible
### Component

Network MCP server

### Problem or use case

There is no tool for site setting `key: "mgmt"`, which holds the device-SSH management plane: `x_ssh_enabled`, `x_ssh_username`, `x_ssh_auth_password_enabled`, `x_ssh_keys`, `debug_tools_enabled`, `auto_upgrade`, `auto_upgrade_hour` (plus the secrets `x_ssh_sha512passwd`, `x_ssh_password`, `x_api_token`, `x_mgmt_key`).

So whether device SSH is enabled, what user it runs as, and whether authorised keys are installed cannot be answered through the MCP, which is a material gap for any security review. `POST set/setting/mgmt` applies correctly, so this is a wrapper gap, not an API limitation. At `main` @ `5302363` `system_manager` mentions `mgmt` only in a docstring.

### Proposed solution

Add `unifi_get_mgmt_settings` / `unifi_update_mgmt_settings` next to the snmp and autobackup pair in `system.py`. The getter must redact `x_ssh_sha512passwd`, `x_ssh_password`, `x_api_token` and `x_mgmt_key` the way `unifi_get_snmp_settings` redacts `community`; none of those four names is currently matched by `unifi_core.redaction.is_sensitive_key` (`passwd` is not a recognised segment, and `x_mgmt_key` has no qualifier), so the vocabulary needs those entries too. Setter follows the preview-then-confirm pattern with `current_state` populated from the live record.

### Alternatives considered

Read-only getter first. Enough for audits; the setter can follow once the redaction is in.

### Implementation notes

Same shape as `update_snmp_settings` / `update_autobackup_settings`. Needs `make manifest` and a policy gate (`UNIFI_POLICY_NETWORK_SYSTEM_UPDATE`).

### Contribution

- [x] I am willing to help implement or test this change.

### Code of conduct

### Environment

### Preflight

tmowbrey:
## Summary

Closes #634.

There was no tool for site setting `key: "mgmt"`, the device-SSH management plane, so whether device SSH is enabled, which user it runs as, and whether authorised keys are installed could not be answered through the MCP. `POST set/setting/mgmt` applies correctly; this was a wrapper gap.

Two tools beside the snmp and autobackup pairs in `tools/system.py`:

- `unifi_get_mgmt_settings`: `x_ssh_enabled`, `x_ssh_username`, `x_ssh_auth_password_enabled`, `x_ssh_password` (redacted), `x_ssh_keys`, `x_ssh_bind_wildcard`, `debug_tools_enabled`, `auto_upgrade`, `auto_upgrade_hour`, `advanced_feature_enabled`, `unifi_idp_enabled`, `wifiman_enabled`. Key names verified on Network 10.6.102 (there is no `led_enabled` on this record, contrary to the issue text).
- `unifi_update_mgmt_settings(update_data, confirm)`: preview-then-confirm with `current_state` from the live record, warnings when disabling SSH, changing SSH credentials or replacing the authorised-key list. New permission category `mgmt` (`UNIFI_POLICY_NETWORK_MGMT_UPDATE`, or the blanket `UNIFI_POLICY_NETWORK_UPDATE`), following the single-pair `snmp` precedent so SSH-credential writes can be gated apart from autobackup. The issue guessed `UNIFI_POLICY_NETWORK_SYSTEM_UPDATE`.

Three stored credentials on this record deserve a stronger rule than redaction. `MgmtSettings` does not carry `x_ssh_sha512passwd`, `x_mgmt_key` or `x_api_token` at all; it reports read-only presence flags (`ssh_password_hash_set`, `mgmt_key_set`, `api_token_set`) derived from keys it deliberately omits, so `UNIFI_NETWORK_REDACT_SENSITIVE_FIELDS=false` never hands an agent a device root hash or the adoption key. This is field selection, not a redaction exception: the model is the allowlist on every surface (MCP getter, preview, GraphQL, REST, action endpoint), and an unknown secret spelling in the record cannot reach output. If you want this as a written rule, a sentence for AGENTS.md's "Response redaction" section would be: *a stored credential the tool can never usefully return is not modelled; report its presence as a read-only boolean.*

Redaction kernel (first commit, separable): `passwd` spellings (`x_ssh_sha512passwd`, `x_ssh_md5passwd`, and hash-prefixed variants) and `x_mgmt_key` join the vocabulary; neither was matched before. A boolean value now passes through `redact_sensitive_fields` / `redact_value` regardless of its key name: `x_ssh_auth_password_enabled: true` is state, not a secret. `is_sensitive_key` and the marker write-back guard are unchanged (a bool is never the marker), and a string under such a key is still hidden.

Write path: `mgmt_to_controller_update` rejects unknown and read-only keys by name (a typo beside a valid field must not become a silent drop on the highest-blast-radius write in the server), returns the coerced values (an agent's `"true"` becomes `true` on the wire), and raises value-free messages (pydantic's own quote the submitted value, which may be the SSH password). Only the fields passed are sent; `SystemManager.update_settings` is the existing settings-section path (partial body plus `_id`/`key`), the same call the snmp and autobackup tools make. A fetch-merge-put would echo the stored hash, management key and API token back to the controller. Errors are reported by controller `api.err.*` code or exception class, never by message.

API: Strawberry `MgmtSettings` (delegates to the core converter, redacts `xSshPassword`), `mgmtSettings` query, `GET /v1/sites/{site_id}/mgmt-settings`, action catalog entries, translators (the update translator rejects empty or unknown input like the tool). Tool count 193→195 across every counted surface; manifest, server manifest, catalog, skill reference, SDL, openapi.json and both reference docs regenerated; `permissions.md` gains the `mgmt` row (and the previously missing `system` row for the backup tools); the four "what gets redacted" lists (root README, network README, plugin skill, PRIVACY.md) now name device-SSH credentials and the management key.

## Type of change

- [x] `feat:` new feature

## Scope

- [x] Network MCP server
- [x] API server
- [x] Shared packages
- [x] Documentation

## Checklist

## Testing

```
uv run --package unifi-core --all-extras pytest packages/unifi-core/tests -n auto   # 1864 passed
uv run --package unifi-network-mcp pytest apps/network/tests                        # 1106 passed
uv run --package unifi-api-server pytest apps/api/tests -n auto                    # 1379 passed (all four coverage gates and the drift gates included)
```

New tests: redaction (`passwd` spellings incl. camelCase, `x_mgmt_key`, booleans pass through while strings under the same names stay hidden, marker paths unchanged, `x_ssh_keys` visible); model (field sets, list unwrap, presence flags, secrets absent from the dump, unknown/read-only keys rejected by name, coerced values, value-free validation errors, hour range); manager (the PUT carries only the supplied keys plus `_id`/`key`); tools (getter redacts and keeps flags and keys; policy-off returns the SSH password but never the stored secrets, including an unknown `*_passwd` spelling; preview uses the live state, redacts current and proposed, warns on SSH disable, credential change and key-list replacement; confirm sends only the supplied fields and echoes them without the manager's `_id`/`key`; empty, unknown, read-only and mixed payloads rejected by name; invalid or malformed fields named without their value; controller `api.err.*` code surfaced and message-shaped codes not; getter, preview and confirm failures leak nothing; manager rejection reported; `x_ssh_keys` round trip); policy gate (every manifest category resolves to its documented `UNIFI_POLICY_NETWORK_*` variable, `mgmt` included); API (cross-layer symmetry, serializer, GraphQL default and policy-off, REST route, translators, action endpoint against the regenerated catalog).

Not run: a live write to a controller. The partial-PUT merge behaviour of `/set/setting/mgmt` rests on the same endpoint family as the existing snmp and LED writes (a measured partial write to `set/setting/snmp` changed one field and kept the v3 credentials); if you know a controller version that replaces the document instead, say so and I will switch to fetch-merge-put with the stored secrets stripped.

## Notes for reviewers

- `x_ssh_keys` is accepted as `list[Any]` because the item shape could not be verified (the list was empty on the reference site); the preview warns that it replaces the whole list.
- `_error_detail` also appears in #645 (SNMPv3); whichever lands second collapses the two into one helper.
- Tests for the new tools live in `test_backup_tools.py` beside `TestSnmpTools`, following that file's precedent.
- The plugin skill under `.agents/` lists redacted fields too and is maintainer-owned; left for you.

### Deferred items

- `update_autobackup_settings` echoes the `_id`/`key` the manager adds to its dict (pre-existing; this PR passes a copy for `mgmt` only).
- `SystemManager.update_settings` logs the exception text and the rejected response body at ERROR (pre-existing); #645 changes that for the settings path.
- A typed model for `x_ssh_keys` items once the shape is known.

### Considered & declined

- Carrying the three stored secrets and redacting them: policy-off mode would return a device root hash and the adoption key; presence booleans meet the stated need ("does a credential exist").
- Fetch-merge-put per the update-tool golden path: it would send the stored secrets back to the controller on every write.
- Reusing category `system`: an operator who set `UNIFI_POLICY_NETWORK_SYSTEM_UPDATE=false` for backups would not gate SSH-credential writes, and vice versa.


sirkirby (changes requested):
Tyler, I reviewed `ea0c8904`, including the latest rebase, and reran the credential-error reproduction on this commit. The normal MCP response is redacted, but I need the error path corrected before this can advance.

**P1 — protect the SSH credential before manager logging and API audit handling.** The new management update calls the real `SystemManager`, which still [logs the raw response/exception](https://github.com/sirkirby/unifi-mcp/blob/ea0c890454ba55bfa2d12212ad94df39b088096e/packages/unifi-core/src/unifi_core/network/managers/system_manager.py#L327-L332). With a fake controller exception containing a synthetic submitted SSH password, the MCP response omitted the password but the manager ERROR log included it. Production API dispatch also propagated the same exception to [the route's `detail=str(e)` audit path](https://github.com/sirkirby/unifi-mcp/blob/ea0c890454ba55bfa2d12212ad94df39b088096e/apps/api/src/unifi_api/routes/actions.py#L287-L297). I executed the manager/MCP/API-dispatch chain; audit persistence was source-traced. This is an existing generic error sink newly exposed to the device SSH credential operation.

Please carry value-free errors through the manager and transport layers before logging or rethrowing, and ensure the API audit detail is sanitized too. Cover read failures, rejected responses and write exceptions with a real-manager/fake-connection regression that asserts a sentinel is absent from responses, logs and audit detail. This overlaps #645, so one shared correction should serve both.

Before accepting the writer, I also need evidence that the partial settings PUT preserves omitted controller values, plus the supported SSH-key element schema: [`List[Any]`](https://github.com/sirkirby/unifi-mcp/blob/ea0c890454ba55bfa2d12212ad94df39b088096e/packages/unifi-core/src/unifi_core/network/models/system.py#L142-L144) currently accepts arbitrary key-array members. The existing partial-body test does not establish controller preservation. Please avoid replaying stored hashes merely to force a generic merge pattern; we should establish the settings-specific contract first. The live read I checked had no configured SSH keys, so it did not settle that contract.

The focused source suites and normal read/redaction checks passed, but no live SSH/password/key mutation or recovery cycle was performed. The new Core exports also need a published Core release and corresponding Network/API dependency floors before installed-package validation. I have re-enabled CI for this rebased head; its results remain pending.

sirkirby (commented):
Tyler, I need to narrow the product decision before continuing with the writer. I accept the read-only SSH audit need in [#634](https://github.com/sirkirby/unifi-mcp/issues/634): enabled state, username, password-auth state and whether authorized keys or stored credentials exist. I want to ship that getter/presence view first and defer the broad management setter, password changes and whole-key-list replacement to a separately scoped feature. The issue already identifies read-only first as a workable alternative.

Please split the writer from this change; separable shared redaction fixes can stand on their own. A read-only posture view does not need raw passwords or arbitrary key objects. Our real read verified redaction with an empty key list, which does not establish a populated-key schema. An existing sanitized populated-key shape, if available, would support accurate presence/count handling without changing controller keys.

For a future writer, I need a concrete operator workflow beyond the audit questions and evidence for this exact settings endpoint. The measured SNMP partial update does not establish management-setting preservation. Once its security and schema issues are corrected, that evidence should come from an isolated lab gateway with test credentials and independent recovery access, fresh read-back of omitted fields, confirmation/policy no-write checks and restoration. Please do not change production SSH credentials or keys to satisfy this request while the known error-path issue remains.

Acceptance follow-up at `ea0c890454ba55bfa2d12212ad94df39b088096e`.

tmowbrey:
Rewritten as the read-only posture view you accepted. Tested head: `81d5745` (branch rebuilt on #660 `9ff0624`, the shared credential error-path correction; the 5 conflicts are gone).

**What changed against each finding**

- P1 credential error path (`system_manager.py:327-332`, `routes/actions.py:287-297`): fixed in #660, which this branch sits on. Nothing in this PR writes to the controller any more, so the mgmt credential never enters a request body.
- Writer split off: `unifi_update_mgmt_settings`, `mgmt_to_controller_update`, the `mgmt` permission category, the API translator and the mutation serializer entry are gone. A future writer is a separate feature per your note (operator workflow, isolated lab gateway evidence).
- `List[Any]` for SSH keys (`models/system.py:142-144`): the key list is no longer modelled. `MgmtSettings` reports `ssh_keys_present` and `ssh_keys_count` only. Our controller, like yours, has no configured keys (`x_ssh_keys: []`), so a populated-key element schema is still unobserved; I did not invent one.
- Raw passwords: the controller returns `x_ssh_password` in clear text on read. The view omits it (and `x_ssh_sha512passwd`, `x_mgmt_key`, `x_api_token`) entirely and reports `ssh_password_set`, `ssh_password_hash_set`, `mgmt_key_set`, `api_token_set` as booleans. The Core converter is the single allowlist; the MCP tool, GraphQL type and REST route all project through it, and the policy flag has nothing left to reveal (`redact_sensitive=false` returns the same shape; tests cover both).
- Kept: `unifi_get_mgmt_settings`, `mgmtSettings` GraphQL query, `GET /v1/sites/{site_id}/mgmt-settings`, the redaction-vocabulary fix (`passwd`/`sha512passwd` spellings, `x_mgmt_key`, booleans pass through). The feature is 34 files on top of #660 (the PR view shows 40 while #660 is unmerged), down from 40; tool count 193 → 194.

**Read-only smoke** — owned controller, Network Application 10.6.102, this head, entry points `unifi_get_mgmt_settings` (MCP tool → real `SystemManager` → real `ConnectionManager`) and the API `MgmtSettings.from_manager_output` projection on the same record:

| check | result |
|---|---|
| getter `success` | pass |
| MCP payload equals the API projection (both policy settings) | pass |
| field set = 16 posture fields, none of `x_ssh_password` / `x_ssh_sha512passwd` / `x_mgmt_key` / `x_api_token` / `x_ssh_keys` | pass |
| controller record holds 4 stored credential values; any of them in MCP or API output | false |
| any of them, or the login password, in 18 captured DEBUG log lines | false |
| `ssh_keys_present` / `ssh_keys_count` on a controller with no keys | false / 0 |

Suites: Core 1922, Network unit 1089, API 1379, catalog 1177 passed; SDL, openapi, reference docs, manifest, catalog and skill references regenerated; `make check-generated` clean. No SSH credential or key on the controller was changed.

Floor alignment: the Core model/converter and the redaction-vocabulary fix ship here; the Network/API floor bump follows the #657 pattern once you cut Core, or I can split the Core half out (same question as in #658).

sirkirby (commented):
Tyler, I reviewed `81d57458` and accept the narrowed read-only posture feature. The writer, mutation translator and policy category are removed; the Core allowlist exposes 16 posture/presence/count fields without passwords or SSH key objects under either response-policy setting. MCP, REST and GraphQL project through that shared converter. I independently passed 501 focused Core/Network/API checks, including all API artifact drift gates, with no new blocking defect in the getter's unique delta.

Your real-controller getter and matching in-process API projection record materially addresses the acceptance request. It establishes that read/projection behavior; installed MCP discovery and actual REST/GraphQL reads remain final delivery checks. There is no need to add or replace production SSH keys just to manufacture a populated-key sample.

This branch still contains the exact #660 escaped-credential flaw, so it remains held on that correction and the published Core/Network/API floor alignment. The old writer objections do not apply to this revision. Please also align the title/body with the actual read-only implementation; their remaining writer discussion is stale.

sirkirby (approved):
Approving. I rebased and resolved this myself — #645 landed on the same surfaces while this was open, so it needed real merge work rather than another round-trip.

## The posture design is right, and I proved it on hardware that has all four secrets set

The reference controller stores every secret-bearing key in the `mgmt` section — SSH password, SHA-512 hash, management key and API token are all populated. Running the tool and comparing its output against the raw record:

```
Secret-bearing keys populated on this controller:
  x_api_token, x_mgmt_key, x_ssh_password, x_ssh_sha512passwd

leaked secret values: NONE
```

What comes back instead is `ssh_password_set`, `ssh_password_hash_set`, `mgmt_key_set`, `api_token_set` — presence, not material — alongside the state an operator actually needs: `x_ssh_enabled`, the SSH user, `x_ssh_auth_password_enabled`, `x_ssh_bind_wildcard`, key count, `debug_tools_enabled`, `auto_upgrade`. That answers the security-review question in #634 without ever handing an agent a credential, and it is a better answer than returning the fields redacted would have been.

Narrowing to a read-only getter was the right call.

## Vocabulary

The four names #634 identified are now matched, including the two that were not before:

```
x_ssh_sha512passwd  True      x_mgmt_key   True
x_ssh_md5passwd     True      xMgmtKey     True
x_ssh_bcrypt_passwd True      x_api_token  True
```

`x_ssh_keys` correctly stays visible — public keys are posture, not secrets.

The boolean exemption is the subtle part and it is handled correctly. `x_ssh_auth_password_enabled` returns `True` rather than a mask, while a *string* under a secret-looking name is still hidden, and the key itself stays sensitive so the write-back guard is untouched:

```
redaction_marker_paths({'x_ssh_auth_password_enabled': True, 'x_ssh_password': REDACTED})
  -> ['x_ssh_password']
```

That last property is what makes the exemption safe, and I am glad the test asserts it directly.

## One thing I changed rather than accepted

The rebase surfaced a conflict in `_scrub_any` where this branch and the merged #660 disagree. Taking this side verbatim would have regressed two things:

- it returns `marker` where #660 returns `REDACTED`. `ConnectionManager` passes its own credential mask, so a payload value would end up carrying a marker `redaction_marker_paths` does not recognise — the guard would stop seeing it and a masked value could reach a controller write.
- it returns a plain dict, dropping #660's mapping-type rebuild that preserves aiohttp's case-insensitive multidict.

Resolved by keeping #660's structure and swapping in only your `_is_secret_entry` predicate, which is the part that is actually new here:

```python
key: REDACTED if _is_secret_entry(key, item) else _scrub_any(item, ordered, marker)
```

Both behaviours are covered by tests that pass.

## Also fixed

`apps/network/docs/tools.md` still advertised `unifi_update_mgmt_settings`, left over from when the setter was in scope. Only the getter exists, so the line is gone — a doc promising a tool the schema does not have is the exact failure #635 catalogued.

## Verification

Rebase dropped `9ff0624d` (#660, now merged) and replayed the two real commits. Fifteen conflicts across eleven files; the generated artifacts were regenerated rather than hand-merged and came back byte-identical to my resolutions, and `make check-generated` reports no drift.

Suites: core 2240, API 1425, Network 1147, Shared 436, Protect 513, Access 222. Repo-wide ruff clean. The resolver-coverage gate caught a missing `# tool:` marker during my resolution and I restored it — that gate earned its keep.

## Release

This changes `packages/unifi-core`, so it shares `core/v0.4.47` with #645, then Network and API floors to `>=0.4.47` and those two releases.

PR #649, opened 2026-09-05
fix: guide callers to accepted MAC argument names

tmowbrey:
Network tools use several MAC argument names. When a caller supplies a different MAC spelling, strict dispatch now adds the tool's accepted parameter name to the existing unknown-argument error. It offers a hint only when exactly one MAC parameter is accepted and the caller has not already provided it. Ambiguous cases retain the existing error.

Client, blocked-client, and device list descriptions now identify the parameter names accepted by their follow-up tools. The generated Network manifest includes those descriptions. No aliases are accepted, no arguments are rewritten, and schemas and permission handling are unchanged.

Addresses the MAC argument discovery portion of #635. Other portions of that issue remain separate; this PR does not close the whole issue.

Validation: Shared strict-dispatch regression tests cover accepted, unknown, ambiguous, and conflicting spellings. Network tests check manifest guidance. Independently installed Shared and Network wheels passed real read-only MCP checks in eager, lazy, and meta-only registration modes: the wrong spelling was rejected with the canonical-name hint, retrying with that name succeeded, and the supplied MAC was not echoed in the error.

Full `make pre-commit`, all GitHub checks, and a fresh installed-wheel repeat of all three registration-mode probes passed on 5b12776b81f69f508ec75f691eacd6e1e22db37e. Publish the Shared change first, then require that released Shared version in downstream app metadata before app publication.

### DIFF
--- a/apps/network/src/unifi_network_mcp/tools/clients.py
+++ b/apps/network/src/unifi_network_mcp/tools/clients.py
@@ -71,7 +71,12 @@ async def lookup_by_ip(
         "include_offline=true for historical clients. Use search to filter by "
         "name/hostname/IP/MAC (case-insensitive). For a single client's full raw "
         "object, use unifi_get_client_details. For IP-to-client lookup, use "
-        "unifi_lookup_by_ip."
+        "unifi_lookup_by_ip. Pass an entry's mac to the client tools as "
+        "mac_address (unifi_get_client_details, unifi_block_client, "
+        "unifi_rename_client, ...), as client_mac (unifi_get_client_sessions, "
+        "unifi_get_client_dpi_traffic, unifi_get_client_wifi_details) or as mac "
+        "(unifi_recent_events); the parameter name differs per tool and each tool "
+        "accepts only its own."
     ),
     annotations=ToolAnnotations(readOnlyHint=True, openWorldHint=False),
 )
@@ -168,7 +173,11 @@ async def get_client_details(
 
 @server.tool(
     name="unifi_list_blocked_clients",
-    description="Lists clients/devices currently blocked from the network. Each entry includes mac, hostname, name, last_seen, blocked.",
+    description=(
+        "Lists clients/devices currently blocked from the network. Each entry includes "
+        "mac, hostname, name, last_seen, blocked. Pass an entry's mac to "
+        "unifi_unblock_client as mac_address."
+    ),
     annotations=ToolAnnotations(readOnlyHint=True, openWorldHint=False),
 )
 async def list_blocked_clients() -> Dict[str, Any]:
--- a/apps/network/src/unifi_network_mcp/tools/devices.py
+++ b/apps/network/src/unifi_network_mcp/tools/devices.py
@@ -43,7 +43,14 @@ logger = logging.getLogger(__name__)
         "Use search to filter by name, IP, or MAC (case-insensitive). "
         "Set include_details=true for additional per-device fields: by default (summary=true) "
         "compressed radio/port summaries; set summary=false to return the full raw tables "
-        "(radio_table, port_table, network_table, system_stats, wan1/wan2)."
+        "(radio_table, port_table, network_table, system_stats, wan1/wan2). "
+        "Pass an entry's MAC to the device tools as mac_address "
+        "(unifi_get_device_details, unifi_reboot_device, unifi_rename_device, "
+        "unifi_upgrade_device, ...), as device_mac (unifi_get_switch_ports, "
+        "unifi_get_port_stats, unifi_set_switch_port_profile, unifi_locate_device, ...) "
+        "or as ap_mac on the RF tools (unifi_trigger_rf_scan, unifi_get_rf_scan_results), "
+        "which take the access point to scan from; the parameter name differs per tool "
+        "and each tool accepts only its own."
     ),
     annotations=ToolAnnotations(readOnlyHint=True, openWorldHint=False),
 )
--- a/apps/network/src/unifi_network_mcp/tools_manifest.json
+++ b/apps/network/src/unifi_network_mcp/tools_manifest.json
@@ -10943,7 +10943,7 @@
         "openWorldHint": false,
         "readOnlyHint": true
       },
-      "description": "Lists clients/devices currently blocked from the network. Each entry includes mac, hostname, name, last_seen, blocked.",
+      "description": "Lists clients/devices currently blocked from the network. Each entry includes mac, hostname, name, last_seen, blocked. Pass an entry's mac to unifi_unblock_client as mac_address.",
       "name": "unifi_list_blocked_clients",
       "schema": {
         "input": {
@@ -11111,7 +11111,7 @@
         "openWorldHint": false,
         "readOnlyHint": true
       },
-      "description": "Returns connected clients with mac, name, hostname, ip, status (online/offline), connection type (wired/wireless), and for wireless clients: ssid, signal_dbm, channel, radio. Filter by filter_type (all/wired/wireless), set include_offline=true for historical clients. Use search to filter by name/hostname/IP/MAC (case-insensitive). For a single client's full raw object, use unifi_get_client_details. For IP-to-client lookup, use unifi_lookup_by_ip.",
+      "description": "Returns connected clients with mac, name, hostname, ip, status (online/offline), connection type (wired/wireless), and for wireless clients: ssid, signal_dbm, channel, radio. Filter by filter_type (all/wired/wireless), set include_offline=true for historical clients. Use search to filter by name/hostname/IP/MAC (case-insensitive). For a single client's full raw object, use unifi_get_client_details. For IP-to-client lookup, use unifi_lookup_by_ip. Pass an entry's mac to the client tools as mac_address (unifi_get_client_details, unifi_block_client, unifi_rename_client, ...), as client_mac (unifi_get_client_sessions, unifi_get_client_dpi_traffic, unifi_get_client_wifi_details) or as mac (unifi_recent_events); the parameter name differs per tool and each tool accepts only its own.",
       "name": "unifi_list_clients",
       "schema": {
         "input": {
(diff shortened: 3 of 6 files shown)


sirkirby (changes requested):
Tyler, I reviewed `656841fb`. I found no functional or security defect in the alias dispatch: the actual handler passed preview, confirmation, collision rejection, policy denial, execute and batch probes in eager, lazy and meta-only modes. The combined candidate with #651/#652 also passed 437 focused tests after preserving both independently added helpers in `tool_index.py`.

I’m requesting changes for the dependency contract. All three apps now [import `create_import_safe_tool_decorator`](https://github.com/sirkirby/unifi-mcp/blob/656841fb694ae92388a8c551ab1e94dfd6b1e0af/apps/network/src/unifi_network_mcp/runtime.py#L30), but their requirements still allow Shared versions without it. I reproduced the import failure with Shared 0.6.13, which is also allowed by current main’s newer floor. Workspace tests supply the new Shared source and therefore miss this installed-app startup failure.

The Shared foundation needs to be published, all three app lower bounds raised to that actual release, and installed-wheel startup verified before the app changes are delivered. This can be handled through a coordinated foundation release; it is the only change request from my review.

sirkirby (commented):
Tyler, I want to separate the valid MAC-consistency problem in #635 from accepting this particular compatibility layer. These spellings were previously invalid on each target tool, and the proposal accepts them outside the advertised canonical schema while leaving REST/GraphQL and the different canonical names unchanged. I’d defer the broad alias layer until we choose that contract; clearer canonical discovery and argument-error guidance are alternatives worth comparing.

Please show the concrete supported-client workflow this improves, including client/plugin versions, registration mode and tested commit. For acceptance, a canonical and alias read-only lookup against the same real controller should return the same resource/result; report that comparison without MACs or other private identifiers. Show how the alias-only request reaches the server despite the canonical required field, and whether the client validates the schema first. Also clarify the canonical-name/migration policy and MCP-only scope. No reboot, block or other controller mutation is needed.

The previous handler tests establish server-side behavior, not that client workflow. My earlier statement that package floors were the only change request did not settle feature acceptance. This note concerns current proposal `d771ed13`; earlier technical evidence remains pinned to `656841fb` pending revision review.

Acceptance follow-up at `d771ed135bde2eb396fc3b984e227fa5268da3eb`.

tmowbrey:
Evidence for the questions in your acceptance note, then a narrower proposal. Tested revision `d771ed13`; baseline is the installed plugin. No MACs or controller identifiers below.

**Client workflow.** Claude Code CLI 2.1.263 (earlier failures span 2.1.221 → 2.1.261) with the `unifi-network` Claude plugin at 0.29.8 (`uvx unifi-network-mcp==0.29.8`; installed wheels Shared 0.6.13, Core 0.4.43, mcp 2.1.1). Registration mode: lazy, the plugin default, so the client sees five meta-tools and calls domain tools through `unifi_execute`.

**How an alias-only request reaches the server.** In lazy mode the only client-visible schema is `unifi_execute` with `arguments: {type: object, additionalProperties: true}`; per-tool required fields are not in any schema the client can see. In eager mode the per-tool schema does list `required: ["mac_address"]`.

**Does the client validate first? No.** Eager mode, Claude Code 2.1.263, installed 0.29.8, controller deliberately unreachable so dispatch is the first thing a call hits: an alias-only call `{"mac": …}` to `unifi_get_client_details` returned the server's own `Invalid params for 'unifi_get_client_details': unknown arguments {mac}. Valid arguments: [include, mac_address, summary].` (built in `strict_dispatch.py`). The request reached the server; the missing required field was not caught client-side. The Python MCP client used below behaves the same.

**Who produces the alias spellings.** A census of every Claude Code session on this host, 2026-08-05 → 2026-09-04: 14 alias failures, all via `unifi_execute` in lazy mode, all composed by the model, every one retried with the canonical name within a minute and succeeded. Direction: `mac` where `mac_address` was canonical 6; `mac_address` where `client_mac` was canonical 5; `mac_address` where `device_mac` 2; `mac` where `device_mac` 1. The bare `mac` comes from list output: `unifi_list_clients` / `unifi_list_devices` items carry the field as `mac`, and the model copies the output field name into the next call.

**Same-controller comparison** (one online client, one switch, read-only, lazy mode; repeated in eager mode with identical outcomes). Same result = identical `data` payload hash, or identical identity keys when counters moved between reads.

| Build | Canonical calls | Alias calls | Alias outcome | Same resource |
|---|---|---|---|---|
| installed 0.29.8 | 4/4 success | 12 | 12/12 rejected `unknown arguments`, no controller request | n/a |
| `d771ed13` | 4/4 success | 12 | 12/12 success | 12/12 true |

Tools and canonical names: `unifi_get_client_details` (`mac_address`), `unifi_get_device_details` (`mac_address`), `unifi_get_switch_ports` (`device_mac`), `unifi_get_client_sessions` (`client_mac`). No mutation, no REST/GraphQL.

**Proposal: narrow this PR to your alternative.** The evidence shows the alias layer works, but the cost it removes is one self-corrected round trip per failure, never a wrong result, and the failures trace to two discoverability gaps, not to the schema. I will rewrite this branch as: (1) the `unknown arguments` message names the canonical parameter when the rejected key is a known MAC spelling (`mac`, `mac_address`, `device_mac`, `client_mac`), e.g. `… unknown arguments {mac}: this tool takes mac_address`; (2) list-tool descriptions state which parameter name the next call takes. No aliases accepted, no schema or catalog change, MCP-only by construction, canonical names unchanged. Migration policy: none needed, since nothing new is accepted. The Shared floor concern goes away with the dispatch layer. Unless you would rather see the rename question settled first, I will push that revision here.

sirkirby (commented):
Tyler, I agree with the narrower revision you proposed: keep the canonical argument contract, make the unknown-argument error point to the correct MAC parameter, and clarify the next-call parameter in the relevant list descriptions.

Your client/version and read-only comparison evidence answers the workflow question. The 14 self-corrected calls support improving discovery and guidance; I do not think they justify the broad implicit alias layer currently at `d771ed13`. Please push the guidance-only revision here. No rename or alias migration is needed for that direction. I will verify the resulting diff and any actual Shared dependency requirement rather than assuming the old floor issue disappears.

This is a scope decision, not approval of the current implementation. The branch is still 10 commits behind current main, although its trial merge is clean; we will handle final integration and checks on the revision we intend to deliver.

tmowbrey:
Tested SHA: `5671dbb3` (rebased onto `1ed216cc`; the alias branch is gone).

## The alias layer is removed

> I agree with the narrower revision you proposed: keep the canonical argument contract, make the unknown-argument error point to the correct MAC parameter, and clarify the next-call parameter in the relevant list descriptions.

Rewritten to that, from scratch on current main. Gone: `argument_aliases.py`, the mixin, the alias declarations across 40 tools, the alias metadata in `tools_manifest.json`, and the dispatch rewriting. The diff is now 6 files, +187/-7, and no alias is accepted anywhere — `unifi_get_client_details(client_mac=…)` still fails.

All four Copilot findings were against that deleted code and go with it.

**(a) The rejection names the parameter.** `StrictKwargFastMCP.call_tool` already answered `unknown arguments {device_mac}. Valid arguments: [include, mac_address, summary]`, which lists the right name without saying which one holds the MAC. It now appends `'unifi_get_client_details' takes the MAC address as 'mac_address'.` The hint fires only when a rejected key is one of the five spellings the Network tools actually use (`mac`, `mac_address`, `client_mac`, `device_mac`, `ap_mac`) **and** the tool's own schema contains exactly one of them. A tool with no MAC parameter, or a rejected key that is simply wrong, reads exactly as before.

**(b) The list tools say what the next call takes.** `unifi_list_clients` and `unifi_list_devices` each name both families (`mac_address` on the get/mutate tools, `client_mac` / `device_mac` on the others) and state that the parameter name differs per tool and no other spelling is accepted; `unifi_list_blocked_clients` names `unifi_unblock_client`. A test asserts every tool and parameter those descriptions name exists in the manifest, so the guidance cannot rot into pointing somewhere that is not there.

## The dependency floor

> I will verify the resulting diff and any actual Shared dependency requirement rather than assuming the old floor issue disappears.

**This revision needs no floor bump, and I checked rather than assumed.** The reason the old one did is gone: no app imports a new symbol. The Shared change is a behaviour change *inside* `StrictKwargFastMCP.call_tool` — a module-private constant and a module-private helper, neither exported nor imported by anything. An app running an older installed Shared therefore degrades to the old message; it does not fail to import.

Two clean venvs, the app installed from this branch with `--no-sources` so the dependencies resolved from the index, driven over stdio MCP in `eager` mode against an unreachable host (no controller, no login):

| | `unifi-mcp-shared` | server starts | argument still rejected | rejection names the parameter |
|---|---|---|---|---|
| released | **0.6.15** (the floor on main) | yes | **yes** | no |
| this branch | 0.6.16.dev | yes | **yes** | **yes** |

Cases exercised, identically in both: `unifi_get_client_details(device_mac=…)`, `unifi_get_client_sessions(mac_address=…)`, `unifi_get_switch_ports(mac_address=…)`, `unifi_recent_events(mac_address=…)`, and `unifi_list_networks(mac_address=…)` as the negative — a tool with no MAC parameter, where the hint must not appear and does not, on either version. The probe MAC was `00:00:00:00:00:00`.

The list-description half is app-side and reaches clients through `tools_manifest.json`, so it is present on both.

## Review findings fixed

Gate 2 on this revision found three things worth having:

- **`ap_mac` was in the spelling set, and it names a different thing.** An AP MAC is the access point to scan *from*, not the subject of the call. `unifi_get_client_sessions(ap_mac=…)` answered "takes the MAC address as `client_mac`", and an agent that obeys re-sends an AP MAC as a client MAC and gets `success: true, count: 0` — a silent wrong answer, which is the class of bug this module exists to prevent. Dropped, with a test in both directions.
- **The hint fired when the canonical name was already in the call.** `unifi_get_client_details(mac_address=…, client_mac=…)` was told to send `mac_address`, which it had sent. It now says nothing.
- **"no other spelling is accepted" was false.** Both descriptions enumerated two families, and `ap_mac` and `mac` exist. They now name every spelling in use, and a test asserts that: any MAC parameter appearing on any tool must be named by one of the two descriptions.

Two more tests came out of it: the "two MAC parameters, say nothing" guard was only ever reached through the zero-parameter case, and nothing asserted the registered decorator description matches the manifest (eager registration serves one, lazy mode the other).

## Checks

| | |
|---|---|
| `make shared-test` | **490 passed** |
| `apps/network/tests` | **1163 passed** |
| `apps/protect/tests` / `apps/access/tests` | **514** / **223 passed** (they share `StrictKwargFastMCP`; neither manifest has a parameter in the spelling set, so neither can emit the hint today) |
| `make check-generated` | **pass** (no drift) |
| `ruff format` / `ruff check` | **pass** |

Run separately — `apps/` paths collide in a single pytest invocation.

## Scope

MCP only, as agreed. The REST and GraphQL surfaces reject unknown arguments through `_validate_action_args` in `apps/api/.../services/actions.py`, which is untouched; their canonical names are unchanged and no migration is implied anywhere.

### Deferred items

- **The canonical names themselves.** Five spellings for one value across 39 tools is the underlying problem; this only makes the mismatch self-correcting at the point of failure. Renaming is a breaking change to the advertised schema and wants its own decision, which is why nothing here renames anything.
- `source_mac` → `source_macs` (the singular/plural slip on `unifi_create_acl_rule`) gets no hint. It is a different mistake and the hint table is deliberately only the MAC-spelling family.
- **There is no drift gate on `tools_manifest.json`.** `make check-generated` covers skill references, the API action catalog and support skills, but not the tool manifests, so a description edited in a decorator and not regenerated would ship half-applied. The new parity test covers the three tools this PR touches; a general gate is a separate change.

### Considered & declined

- **Extending the hint to the REST/GraphQL validator.** You scoped this to MCP, and the two surfaces have different canonical names in places; making them agree is the rename decision above, not this.

sirkirby (approved):
Tyler, I re-reviewed the guidance-only implementation and rebased it onto current main. I removed two stale issue references from test text and rewrote the PR description to match the final implementation without closing the remaining portions of #635.

On 5b12776b81f69f508ec75f691eacd6e1e22db37e, the full `make pre-commit` passed and every current GitHub check is green. I rebuilt and independently installed Shared and Network wheels from this commit. In eager, lazy, and meta-only modes, the wrong MAC spelling was rejected with the canonical-name hint, the supplied MAC was not echoed, and retrying with the correct name completed a real controller read.

I am merging this, publishing Shared before raising downstream dependency floors, and keeping app releases behind the remaining debt checks.

PR #650, opened 2026-09-05
feat(core): add NAT rule model and manager

tmowbrey:
## Summary

Adds the NAT rule domain to `unifi-core`: a `NatRule` model and a `NatManager` for the V2 endpoint `/v2/api/site/{site}/nat` (DNAT, SNAT, Masquerade). This is the core half of the DNS-redirect ask in #216; the Network tools, API mirror and skill docs follow in a second PR once this ships in a `core/v*` release, because `check-pin-alignment` builds the Network app against the published core floor and a runtime that imports `nat_manager` cannot pass at `>=0.4.41`. Nothing under `apps/` changes here.

Refs #216. Follows #646.

### What the controller told us

Enum spellings and per-type key sets come from a Network 10.6.102 controller: rules of each UI variant were created (disabled) through the endpoint and read back, and the controller's own 400 bodies enumerate the closed sets.

- `type`: `DNAT`, `SNAT`, `MASQUERADE`.
- `filter_type`: `NONE`, `ADDRESS_AND_PORT` (`address` and/or `port`; either alone is accepted), `FIREWALL_GROUPS` (`firewall_group_ids`), `NETWORK_CONF` (`network_conf_id`, singular), `IID_AND_PORT` (IPv6; keys not read back, passes through unvalidated).
- `firewall_group_ids: []`, `invert_address`, `invert_port` are present on every filter regardless of type.
- Controller-side rules mirrored by the validator: DNAT needs `in_interface`, `ip_address` and a destination filter other than `NONE` ("If a port/IP translation is set, a corresponding destination filter must also be set"); SNAT needs `out_interface` and `ip_address`; MASQUERADE needs `out_interface` and rejects `ip_address`/`port`; a selector under a filter type that does not use it is rejected ("Address/Port filter may not have Network Configuration or Firewall Groups set").
- `rule_index` must be unique. A create without one collides with the controller default (`duplicate rule_index`), so `create_nat_rule` assigns max+1 when the caller omits it.
- Write paths confirmed live: `POST /nat`, `PUT /nat/{id}`, `DELETE /nat/{id}`. There is no `GET /nat/{id}`.

### Model (`models/nat.py`)

- Read-only `id`, `is_predefined`, `setting_preference`; everything else mutable. `MUTABLE_FIELDS` / `READ_ONLY_FIELDS`, `from_controller`, `reject_unknown_fields`, `to_controller_create`, `to_controller_update`, as in `models/dynamic_dns.py`.
- One selector table per level, `RULE_SELECTORS` (`ip_address`, `port` under `type` DNAT/SNAT) and `FILTER_SELECTORS` (`address`, `port`, `network_conf_id`, `firewall_group_ids` under their `filter_type`), drives four things: the requirement check (an activating value needs one of its selectors), the reverse check (a set selector under any other observed value is rejected), the read view (`from_controller` hides a selector the controller ignores) and retirement on update (a selector the update deactivates is dropped, `firewall_group_ids` reset to `[]`). Unobserved `type` / `filter_type` values pass through every one of those.
- `normalize_nat_update` rejects unknown keys, normalises enums, type-checks through the pydantic model in strict mode and drops `None` inside filters; `normalize_nat_create` adds `nat_rule_error`. Port strings: `re.fullmatch`, ASCII digits, 1-65535, `low-high`, comma lists, no whitespace.
- `merge_nat_update` produces the document to PUT; `nat_update_error` validates it and reports a problem when it concerns a key the update changed or the stored rule did not already have it, so a stored rule this project did not author is not held against an unrelated update while a swapped stale value still reports.

### Manager (`managers/nat_manager.py`)

`list_nat_rules` (cached, `nat_rules` prefix, bare-list or `{data}` envelope; raises rather than caching `[]` when the body cannot be decoded), `get_nat_rule` (list + filter, `UniFiNotFoundError`), `create_nat_rule` (fresh list for the auto `rule_index`, skipping predefined rules), `update_nat_rule`, `delete_nat_rule` (resolves the id against the list before sending it), `toggle_nat_rule` (one fetch). A 404/405 on the list, excluding aiounifi's auth exceptions, maps to a `UniFiOperationError` saying NAT rules need Network 9.0+ with a UniFi gateway. `exceptions.http_status` reads the status aiounifi folds into its message or the V2 error body. This module's logger calls carry operation names and exception class names only.

## Type of change

- [x] `feat:` new feature

## Scope

- [x] Shared packages

## Checklist

- [x] I updated generated manifests with `make manifest` when changing MCP tools. (No tool changes in this PR.)
- [x] I updated documentation for user-visible changes. (Module docstrings; user-facing docs land with the tools PR.)

## Testing

```
uv run --package unifi-core --all-extras pytest packages/unifi-core/tests -q
uv run python scripts/check_pin_alignment.py
make pre-commit
```

Results: 1988 core tests pass (49 for the NAT model and 39 for the manager, plus one for `http_status`). Ruff format and check clean. The full `check_pin_alignment.py` run passes; note that the `--network-core-floor-only` shortcut fails on `upstream/main` itself since #625, because that mode installs `unifi-mcp-shared` from PyPI (0.6.12, no `support_bundle` module) instead of the workspace wheel, while CI runs the full script. `make pre-commit` passes.

Live: the seven probe rules created during discovery were deleted through `DELETE /nat/{id}`; the controller reports none remaining.

## Notes for reviewers

- Sequencing: the tools PR needs a `core/v*` tag that includes this module and will bump the floor in `apps/network/pyproject.toml` and `apps/api/pyproject.toml`.
- `in_interface` on SNAT and MASQUERADE is stored when supplied and not required; the validator does not touch it.

### Deferred items

- `update_nat_rule` returns the locally merged document, not the PUT response body; the tools PR does a read-back after every mutation, which is where a controller-side normalisation would show.
- `ConnectionManager.request` logs `str(e)` for aiounifi request errors and the decoded V2 error body with a traceback for other exceptions, one layer below this manager. The manager's own logger calls carry class names only; the connection-layer line is pre-existing and repo-wide.
- `is_predefined` is modelled but not enforced: update and delete will act on a controller-predefined rule. `firewall_manager` has the same gap, so this is a repo-level decision.
- Ten managers interpolate a caller-supplied id straight into a V2 path; NAT resolves the id against the list first. A shared id validator would close the others.
- `IID_AND_PORT` (IPv6) keys were not read back and pass through unvalidated; `exclude` and `pppoe_use_base_interface` are carried but their semantics were not exercised.

### Considered & declined

- Showing selectors the controller ignores (an `ip_address` on a MASQUERADE, an `address` under `filter_type: NONE`) in the read view. The controller stores and enforces a selector only under its activating value, and the write path reports any attempt to change a hidden value, so surfacing them would present dead configuration as live. Do-nothing test: a reader sees a target address on a rule that translates nothing.
- Dropping the `ensure_connected()` guard before each request because `ConnectionManager.request` repeats it. Kept for parity with `content_filter_manager`, which this manager is modelled on.
- Mapping a 404 on `DELETE /nat/{id}` to the endpoint hint. A 404 on a per-id path is ambiguous with "rule gone"; the id is resolved against the list first, so a stale id already surfaces as `UniFiNotFoundError`.


sirkirby (changes requested):
Tyler, I reviewed `4ea1b8d54924a1f9b96bbd382b92f7ec5e2c3e2d` and am requesting a P2 cache-preservation correction.

[Update and toggle](https://github.com/sirkirby/unifi-mcp/blob/4ea1b8d54924a1f9b96bbd382b92f7ec5e2c3e2d/packages/unifi-core/src/unifi_core/network/managers/nat_manager.py#L124-L175) take the full replacement document from the cached list. I reproduced a list followed by an external disable/address change, then a description-only update: the PUT restored the old address and re-enabled the rule. Please fetch fresh state before merging a replacement or calculating a toggle.

The same finding includes invalidation after an uncertain mutation. PUT/DELETE exceptions leave the old cache valid. With production ConnectionManager caching and simulated controller I/O, I reproduced a committed PUT followed by a reply timeout returning the old document on the next get; a committed DELETE followed by a timeout likewise returned the deleted rule. Please invalidate after attempted mutations with ambiguous outcomes, as create already does.

The 148 focused tests passed, and a read-only hardware check listed/projected one SNAT rule. The state-loss and timeout cases were isolated simulations, with no hardware writes or production incident established. Please add regressions for these transitions.

sirkirby (commented):
Tyler, I accept a bounded Core-first NAT slice for the DNS-redirect need in [#216](https://github.com/sirkirby/unifi-mcp/issues/216). This is a concrete controller capability, and the published-Core dependency gate justifies the split. It does not complete the user-facing feature by itself: please link the downstream MCP/API/skill delivery plan, including confirmation, resolver exemptions, post-write verification and package floors. Unverified IPv6, PPPoE and exclude semantics should remain outside the support claim.

Your Network 10.6.102 disabled-rule CRUD discovery is relevant real-controller evidence. Please attach or summarize the existing sanitized per-variant record—submitted shape, fresh read-back and cleanup—tied to the tested revision. I do not need discovery repeated merely because you performed it. Our independent hardware check covered one SNAT read/projection; it did not cover DNAT writes or DNS interception.

After the cache correction is verified, any missing CRUD preservation check should use disposable lab rules with fresh read-back and cleanup. The downstream DNS-redirect feature will also need a test-client query aimed at another resolver reaching the approved resolver, while that resolver's own upstream continues working without a loop. Persisting a disabled rule is useful schema evidence but does not prove that traffic behavior. No enabled production rule changes are requested while the correctness issue remains.

Acceptance follow-up at `d9156c5c5171a8792b571c2ad8770d4a0a44e50e`.

tmowbrey:
Cache-preservation finding addressed, regressions added, per-variant CRUD record attached. Tested SHA `09b7348097ea02423ad52cb5d6e73c2fcf72b424` (fix commit `ef343f7` + merge of upstream `main`).

**Fix** (`packages/unifi-core/src/unifi_core/network/managers/nat_manager.py`):
- L124-175 (update/toggle built the replacement from the cached list): `update_nat_rule` and `toggle_nat_rule` now read the rule through a refreshed list (`get_nat_rule(rule_id, refresh=True)`) before merging or calculating the flip, so an external disable or address change is carried into the PUT instead of reverted. `delete_nat_rule` keeps the cached read: it only resolves the id and builds nothing from the document.
- Invalidation after an uncertain mutation: one `_mutate` helper sends POST/PUT/DELETE and invalidates the cached list in `finally`, so a PUT or DELETE whose reply never arrived cannot leave the old document, or a deleted rule, being served by the next `get`. Create already did this; it now goes through the same helper.

**Regressions** (`packages/unifi-core/tests/network/managers/test_nat_manager.py::TestCachePreservation`, written first against the unfixed code): cached list then an external `enabled=false` + address change, description-only update → PUT carries the fresh address and `enabled=false`, request sequence `get, get, put`; toggle flips the fresh state, not the cached one; explicit-state toggle still uses the fresh document; PUT raising `TimeoutError` / `ResponseError` → cache invalidated (both parametrised); DELETE likewise; the read after a failed PUT goes to the controller. Core suite: 2048 passed; ruff and `make check-generated` clean.

**Per-variant CRUD record**, Network Application **10.6.102** (self-hosted path, aiounifi 95), run through this revision's `NatManager` on disposable rules that were disabled for their whole life and deleted in the same run. "Read-back" is a refreshed list. Interfaces were the site's LAN (`in_interface` / `network_conf_id`) and WAN (`out_interface`) network ids; translation targets were TEST-NET-1 addresses.

| Variant (submitted shape) | Created (disabled) | Fresh read-back equals every submitted key | Extra keys the controller added | Description update applied / other keys preserved | Toggle to disabled → read-back disabled | Deleted → absent on fresh list |
|---|---|---|---|---|---|---|
| DNAT DNS-redirect: `tcp_udp`, `port 53`, `ip_address`, `destination_filter ADDRESS_AND_PORT` inverted on the target, `source_filter NONE` | true | true | `_id`, `is_predefined`, `pppoe_use_base_interface`, `rule_index` | true / true | true | true |
| DNAT `exclude: true` with the same translation, destination not inverted | true | true | same four | true / true | true | true |
| DNAT with `source_filter NETWORK_CONF` (`network_conf_id`) | true | true | same four | true / true | true | true |
| SNAT: `protocol all`, `out_interface`, `ip_address`, both filters `NONE` | true | true | same four plus `exclude` | true / true | true | true |
| MASQUERADE: `protocol all`, `out_interface`, both filters `NONE` | true | true | same four plus `exclude` | true / true | true | true |

Site rule count before and after: 6 and 6; the six production rules are byte-identical before and after; no disposable rule left behind; no rule was enabled at any point.

**Downstream plan** for the MCP/API/skill delivery (confirmation, post-write verification, resolver exemptions, package floors; IPv6/PPPoE/`exclude` semantics outside the support claim): sirkirby/unifi-mcp#216 (comment 5557341730).

Script and output stay in my session scratchpad; no address, id or credential appears above.

sirkirby (commented):
Tyler, I re-reviewed `09b73480`. The cache-preservation change resolves the prior request: update and toggle fetch fresh state, and the shared mutation helper invalidates cached rules in finally after successful or uncertain POST/PUT/DELETE outcomes. Independent focused model/manager/exception checks passed: 156 tests, including the fresh-state and failure-invalidation regressions.

The revision-specific five-variant disabled-rule record now supplies the requested submitted/read-back/preservation/cleanup evidence, and the downstream plan is linked from #216. I accept that evidence for the bounded Core-first scope; it does not establish enabled DNS interception or untested IPv6/PPPoE/exclude semantics.

No new source blocker found. This can advance into the Core foundation integration checks. The later MCP/API feature still needs its confirmation, read-back, floor and traffic-behavior gates. No additional contributor correction is requested by this reassessment.

sirkirby (approved):
Tyler, I approve the corrected head `09b73480`. The fresh-state partial update/toggle and uncertain-mutation cache corrections are verified. Your exact-head disabled CRUD/preservation/cleanup record supplies the mutation evidence, and our installed controller read/projection passed. This delivers Core NAT support only; downstream app tools and DNS traffic behavior remain separate.

PR #653, opened 2026-09-05
fix(protect): preserve sensor summary fields in protect_list_sensors

### LINKED ISSUE #621: Protect: preserve existing sensor summary fields in protect_list_sensors
## Problem

`SensorManager._format_sensor_summary()` already produces the existing Protect sensor summary fields, but `protect_list_sensors` projects each record through a narrower `unifi_core.protect.models.sensors.Sensor` model. Most fields are silently discarded before the MCP response is returned.

This was surfaced while triaging #523, but it affects ordinary Protect sensors independently of UP-AirQuality support.

## Reproduction evidence

Verified independently on current `main`:

- Feeding the real 19-key manager output shape through `sensor_from_controller(...).model_dump(exclude_none=True)` returns only `id`, `name`, `type`, and `motion_detected_at`.
- A reference-controller run with six sensors reproduced the same projection loss. The manager returned battery, state, connection, firmware, last-seen, mount, activity, timestamp, and stats fields; the tool projection discarded them.
- The current model tests use hand-written keys and types that do not match the manager output, so they do not exercise the production seam.

## Scope

- Align the Protect sensor read model with the fields and nested shapes already emitted by `_format_sensor_summary()`.
- Preserve existing fields including model/state/connection data, firmware and last-seen, mount/activity state, event timestamps, battery details, environmental stats, and camera association.
- Replace the hand-written projection fixture with a production-shaped regression test that passes a manager summary through the tool boundary.
- Verify API consumers of the shared model and regenerate any affected GraphQL/OpenAPI/reference artifacts.

## Out of scope

Adding UP-AirQuality-specific PM/AQI/VOC/CO2 readings remains tracked by #523. Those fields require a sanitized real-device payload before their names and shapes can be mapped safely.

## Acceptance criteria

- [ ] `protect_list_sensors` preserves every supported field currently emitted by the manager.
- [ ] Battery and `stats` use the same names and nested types at the manager/model/tool boundaries.
- [ ] Regression tests use the actual manager output shape and fail on the current implementation.
- [ ] Existing update-field mutability and read-only-field invariants remain correct.
- [ ] Shared API consumers and generated artifacts are updated when affected.
- [ ] `make pre-commit` passes.
- [ ] A targeted reference-controller comparison confirms the fields survive the MCP tool projection.

Related diagnosis: https://github.com/sirkirby/unifi-mcp/issues/523#issuecomment-5541038214

tmowbrey:
## Summary

Closes #621.

`protect_list_sensors` ran each `SensorManager._format_sensor_summary()` record through `Sensor.model_dump(exclude_none=True)`, and the `Sensor` read model only knew `id`, `mac`, `name`, `type` plus three status fields the manager never emits. Fifteen of the manager's nineteen keys were dropped before the MCP response.

This aligns the `Sensor` model with the manager's field names and nested shapes:

- Adds `model`, `state`, `is_connected`, `firmware_version`, `last_seen`, `mount_type`, `is_motion_detected`, `is_opened`, `open_status_changed_at`, `alarm_triggered_at`, `leak_detected_at`, `tampering_detected_at`, `camera_id`.
- Adds nested `battery: SensorBattery {percentage, is_low}` and `stats: SensorStats {light, humidity, temperature: SensorStatReading {value, status}}`, the same names the manager emits.
- Removes `battery_status`, `humidity_status`, `light_status`, which no producer populated.
- Timestamp fields use an `IsoTimestamp` annotated type so datetime coercion lives on the model, not only in `from_controller`.
- `from_controller` projects every model field from the summary generically.

`MUTABLE_FIELDS` stays `{"name"}`; every new field is read-only, so `protect_update_sensor_settings` rejects them with the existing read-only message.

Out of scope: UP-AirQuality PM/VOC/CO2/AQI readings stay tracked by #523. No air-quality fields were added.

## Type of change

- [x] `fix:` bug fix
- [ ] `feat:` new feature

## Scope

- [ ] Network MCP server
- [x] Protect MCP server
- [ ] Access MCP server
- [ ] Relay sidecar
- [ ] Cloudflare Worker / CLI
- [ ] API server
- [x] Shared packages
- [ ] Plugin packaging
- [ ] Documentation

## Checklist

- [x] I updated documentation for user-visible changes. (None needed: the tool description and the generated `protect-tools.md` reference already describe these fields; no doc names the removed ones.)

## Testing

Regression tests were written first and failed on `main` before the model change:

- `packages/unifi-core/tests/protect/models/test_sensors.py::TestSensorProjectionFromManagerSummary` feeds the real `SensorManager._format_sensor_summary()` output through `from_controller(...).model_dump(exclude_none=True)` and asserts equality. On `main` this returned only `id`, `name`, `type`, `motion_detected_at`.
- `apps/protect/tests/unit/test_devices.py::TestProtectListSensorsTool::test_success` now builds the summary through `SensorManager.list_sensors()` and asserts the tool response carries every non-null key.
- The hand-written fixture in the model tests was replaced with the production-shaped summary.

Commands run from the repo root:

```
make pre-commit                                             # passes
uv run --package unifi-core pytest packages/unifi-core/tests/protect -q          # 438 passed
uv run --package unifi-protect-mcp pytest apps/protect/tests -q                  # 512 passed
uv run --package unifi-api-server pytest apps/api/tests/unit/test_cross_layer_symmetry.py \
  apps/api/tests/serializers/test_protect_serializers.py \
  apps/api/tests/routes/resources/test_protect_cameras_lights_sensors.py \
  apps/api/tests/graphql/test_sdl_drift.py -q                                    # 94 passed
make -C apps/protect manifest                               # no diff; tool signature unchanged
```

**Not run: the reference-controller comparison.** I do not have a Protect controller with sensors available from this environment, so the last acceptance criterion in #621 is unverified here. Please run `protect_list_sensors` against the reference controller and confirm the battery, state, connection, firmware, last-seen, mount, activity, timestamp and stats fields survive.

## Notes for reviewers

- **API server:** verified the `apps/api` consumers. The Strawberry `Sensor` type in `apps/api/src/unifi_api/graphql/types/protect/sensors.py` and the REST route read the manager dict directly and do not import the pydantic model, so `schema.graphql` and the OpenAPI output are unchanged and the cross-layer symmetry test still passes. That GraphQL type still exposes a flattened `battery_status` / `humidity_status` / `light_status` projection; widening it is a separate API-contract decision I left alone.
- `mac` stays on the model even though the manager does not emit it. It is dropped by `exclude_none` and keeps the specific read-only rejection message in `protect_update_sensor_settings`.
- The same model-narrower-than-manager pattern exists for `protect_list_lights` and `protect_list_cameras` (the `Light` model lacks `type`, `is_connected`, `firmware_version`, `last_seen`, `camera_id` and the nested settings the manager emits). Not touched here; happy to file a follow-up issue if you want the same treatment.

### Deferred items

- `protect_list_lights` and `protect_list_chimes` have the same projection loss: the `Light` model drops 11 of the manager's 17 keys and reads `led_level` / `sensitivity` / `duration_seconds` at top level while the manager nests them under `device_settings`; `Chime` drops `camera_ids`, `last_ring`, `last_seen`. Out of scope for #621; I can open a follow-up issue if you want the same fix.
- The GraphQL/REST `Sensor` type still returns the eight-field flattened projection, so the same tool name returns more over MCP than over the API. Separate API-contract decision.
- `protect_list_sensors` validates the whole list in one `try`, so one sensor failing model validation fails the call. Pre-existing pattern shared by every list tool; uiprotect's types match the model exactly, so there is no current trigger.

### Considered & declined

- `extra="forbid"` on `SensorBattery` / `SensorStats` so a future nested manager key raises instead of being ignored. Declined: every other Protect read model ignores extras, and a forbid would turn a manager addition into a whole-tool outage instead of a missing field. The `_manager_summary` fixture already raises if the manager reads a new sensor attribute.
- Making `IsoTimestamp` reject or null out non-datetime, non-string values. Declined: `_stringify_dt` is pre-existing code moved onto the type, and no producer reaches that branch (the manager stringifies first).
- `StrictBool` / `StrictInt` on the read model fields. Declined: uiprotect's types already match, and the read model follows the sibling models' non-strict convention.


sirkirby (commented):
Tyler, I reviewed `3b93ccf7c6743ff712e66a6ccb6e6bb8f9c7eccc` and found no functional regression in the MCP sensor projection change.

The [expanded model](https://github.com/sirkirby/unifi-mcp/blob/3b93ccf7c6743ff712e66a6ccb6e6bb8f9c7eccc/packages/unifi-core/src/unifi_core/protect/models/sensors.py#L68-L87) preserves the manager's battery, readings, connection and timestamp fields through `protect_list_sensors`. All 179 focused tests passed. A read-only comparison on six live sensors also confirmed that every non-null manager field survived the tool projection.

The separate REST/GraphQL sensor projection still has the previously existing field loss; I am keeping that as follow-up work outside this fix. Delivery still needs a Protect dependency floor that includes the corrected Core model and the final integrated checks.

sirkirby (approved):
Tyler, I approve this revision. I verified the installed candidate against six live Protect sensors on Protect 7.2.105: every non-null SensorManager summary field survived. Zero and false values also passed the focused combined probe. The installed Protect safe smoke passed 59 checks and cleaned up its temporary resource, and the full repository gate passed. This acceptance covers the MCP sensor projection; the separate API projection is unchanged. The release will raise the Protect Core dependency floor.

PR #654, opened 2026-09-05
feat(shared): read credentials from a file

### LINKED ISSUE #604: feat: let the server fetch its own credential (UNIFI_PASSWORD_FILE / _COMMAND) so it need not live in the MCP client's environment
The plugin's `.mcp.json` sources the password by interpolating from the client's environment:

```json
"env": { "UNIFI_NETWORK_PASSWORD": "${UNIFI_NETWORK_PASSWORD:-}" }
```

That requires the secret to be exported into the MCP **client's** process before it starts. On Claude Code the client spawns arbitrary child processes for its shell tool, so every one inherits it. I found my UniFi password in the argv/environment of a `vite` dev server in an unrelated repository — readable by any local process listing, and it would appear in crash dumps and process-tree logs.

This isn't a config mistake. Given the interpolation, there is no way to supply the credential *without* putting it in the client's environment.

### Mitigations that don't work

- **macOS Keychain / password manager at shell startup** — changes where the secret rests, not who inherits it. It still ends up exported into the client.
- **A wrapper script as the server `command`** — works, but a user-level `mcpServers` entry *coexists with* rather than shadows the plugin's (different tool namespaces: `mcp__plugin_unifi-network_unifi-network__*` vs `mcp__unifi-network__*`), so the plugin has to be disabled entirely — losing its five bundled skills.

### Proposal

Either would fix it; the first is minimal.

1. **`UNIFI_PASSWORD_FILE`** — read the credential from a path (the Docker/systemd secrets convention). Trivial, no shell-out.
2. **`UNIFI_PASSWORD_COMMAND`** — run a command and take stdout, so Keychain / `pass` / 1Password / Vault all work:

   ```
   UNIFI_PASSWORD_COMMAND="security find-generic-password -a $USER -s unifi-mcp -w"
   ```

With either, the `.mcp.json` shipped by the plugin can reference a **non-secret** indirection, and the credential only ever exists inside the server process.

Given #348 and #351 already treat secret exposure as in scope, this seems consistent with that direction. Happy to open a PR if you'd like a particular shape.

tmowbrey:
## Summary

Closes #604. Scope narrowed to file-backed credentials per the review; the `_COMMAND` provider is split out and will come as a separate PR with the platform / executable-resolution / process-lifecycle contract and the three fixes you found.

The plugin `.mcp.json` interpolates the password from the MCP client's environment, so the secret has to be exported into the client process and every child it spawns inherits it. This adds one indirect spelling per credential so the value only ever exists inside the server process:

- `UNIFI_<SERVER>_PASSWORD_FILE` (and `UNIFI_PASSWORD_FILE`): path to a regular file holding the value. UTF-8, BOM tolerated, trailing newlines dropped, at most 64 KiB.
- `UNIFI_<SERVER>_API_KEY_FILE` (and `UNIFI_API_KEY_FILE`): the same for the API key.

Resolution lives in one function, `resolve_env`, in `unifi-mcp-shared`'s `load_server_config`, so Network, Protect and Access pick it up with no per-app config change.

Everything fails closed with exit code 6 and the variable name in the log, never the file contents:

- Both spellings at the same precedence level (`UNIFI_NETWORK_PASSWORD` and `UNIFI_NETWORK_PASSWORD_FILE`) is refused as ambiguous rather than silently picking one.
- Missing, non-regular, unreadable, oversized, non-UTF-8, empty or multi-line file.
- An empty variable counts as unset, so the plugin's `"${UNIFI_NETWORK_PASSWORD:-}"` interpolation does not shadow the shared level.

**Trust boundary.** The app bootstraps load a `.env` from the working directory, and MCP clients set that to the user's project. Each app calls `snapshot_process_env()` before `load_dotenv()`; `resolve_env` honours `_FILE` only for variable names that were in the real process environment, so a hostile repo's `.env` cannot point the server at an arbitrary file. A `.env` can still supply a plain value, as before. A per-app subprocess test asserts a `.env`-supplied `_FILE` is refused and its contents never reach the log.

Also fixed while here: the `DEBUG` line "Applying env overrides" logged the merged dict with the password and API key in clear; it now goes through `redact_sensitive_fields`. Every env override is escaped so `${` stays literal (a generated password survives, and a `.env` cannot use `UNIFI_SITE=${oc.env:UNIFI_PASSWORD}` to pull the password into a field that reaches tool output). A secret equal to `???` (OmegaConf's missing marker) is refused rather than silently emptied. A secret file readable by group or others gets a warning.

Plugin `.mcp.json` and `plugin.json` pass `UNIFI_<SERVER>_PASSWORD_FILE` through as a self-mapping (Access also `API_KEY_FILE`, since it already forwards `API_KEY`). The setup skills offer the file form after collecting the password.

## Type of change

- [x] `feat:` new feature

## Scope

- [x] Network MCP server
- [x] Protect MCP server
- [x] Access MCP server
- [x] Shared packages
- [x] Plugin packaging
- [x] Documentation

## Checklist

- [x] I updated generated manifests with `make manifest` when changing MCP tools. (No tool signature changed; `make check-generated` reports no drift.)

## Testing

See the evidence comment for the file-source smoke (server started over stdio from the plugin entrypoint with the password and API key supplied only through `_FILE`, one session-auth read and one API-key read, three failure modes, environment-isolation booleans).

```
uv run --package unifi-mcp-shared pytest packages/unifi-mcp-shared/tests     468 passed
uv run --package unifi-<app>-mcp pytest apps/<app>/tests/unit/test_bootstrap_*  4 passed x3
make check-generated                                                           no drift
```

## Notes for reviewers

The Shared change ships here; the three app floors move once you cut Shared, per the #657 pattern (or split, as asked once in #658).


sirkirby (changes requested):
Tyler, I reviewed `3bfbccb84ce6993e9cb0f52859f58b5d82d21fa0` and am requesting changes for three independently reproduced helper-boundary issues:

- **P1 — use a trusted helper working directory.** The [subprocess call](https://github.com/sirkirby/unifi-mcp/blob/3bfbccb84ce6993e9cb0f52859f58b5d82d21fa0/packages/unifi-mcp-shared/src/unifi_mcp_shared/bootstrap.py#L117-L125) freezes the environment but inherits the project directory. Through the real Network bootstrap, an absolute Python interpreter running a trusted module-based helper instead loaded a same-name project module, executed its marker, and accepted its output as the password. Please define a trusted working directory and executable-resolution contract, and keep a real startup shadow-module regression. This applies to helpers with cwd-sensitive resolution; I have not established the same behavior for the documented `security` or `pass` commands.
- **P1 — suppress arbitrary helper stderr.** The [failure branch](https://github.com/sirkirby/unifi-mcp/blob/3bfbccb84ce6993e9cb0f52859f58b5d82d21fa0/packages/unifi-mcp-shared/src/unifi_mcp_shared/bootstrap.py#L139-L147) copies the stderr tail into an ERROR log. A failing helper that printed a synthetic password only to stderr exposed the complete value in the captured log. Please log fixed operation/status context without arbitrary helper text, and test secret sentinels on both output streams.
- **P2 — clean up helper descendants on timeout.** The [timeout path](https://github.com/sirkirby/unifi-mcp/blob/3bfbccb84ce6993e9cb0f52859f58b5d82d21fa0/packages/unifi-mcp-shared/src/unifi_mcp_shared/bootstrap.py#L117-L135) terminates only the direct child. With the timeout shortened for the probe, a descendant wrote its marker after the resolver had exited with code 6. Please own and terminate the process tree using the appropriate platform mechanism, with a real descendant-cleanup regression.

The functional resolver/startup suites passed; the broad Network run had one environment failure because the isolated image lacked `git`. These security/lifecycle probes ran on Linux without credentials or network access. Windows cleanup, real credential-provider behavior, and shared/app/plugin release alignment remain delivery checks.

sirkirby (commented):
Tyler, #604 establishes a valid credential-isolation problem, including the limitation of adding a wrapper alongside the bundled plugin. I support file-backed credentials as the minimal first feature. Native Keychain/password-manager access is also a valid use case, but I want `_COMMAND` separated into a provider feature with an explicit supported-platform, executable-resolution and process-lifecycle contract. File support can address the immediate inherited-secret problem without making arbitrary command execution part of its acceptance. Please split those scopes; this is in addition to the security fixes already requested.

Your live-testing paragraph describes `load_config()` producing the credential. That does not yet establish controller authentication or a successful MCP call. For file-source acceptance, please show the installed plugin starting with the direct credential removed, authenticating to a real supported controller, and completing one read-only tool call. Include an API-key source if that capability remains advertised. Report only whether the credential stays out of the client/unrelated child environment, never the environment contents. Include versions, platform and tested commit, plus a missing/unreadable-source failure with no secret-bearing logs.

For the separate command-provider proposal, please demonstrate the actual provider in the intended startup context through authentication and a read-only result, with a representative provider-failure case. Either you or a maintainer can supply these sanitized observations; no controller mutation is needed.

This is a scope and acceptance request for current proposal `0803504f`. The earlier security review was on `3bfbccb8`; it does not establish the status of fixes on the new head.

Acceptance follow-up at `0803504f96d94a3823383db95fcc59e30fab30c8`.

tmowbrey:
Scope narrowed to `_FILE` only, as requested. Tested head: `7a074c5` (rebased onto `main` `bc09e3c`).

**Against each finding**

- `_COMMAND` split off: `_run_secret_command`, its constants, docs, plugin env self-mappings, setup-skill wording and 15 tests are gone. The three helper-boundary findings (trusted cwd, stderr copied into the ERROR log, descendant cleanup on timeout) belong to that code and will be addressed on the provider branch with the contract you asked for, not here.
- `_FILE` kept as accepted: `resolve_env` → `_read_secret_file`, snapshot trust boundary unchanged. The per-app startup regression now proves a `.env`-supplied `_FILE` is refused (exit 6) and its contents never reach stderr.
- PR body rewritten for the narrowed scope.

**File-source smoke** — Linux x86_64, Python 3.13.14, Network Application 10.6.102 (UniFi OS Server), this head. The server was started from the plugin's own entrypoint (`unifi-network-mcp`, eager tool registration) over stdio by an MCP client (`mcp` Python client) whose environment held `UNIFI_NETWORK_HOST/USERNAME/PORT` and the two `_FILE` paths (mode 600) and nothing else credential-shaped. Booleans only:

| check | result |
|---|---|
| MCP client environment contains the password or API-key value | false |
| an unrelated child spawned by the client contains either value | false |
| server process environment (own descendants, 2 processes) contains either value | false |
| server process environment contains the `_FILE` path | true |
| `initialize` + `unifi_get_system_info` (session auth) succeeds | true |
| `unifi_list_firewall_zones` → `unifi_get_firewall_policy_ordering` (API-key source) succeeds | true |
| either value in tool output or in 29 captured server stderr lines | false |
| `_FILE` → missing path: exit 6, variable named, no secret in output, 0.9 s | true |
| `_FILE` → not a regular file (directory): exit 6, variable named, no secret, 0.9 s | true |
| `_FILE` and plain value both set: exit 6, both names in log, no secret, 0.9 s | true |

Not an installed wheel: the worktree's package via the same console script, since the Shared change is unreleased. Happy to repeat on the published wheel after the Shared release.

Suites: Shared 468, per-app bootstrap 4 × 3, `make check-generated` clean.

sirkirby (commented):
Tyler, I reviewed the FILE-only revision `7a074c56`. Removing COMMAND resolves the earlier executable/cwd/stderr/descendant findings for this PR; those are no longer part of this implementation. I accept this narrower feature and found no new FILE blocker in the reviewed paths.

I independently passed 62 focused resolver/bootstrap tests. A separate three-app audit-hook probe confirmed that a dotenv-selected FILE path is rejected with exit 6 before the selected secret file is opened. Your SHA-tied session-auth and API-key controller reads provide useful behavioral evidence, and I appreciate the explicit statement that they used worktree packages.

The existing delivery hold remains: `load_process_env` is a new unconditional Shared import, so publish Shared and raise Network/Protect/Access floors together. Then verify the installed app/plugin FILE flow, a harmless session-auth read and the advertised API-key read, plus a source-failure case and sanitized environment/log checks. The worktree console entrypoint does not yet prove the bundled installed plugin configuration. I will coordinate the foundation/floor work; no further provider split is needed here.

sirkirby (approved):
Approving, and the delivery hold is discharged — I built wheels from this branch and ran the installed-artifact verification the hold was waiting on, rather than deferring it to after publish.

Rebased onto current `main` (it was 22 behind; `core/v0.4.47` and the floor bumps landed today). Clean rebase, no conflicts.

## Why the worktree could never have proved this

Worth recording, because it explains the hold rather than just satisfying it. `load_process_env()` calls `load_dotenv()`, whose `find_dotenv()` walks up from **`bootstrap.py`'s own location**. From a source checkout that reaches the repository's own `.env` — so every worktree run silently resolves the repo's real credentials and the trust boundary can't be observed at all. My first three probe runs were contaminated exactly that way and reported "both set" for variables I had never set. Your docstring already says the walk "under an installed wheel never reaches the user's project"; the corollary is that the guarantee is unobservable until it is installed.

## Installed-artifact verification

Wheels built from this branch, installed into a clean venv **outside** the repository (`in-repo=False` confirmed for all three packages), run with a scrubbed environment and a synthetic home:

**Supported path**
```
resolved: <the file's contents>
value NOT in os.environ: True
```

**The attack the feature must not open** — `_FILE` supplied only by a project `.env`:
```
[credentials] UNIFI_NETWORK_PASSWORD_FILE was supplied by a .env file, not by the
environment the server was started with; credential indirection is only honoured
from the process environment.
exit=6
```

**Unresolvable source** — missing file: exit 6, named path, no traceback.

## End-to-end through the installed app entrypoint

The piece the hold specifically called out. Installed `unifi-network-mcp` + Shared + Core, credential supplied *only* via `UNIFI_NETWORK_PASSWORD_FILE`, against the reference controller:

```
password var in environ : False
secret value in environ : False
controller login via _FILE credential: True
authenticated read returned rows: 1
```

Session auth completes and an authenticated read succeeds while the secret is in neither the variable nor any value in the environment children would inherit. That is #604's actual requirement, demonstrated rather than argued.

The advertised API-key path behaves identically (`UNIFI_NETWORK_API_KEY_FILE` resolves, absent from the environment), and a full DEBUG-level run contains **zero** occurrences of the secret.

## Gates

Repo-wide ruff clean, no generated-artifact drift, no f-string loggers, no issue references. No reverse imports into Shared (Procedure J-1). Suites: Core 2240, Shared 477, Network 1148, Protect 514, Access 223, API 1425.

The 0600-permission warning, the empty-value and multi-line refusals, and the `~`-without-home case are all good defensive touches — a second line silently accepted would have surfaced later as an unexplained 401, which is the failure that would have cost someone an afternoon.

## Release

`load_process_env` is a new unconditional Shared import, so this ships as Shared first, then the Network, Protect and Access floors raised together — I'm running that now.

Closes #604.

PR #659, opened 2026-09-06
fix(core): resolve client lookups via /stat/user past the /rest/user cap

tmowbrey:
## Summary

Core half of #642 (refs #632), split out so it can be released before the Network floor is raised, as requested in the #642 review.

`get_client_details` scanned `/stat/sta` and `/rest/user`; the controller caps `/rest/user` at 3,000 rows, so a client past that window was reported as `client '<mac>' not found` although the controller has its record. When the `/rest/user` scan misses a MAC-shaped identifier, `ClientManager` now consults the authoritative per-MAC `GET /stat/user/<mac>` (200 with the user-table record including `_id` for a known client; HTTP 400 `api.err.UnknownUser` for an unknown one). Callers that only need existence (block, unblock, kick, guest authorization, `StatsManager` identifier resolution) pass `existence_only=True` and stop at a live `/stat/sta` record.

Outcomes of the per-MAC request:
- `api.err.UnknownUser` is an authoritative not-found.
- HTTP 404 (a controller that does not serve the endpoint) leaves the lists' verdict in place. **The status is read from the status-bearing segment of aiounifi's `ResponseError` message** (`response_status()`), never from the URL or body: aiounifi raises one class for 404 and 429 with no status attribute and formats `Call <url> received <status>[: <body>]`, so the parser anchors on `Call <no-whitespace-url> received <3 digits>`. A 429 from `controller404.example`, or a 429 whose body mentions 404, stays undetermined.
- Both list endpoints failing still re-raises the outage error.
- Any other per-MAC failure (transport, 429, auth, malformed reply) raises `UniFiOperationError` saying existence could not be determined.

`ConnectionManager.request` now logs every failure through one helper that masks MAC-shaped tokens and the configured credentials in the path, the exception text and the rendered traceback. A controller-reported `api.err.*` code, or a 404 answer, is logged at INFO on a read and WARNING on a write, without the body (which echoes the requested MAC). The same status parser replaces the `"received 429" in message` check in the auth circuit. `unifi_core.mac` gains `canonical_mac` and `mask_macs`.

The app half (Network `get_top_clients` with `existence_only`, API route mapping `UniFiOperationError` to 502) stays on #642, rebased on this branch; its `unifi-core[network]>=0.4.44` floor bump is staged locally and will be pushed once PyPI serves the Core release.

## Type of change

- [x] `fix:` bug fix

## Scope

- [x] Shared packages

## Checklist

- [x] I updated generated manifests with `make manifest` when changing MCP tools. (No tool surface changed; `make generate` produced no diff, `make check-generated` passes.)
- [x] I updated documentation for user-visible changes. (No user-facing doc describes the lookup path.)

## Testing

```
uv run --package unifi-core --all-extras pytest packages/unifi-core/tests -n auto   # 1961 passed
uv run --package unifi-network-mcp pytest apps/network/tests/unit -n auto           # 1083 passed (workspace Core)
uv run --package unifi-api-server pytest apps/api/tests -n auto                     # 1371 passed
uv run --package unifi-protect-mcp pytest apps/protect/tests -n auto                #  513 passed
uv run ruff format --check . && uv run ruff check . && make check-generated && make docs-test
```

Regressions written first against the review finding, in `packages/unifi-core/tests/network/managers/test_client_manager_stat_user.py`: 429 from a host named `controller404.example` → `UniFiOperationError`; 429 whose body mentions 404 → `UniFiOperationError`; 404 from a host named `controller429.example` → not-found (list verdict); `ResponseError` without a status segment → undetermined; transport error with 404 in the URL → undetermined with cause; `response_status` table (404/429/503, body containing `received 404`, `4040`, missing prefix, other exception types → `None`). The rest of that module is the per-MAC behaviour suite moved from the app tests so it lives with Core. `test_connection_manager_request_logging.py` covers the masked failure lines, credential masking, the 404-at-INFO rule and the auth-error paths keeping ERROR + diagnostics.

Live evidence on the affected site is on #642 (tied to the app head, which includes this commit).

## Notes for reviewers

- `apps/network/tests/unit/test_mac_case_insensitivity.py` changes its fixture's `conn.request` return from `None` to `[]` because the per-MAC fallback now runs when both lists miss; that is the only app-side change here and it is needed for the workspace suite to pass with this Core.
- Considered and declined: reading a status from `exc.args`, which aiounifi does not populate; a logging `Filter` for MAC masking (broader than this change; every failure line this path emits now goes through one helper).
- `refresh_handler`'s retry log still prints the raw exception; the credential-redaction PR (#645/#648 shared correction) owns that line.

### DIFF
--- a/apps/network/tests/unit/test_mac_case_insensitivity.py
+++ b/apps/network/tests/unit/test_mac_case_insensitivity.py
@@ -44,7 +44,8 @@ def mock_connection():
         return await getattr(_c.controller, name).update()
 
     conn.refresh_handler = _refresh_handler
-    conn.request = AsyncMock(return_value=None)
+    # An empty list answers the per-MAC /stat/user fallback as a soft miss.
+    conn.request = AsyncMock(return_value=[])
     return conn
 
 
--- a/packages/unifi-core/src/unifi_core/mac.py
+++ b/packages/unifi-core/src/unifi_core/mac.py
@@ -29,7 +29,14 @@ turns into a request the controller may still reject, later and more quietly.
 import re
 from typing import Any, Optional
 
-__all__ = ["normalize_mac", "mac_equal", "looks_like_mac", "normalize_mac_list"]
+__all__ = [
+    "normalize_mac",
+    "mac_equal",
+    "looks_like_mac",
+    "normalize_mac_list",
+    "canonical_mac",
+    "mask_macs",
+]
 
 # Six hex pairs separated by ':' or '-', or twelve bare hex digits.
 _MAC_RE = re.compile(r"^(?:[0-9a-f]{2}([:-]))(?:[0-9a-f]{2}\1){4}[0-9a-f]{2}$|^[0-9a-f]{12}$")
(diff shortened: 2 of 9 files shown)


sirkirby (commented):
Tyler, I reviewed `cf269dfb` against current main. The per-MAC Core repair and package split are sound in the inspected paths, and the former broad `404` substring classification is corrected. I checked aiounifi 95's actual error format and independently passed 110 focused lookup/status/logging/MAC tests, including `404` in a non-404 URL/body, live-field preservation, malformed responses and mutation prechecks. No new source blocker found.

The controller record on #642 is tied to a head containing this exact Core commit and addresses the requested beyond-cap ID/live-field preservation and unknown control. I accept it as contributor-supplied targeted evidence; I did not repeat that controller discovery.

This can advance into the Core foundation integration checks. #642 remains dependent on publishing that Core and raising the consuming floor, followed by installed validation. Existing capped IP lookup and the GraphQL live-snapshot resolver are outside this per-MAC repair; this review does not imply those paths gain new behavior.

sirkirby (approved):
Tyler, I approve the corrected head `cf269dfb`. The status parser and per-MAC fallback corrections are verified. Installed real-controller requests preserved the user ID and all live fields, and the unknown control returned not-found. The cap omission was simulated locally; your beyond-cap hardware evidence remains separately attributed. Auth/rate errors remain distinct.

PR #660, opened 2026-09-06
fix(core): scrub credential values from controller errors before logs, rethrow and audit

tmowbrey:
## Summary

Shared correction for the credential error path reported on #645 and #648. A controller error that quotes the rejected request, or a transport error that quotes the login, carried the submitted secret into three sinks:

- `ConnectionManager.request` raw exception log (`connection_manager.py:777` at the reviewed heads)
- `SystemManager.update_settings` raw response / exception log (`system_manager.py:327-332`)
- API action route audit `detail=str(e)` (`routes/actions.py:287-297`)

What changed:

- `unifi_core.redaction` gains `collect_secret_values` (values under sensitive keys, using the existing `is_sensitive_key` vocabulary), `scrub_secret_values` (longest-first replacement; token-boundary matching for values under 4 chars) and `sanitize_exception` (rewrites `args`, message attributes, PEP 678 notes and the `__cause__`/`__context__` chain **in place**, so the exception type and identity survive and every later `str(e)` is clean).
- `ConnectionManager.request` scrubs the configured password plus every sensitive-keyed value in the request payload on all four error branches (aiounifi/aiohttp errors, unexpected errors, both re-auth retry paths) and `refresh_handler`'s retry error, before logging and before re-raising. Because every controller call flows through here, this covers reads, previews and writes for every manager.
- `SystemManager.update_settings` redacts by key and scrubs the rejected response before logging it, and scrubs the exception it re-raises. This is kept in the manager (not only the transport) so a fake-connection regression still holds.
- The API action route scrubs audit `detail` against the request's own sensitive argument values. This also covers schema-validation and translator messages that echo the offending value on the preview path.

No tool descriptions, manifests or catalogs changed (`make check-generated`: no drift).

## Type of change

- [x] `fix:` bug fix

## Scope

- [x] Network MCP server
- [x] API server
- [x] Shared packages

## Checklist

- [x] I updated generated manifests with `make manifest` when changing MCP tools. (No tool changes; `make check-generated` clean.)
- [x] I updated documentation for user-visible changes. (None user-visible.)

## Testing

Regressions written first (16 red on `bc09e3c`, all green on this head). Each drives the real manager with a fake aiounifi controller that fails while quoting a sentinel, and asserts the sentinel is absent from the MCP/API response, the captured log and the audit row:

- `packages/unifi-core/tests/network/managers/test_credential_error_sanitization.py` — real `ConnectionManager`: write errors (ResponseError / RequestError / aiohttp ClientError / unexpected), read error quoting the login password, chained `__cause__`, second `LoginRequired` after re-auth; real `SystemManager` + fake connection: write exception and rejected-response log.
- `packages/unifi-core/tests/test_redaction.py` — helper unit tests.
- `apps/network/tests/unit/test_credential_error_sanitization.py` — `unifi_update_snmp_settings` write and preview, `unifi_get_snmp_settings` read, through real `SystemManager` → real `ConnectionManager`.
- `apps/api/tests/test_action_credential_audit.py` — `POST /v1/actions/unifi_update_snmp_settings` write error and preview validation error, `unifi_get_snmp_settings` read error; audit `detail` read back from the DB.

```
uv run --package unifi-core --all-extras pytest packages/unifi-core/tests      # 1908 passed
uv run --package unifi-network-mcp pytest apps/network/tests/unit             # 1086 passed
uv run --package unifi-api-server pytest apps/api/tests                        # 1374 passed
uv run ruff check . && uv run ruff format --check . && make check-generated    # clean
```

Read-only smoke on an owned controller, this head, Network Application 10.6.102 (UniFi OS Server), entry points `SystemManager.get_system_info`, `SystemManager.get_settings("snmp")` and one `ConnectionManager.request` GET to a non-existent path so the error branch runs:

| check | result |
|---|---|
| connect + authenticate | pass |
| SNMP read | pass |
| negative GET raised `ResponseError` and hit the error log branch | pass |
| configured password present in 20 captured DEBUG log lines | false |

## Notes for reviewers

- Scrubbing is value-based (known secrets: configured password + sensitive-keyed payload values), not pattern-based, so ordinary error text is untouched.
- `sanitize_exception` mutates in place on purpose: callers keep `except ResponseError` matching and the diagnostics `log_api_request({"error": str(e)})` path becomes clean without touching it.
- #645 and #648 will be rebased onto this once it lands; the floor bump for the new `unifi_core.redaction` symbols follows the #657 pattern (a Core release, then app floors) unless you prefer a contributor split.

### DIFF
--- a/apps/api/src/unifi_api/routes/actions.py
+++ b/apps/api/src/unifi_api/routes/actions.py
@@ -7,7 +7,7 @@ from functools import lru_cache
 
 from fastapi import APIRouter, Depends, HTTPException, Request
 from pydantic import BaseModel, Field
-from unifi_core.redaction import redact_sensitive_fields
+from unifi_core.redaction import collect_secret_values, redact_sensitive_fields, scrub_secret_values
 
 from unifi_api.auth.middleware import require_scope
 from unifi_api.auth.scopes import Scope
@@ -118,6 +118,18 @@ def _shape_shared_read_result(
     if metadata:
         shaped["meta"] = metadata
     return shaped
+
+
+def _audit_detail(error: BaseException, args: dict) -> str:
+    """Return ``str(error)`` with every secret-keyed request value scrubbed.
+
+    The audit row is a durable sink. Core managers already scrub controller and
+    transport errors, and ``_validate_action_args`` keeps the submitted value
+    out of its own message, so nothing on either path is expected to carry a
+    credential here. This stays as the last barrier for a translator or
+    serializer message built somewhere else (#645, #648).
+    """
+    return scrub_secret_values(str(error), collect_secret_values(args))
 
 
 class ActionIn(BaseModel):
@@ -254,7 +266,7 @@ async def post_action(request: Request, tool_name: str, body: ActionIn) -> dict:
                 target=tool_name,
                 outcome="error",
                 error_kind="serializer_contract",
-                detail=str(e),
+                detail=_audit_detail(e, body.args),
             )
             await session.commit()
             raise HTTPException(
@@ -273,7 +285,7 @@ async def post_action(request: Request, tool_name: str, body: ActionIn) -> dict:
                 target=tool_name,
                 outcome="error",
                 error_kind="serializer_missing",
-                detail=str(e),
+                detail=_audit_detail(e, body.args),
             )
             await session.commit()
             raise HTTPException(
@@ -292,7 +304,7 @@ async def post_action(request: Request, tool_name: str, body: ActionIn) -> dict:
                 target=tool_name,
                 outcome="error",
                 error_kind=type(e).__name__,
-                detail=str(e),
+                detail=_audit_detail(e, body.args),
             )
             await session.commit()
             return {"success": False, "error": "Failed to execute action. Check the server audit log for details."}
--- a/apps/api/src/unifi_api/routes/actions.py
+++ b/apps/api/src/unifi_api/routes/actions.py
@@ -127,7 +127,7 @@ def _audit_detail(error: BaseException, args: dict) -> str:
     transport errors, and ``_validate_action_args`` keeps the submitted value
     out of its own message, so nothing on either path is expected to carry a
     credential here. This stays as the last barrier for a translator or
-    serializer message built somewhere else (#645, #648).
+    serializer message built somewhere else.
     """
     return scrub_secret_values(str(error), collect_secret_values(args))
 
--- a/apps/api/src/unifi_api/services/actions.py
+++ b/apps/api/src/unifi_api/services/actions.py
@@ -3,10 +3,13 @@
 from __future__ import annotations
 
 import inspect
+import re
+from collections.abc import Mapping
 from dataclasses import dataclass
 from typing import Any, Iterable
 
 from jsonschema import Draft202012Validator
+from pydantic import ValidationError
 from sqlalchemy.ext.asyncio import AsyncSession
 from unifi_core.confirmation import create_preview, delete_preview, preview_response
 from unifi_core.redaction import redaction_marker_paths
@@ -152,6 +155,10 @@ def _effective_preview_args(
     return effective
 
 
+# A constraint longer than this is a schema dump, not a diagnostic.
+_CONSTRAINT_REPR_MAX = 120
+
+
 def _validate_action_args(entry: ToolEntry, args: dict[str, Any]) -> None:
     """Validate raw REST args against the generated MCP input contract."""
     errors = sorted(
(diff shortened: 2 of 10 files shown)


sirkirby (changes requested):
Tyler, I reviewed `9ff0624d` with independent code and adversarial test passes. The shared correction is the right scope, but one P1 privacy defect remains.

At `packages/unifi-core/src/unifi_core/redaction.py:238-245`, scrubbing matches only the literal credential. A credential containing a backslash is represented differently by Python repr or JSON encoding, so the escaped value survives and remains fully recoverable. I reproduced this through real `ConnectionManager.request` logging/rethrow and through `SystemManager` into a persisted API audit row, using synthetic credentials and simulated controller responses.

The independent pass also demonstrated a controller-free case: a real API preview request with `community=[<synthetic backslash credential>]` fails schema validation, resolves no manager, and persists the escaped credential in `AuditLog.detail`. Thus this is not dependent on an assumed controller error format. The outward API error stays generic, but the persistent/log sink leaks.

Please prevent arbitrary credential-bearing diagnostic text reaching these sinks, or handle the actual encoded representations fail-closed. Add regressions through transport errors and schema-validation audit errors with backslash/quote/newline cases; checking only that the raw credential is absent is insufficient when the escaped form can be decoded. Fixed actionable error context is preferable where retaining arbitrary text cannot be made reliable.

The existing 59 focused Core/Network/API tests passed independently; the new probes expose missing coverage. I am keeping #645 and #648 held on this shared correction. The API's new Core imports also require the coordinated published Core floor before delivery; that is a separate acknowledged release gate.

tmowbrey:
Fixed on `21df031`, rebased onto current main (`1e61fe7`) and squashed — the previous head was red on its own after the rebase, because #659's `test_failure_logs_mask_the_configured_credentials` expects the free-text marker.

**The reported defect.** The scrub matched only the literal credential. The case is stronger than under-scrubbing: for a value holding a backslash the literal never appears in the message at all, so the scrub was a complete no-op and the value decoded straight back.

**Root cause for the controller-free case you demonstrated.** `apps/api/src/unifi_api/services/actions.py:166` built its message from jsonschema's `error.message`, which embeds the submitted instance — the value entered the text at construction, and every sink downstream had to defend by subtraction. It now uses schema-side facts only: the failing field and the constraint, plus the unknown or missing argument names, which are keys rather than values and which the MCP side already names the same way. Translator pydantic errors are restated the same way; `str(ValidationError)` truncates `input_value=` at about fifty characters, so a long credential survived as fragments that no value-matching scrub can find. The audit scrub stays behind both as a best-effort barrier for a message built elsewhere.

**A value is matched however it was written.** `repr`, `json.dumps` with and without `ensure_ascii`, `ascii`, a PHP-style escaper, `repr(bytes)` and percent-encoding each render the same credential differently, and a message can be escaped twice — aiounifi renders an already-JSON body with `repr(bytes)` on its 429 path. The scrub matches every rendering, closed under applying the emitters twice. Each form is a literal, which keeps matching exact and linear; matching a character at a time either misses the value or lets a caller choose how long the scan blocks the event loop.

**Three further sinks of the same defect, found while fixing it:**
- `ConnectionManager._sanitize_text` masked the configured login literally, so an escaped login reached `last_connection_error`, which is served back to callers in `Not connected to controller: ...`.
- `except LoginRequired:` never bound its exception, so both exits left the raw decoded response — which echoes the submitted record on a rejected write — as `__context__` of the `ConnectionError` the caller receives, rendered in full by every `exc_info=True` caller.
- `ClientResponseError.request_info` is a separate reference to the same object as `args[0]`, so rewriting `args` left the attribute, and `__str__`'s percent-encoded URL, unscrubbed.

**Two behaviour notes.** The login is masked on token boundaries, since its value is often an ordinary word; submitted values are matched wherever they appear. Both come from one pass over the union, longest first, so masking the login cannot chop a longer submitted value and leave its tail beside a mask the reader can identify. And classification now reads the exception before the scrub rewrites it: with the guard after it, a submitted value of `err` or `404` could turn a routine controller rejection into an ERROR with a traceback that logs the whole response body.

**Regressions.** Transport errors (`RequestError`, `aiohttp.ClientError`, `ClientResponseError` built with a real `RequestInfo`) and schema-validation audit errors, with backslash, quote, newline and non-ASCII in the credential. Each asserts the value is unrecoverable — literally and in every rendering, plus a decode round-trip — from the response body, the captured log, the rendered traceback, the re-raised exception, `request_info`, and the persisted `AuditLog.detail`. Checking only that the raw credential is absent is what let this ship the first time, so the oracle is hand-written rather than built from the production enumeration.

Local on `21df031`: `unifi-core` 2184, `apps/network` 1103, `apps/api` 1384, repo tooling 1331 pass; `ruff check` and `ruff format --check` clean; `make check-generated` reports no drift, and no tool signature changed.

Deferred, happy to take either as a follow-up if you want it: the MCP surface still returns pydantic's `input_value=` verbatim through FastMCP — same defect class, but it goes back to the peer that submitted the value, so it is transcript-confined rather than cross-principal. Out of the model deliberately: base64 and HTML entities, neither of which is emitted on these paths.

Unrelated, noticed while running the suites: `tests/test_smoke_mcp_metadata.py:23` imports `validate_tool_titles`, which #611 removed from `scripts/smoke_mcp_metadata.py`. It fails at collection on a clean `main`.

sirkirby (approved):
The P1 is fixed, and I verified it independently rather than from the added tests. Approving — I pushed the one remaining gate fix to your branch myself rather than send this round-tripping again.

## Verifying the P1

I re-ran my own adversarial battery against `21df031`'s successor rather than reuse the earlier evidence, since the head moved twice after my review. Seven synthetic credential shapes — backslash, embedded double quote, forward slash, non-ASCII, space, tab, plain opaque — through all eight emitters, then each of those escaped a second time. No leaks in any of the 112 combinations.

`sanitize_exception` holds the properties it claims: exception type and identity survive, `str(e)` is clean, and the `__cause__` chain is clean. `collect_secret_values` picks up nested `community` / `x_password` / `x_ssh_password` under a list and leaves the non-secret sibling alone.

`_encoded_forms` closing over two applications rather than one is the right call, and the docstring's note on what it deliberately does not cover — base64, HTML entities, mixtures no single emitter produces — is the correct place to draw that line. Matching literal forms rather than building a per-character alternation is also right; the alternation approach is where this class of scrub usually goes quadratic.

## Live validation

Independent run against my reference hardware, on the rebased head:

- **Network, `--phase safe`** — connected, 249 records: 160 ok, 54 skipped for undiscoverable args, 35 correctly held at the confirmation gate, **0 failures**. The three mutating lifecycles (DNS record, client group, firewall group) completed create → update → preview-delete → delete with all three disposable resources cleaned up. `unifi_update_gateway_settings` passed apply → verify-changed → verify-siblings-preserved → revert, with no sibling keys dropped.
- **`--phase api-actions`** — all seven action reads green, plus the confirmation preview control and both read-contract cases.

This mattered more than usual here: every controller call flows through `ConnectionManager.request`, so the blast radius of the scrub is the entire tool surface. Nothing regressed.

Suites on my machine: unifi-core 2222, API 1412, Network 1129, all passing. `make check-generated` reports no drift.

## What I changed on your branch

Two commits, both mechanical:

1. **Rebased onto current main.** The branch had picked up a merge commit and was four behind. The rebase was clean — no conflicts — and it collapses to your single commit.
2. **`chore(core,api): drop issue references from credential-scrub comments`** — the project bans issue and PR numbers in code, comments and test docstrings, and ten had crept in across the manager comments, the audit-route docstring and the three test modules. Those literals ship inside the wheel and go dangling as soon as the tracker moves. Each is replaced by the behaviour it was standing in for, so the comments still say why the scrub is there.

The three remaining `#NNN` references in `connection_manager.py` and `test_redaction.py` predate this branch; I left them rather than widen your diff, and will clean them separately.

One note for the follow-ups: `_encoded_forms` recomputes up to 72 forms per secret on every call, and `sanitize_exception` walks a whole cause chain. It is nowhere near hot enough to matter on an error path, so I am not holding the PR for it — but if the scrub ever moves onto a success path, memoize it first.

Issue #666, opened 2026-09-06
Reject negative limit in the tool layer instead of round-tripping to the controller


sirkirby:
Leftover from #564, which is closed now that #614 fixed the truthiness bug at all four sites. That PR was scoped to `limit=0`; negative values were raised in the issue as a separate question and are still unhandled.

A negative limit reaches the controller and fails there:

```
access_list_events {"topic": "unlocks", "limit": -1}
  -> Failed to list events: Proxy request failed: API code -2 CODE_PARAMS_INVALID ... Invalid parameters.
```

It fails safe, but it costs a round-trip to learn that, and the surfaced error names the internal proxy URL rather than the argument that was wrong.

## Suggested fix

Reject `limit < 0` in the tool layer with a named-argument message, so the caller gets "limit must be zero or greater" instead of a proxy error. The three event tools (`access_list_events`, `protect_list_events`, `unifi_list_events`) plus `unifi_recent_events` are the surfaces that take it.

Small and self-contained; the `limit is not None` guards from #614 give the regression tests a pattern to follow.

Reported by @rmaher001 in #564.

PR #667, opened 2026-09-06
fix(events): reject negative limits and handle Protect zero locally

be-student:
Negative limits on the four event tools reached the controller or Python slicing instead of returning a useful validation error. Network historical/buffered events, Protect events, and Access events now reject negative limits with an operation-specific message before manager access. Generated manifests and the API catalog describe the nonnegative limit contract.

Independent live testing also found that Protect rejects `limit=0` without a time range. The shared Protect manager now returns an empty list locally for zero on both the SDK and raw metadata/camera query paths. This preserves the advertised zero-result behavior for MCP and API consumers without an invalid or potentially unbounded controller query.

Addresses #666; close after the corrected app packages are published.

Validation: tool regressions verify negative limits never invoke their managers. Three new Core regressions verify zero never accesses the controller, including camera and metadata filters. Independently installed wheels passed negative/zero/normal-read checks against Network, Protect, and Access; the Protect zero failure was reproduced before the Core fix and passed afterward. Full `make pre-commit` and all 14 GitHub checks passed on 5157f025a7f7fa1825566b39de383e3021aac27e.

The Protect fix requires a new Core release, followed by a Protect/API Core floor bump before app publication. No app tag should advertise the corrected zero behavior while still allowing the older Core floor.

### DIFF
--- a/apps/access/src/unifi_access_mcp/tools/events.py
+++ b/apps/access/src/unifi_access_mcp/tools/events.py
@@ -64,10 +64,12 @@ async def access_list_events(
     ] = None,
     limit: Annotated[
         int,
-        Field(description="Maximum number of events to return (default 30)."),
+        Field(description="Maximum number of events to return; must be zero or greater (default 30)."),
     ] = 30,
 ) -> Dict[str, Any]:
     """List access events."""
+    if limit < 0:
+        return {"success": False, "error": "limit must be zero or greater"}
     logger.info("access_list_events tool called (topic=%s, door=%s, user=%s, limit=%s)", topic, door_id, user_id, limit)
     try:
         raw_events = await event_manager.list_events(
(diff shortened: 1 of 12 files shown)


sirkirby (approved):
@be-student, I reviewed the four negative-limit guards and added operation context to the errors plus regenerated manifests/catalog metadata.

Independent installed-wheel testing reproduced an additional Protect problem: the controller rejects zero without a time range. I fixed that in the Core manager so zero returns an empty list without accessing the controller, with regressions for both SDK and raw camera/metadata paths. Network, Protect, and Access then passed negative/zero/normal-read checks; the Network buffered-event tool also rejected a negative limit.

The full `make pre-commit` passed on 5157f025a7f7fa1825566b39de383e3021aac27e. I authorized the held contributor workflows, and all 14 checks are now green. I am merging the fix into the Core/app release batch and will close #666 after the app packages are published.

PR #670, opened 2026-09-07
feat(shared): add the credential command provider with a written contract

tmowbrey:
## Summary

Add the credential command provider split from #654. `UNIFI_<SERVER>_PASSWORD_COMMAND` and `_API_KEY_COMMAND` resolve credentials from helper stdout, with the same server-specific/shared precedence as `_FILE`.

Commands are parsed into argv and executed without a shell, using an absolute executable or a bare name resolved on the trusted PATH. Helpers run from a neutral working directory with stdin closed, stderr discarded, a 64 KiB stdout cap, and a shared 30-second execution/read deadline per credential. Ambiguous sources, invalid output, nonzero exits, and timeouts refuse startup with exit code 6.

POSIX cleanup targets the original process group even after its leader exits or is reaped, escalates for TERM-resistant descendants, and checks group membership before reporting termination. The threaded reader accumulates delayed writes until EOF or the size cap. Detached agents remain outside the cleanup boundary.

The existing startup boundary is preserved: project `.env` and working-directory YAML are not discovered automatically. Plugin configuration exposes the provider, and `docs/credential-providers.md` documents its contract and platform evidence.

## Release coordination

Network, Protect, and Access require `unifi-mcp-shared>=0.6.19,<0.7`. Published Shared 0.6.18 does not resolve `_COMMAND`; the feature must ship in Shared 0.6.19 before downstream app releases. No package version is manually set in source.

## Validation

- Maintainer follow-up is rebased onto current main and preserves Tyler's original commits.
- Two new exited-leader regressions failed before the fix and passed afterward, covering both unreaped and already-reaped leaders.
- The 30-test command-provider suite passes on macOS, including the threaded reader exercised with real pipes.
- Fresh installed Shared candidate wheel, outside the checkout: delayed output resolves; helper failure refuses startup without leaking output; a same-group descendant no longer executes after timeout refusal.
- All three built app wheels declare the Shared 0.6.19 floor.
- `make pre-commit` passes: Core 2,500; Shared 529; repository contracts 1,378 (6 skipped); relay 143; Network 1,294; Protect 530; Access 235; API 1,497. Format/lint, generated artifacts, docs checks, stdio protocol smoke, Worker tests, and Worker typecheck also pass.
- `uv run scripts/check_pin_alignment.py` passes, including downstream fresh installs/upgrades, security floors, runtime imports, and Network/API contracts against published Core floors.
- Final maintainer follow-up commit: `57dd79e1`.
- All 14 GitHub checks pass on `57dd79e1ee8d7a1de378acced4787cc9f64c43fd`.

Platform limits remain explicit: native Windows execution/cleanup and Keychain-backed controller authentication have not been verified. The original contributor's Linux GnuPG/controller acceptance was against `1c48b405962e1213c0899c6c9702f115a2fb6f7d`; it is historical evidence, not a live acceptance run against this final revision. The maintainer checks use synthetic credentials and no controller mutations.

## Checklist

- [x] Follows the existing shared bootstrap and app configuration patterns.
- [x] Preserves startup configuration security and secret-output isolation.
- [x] Adds regressions for the remaining process-lifecycle defect.
- [x] Updates platform documentation and downstream dependency floors.
- [x] Preserves contributor attribution.


tmowbrey:
`468be044` changes what this PR should assert, not just where it applies. I have the rebase resolved and green locally, but the resolution makes a semantic choice that is yours, so I am asking before I push it.

## The collision

This PR built a **refuse** boundary: a `.env`-supplied `_FILE` or `_COMMAND` is rejected with exit 6 before the file is opened or the helper is run. `468be044` makes that unreachable — the server no longer loads a project `.env` at all, so nothing arrives to refuse. You already re-expressed the `_FILE` probe accordingly, from exit 6 to exit 0 plus an audit hook proving the path is never opened.

The `_COMMAND` probe here still expects exit 6. I checked rather than assumed: restoring that assertion on top of current `main` fails, because the server starts normally with empty credentials.

```
E       AssertionError: ... Using bundled default configuration: .../config.yaml
packages/unifi-mcp-shared/src/unifi_mcp_shared/testing.py:125: AssertionError
FAILED apps/network/tests/unit/test_bootstrap_env_snapshot.py::test_dotenv_cannot_make_the_server_run_a_credential_helper
```

**And my "Considered & declined" note above is now wrong.** It declined dropping untrusted `.env` spellings before the ambiguity count, on the grounds that it would turn `_FILE` from refuse into ignore and change a contract you had verified. You have since made that change and gone further. I would rather flag that than quietly delete it — if you push this branch again, that entry should go.

## What I resolved, and the one choice I made for you

Both probes now assert exit 0, and the `_COMMAND` probe keeps the assertion that actually carries the weight: `not marker.exists()`, that the helper never *ran*. The helper still writes a marker before printing, so this proves absence of execution rather than absence of output. That assertion survives the semantic change untouched, which is the reason I think exit-0 is the right resolution rather than a weakening.

I also rewrote the justifications that cite machinery which no longer exists:

- `_trusted_environ`'s docstring explained itself in terms of what `load_dotenv` can introduce. The filter still has a job — bounding what the helper inherits to names present at startup — so I restated it that way rather than removing it.
- The `shutil.which(path=...)` comment called `os.environ["PATH"]` "the untrusted post-`.env` value".
- In `docs/credential-providers.md`, the whole **What the boundary does not cover** section existed to disclose the working-directory host redirect. That is the hole `468be044` closes, so the section now describes a limitation that is gone. I replaced it with the one that remains: the launcher is the trust boundary, and an explicitly loaded `--env-file` puts its values inside it.

The feature itself is unaffected. Neutral working directory, absolute-or-trusted-`PATH` resolution, no shell, discarded stderr, process-group termination, and the size cap enforced during the read never depended on `.env` loading — the working directory is still the project the client opened, and still on `sys.path`.

Suites on the rebase, run separately: Shared 523, Core 2389, Network 1215, Protect 521, Access 231, all passing. `ruff check` and `ruff format --check` clean, `make check-generated` reports no drift.

## Two questions before I push

1. **Is exit-0 the resolution you want for `_COMMAND`,** matching what you did for `_FILE`? The alternative is keeping an explicit refusal for a `_COMMAND` that appears from anywhere outside the startup snapshot, which would preserve this PR's original contract but only fires if some future code path injects into `os.environ` after startup.

2. **Does the `_TRUSTED_VARS` gate still earn its place?** With no `.env` loading, nothing mutates the environment between the snapshot and resolution, so the gate cannot fire in normal operation — it is now defence in depth against a path that does not exist yet. You kept it for `_FILE`, so I have kept it for `_COMMAND` and changed nothing, but it is worth a deliberate answer rather than inheritance.

Two stale strings on `main` itself, not touched here because they are not mine to change inside this PR: `_origin()` still returns `"(from a .env file)"`, and `resolve_env` still reports `"%s was supplied by a .env file, not by the environment the server was started with"`. Neither can be reached the way they read now.

Tell me which way on (1) and (2) and I will push the rebase.

packages/unifi-mcp-shared/src/unifi_mcp_shared/bootstrap.py
```diff
@@ -121,6 +138,352 @@ def _read_secret_file(var: str, path_str: str, logger: logging.Logger) -> str:
...
+
+    A reader thread is the only portable option, so it reads at most one byte
+    past the cap and its pipe is never closed from this thread -- closing under
+    a blocked reader is what makes the buffer lock unreleasable.
+    """
+    captured: list[bytes] = []
+
+    def _drain() -> None:
+        assert proc.stdout is not None
+        captured.append(proc.stdout.read(_SECRET_MAX_BYTES + 1))
```
sirkirby: [P2] Read until EOF, the size cap, or the shared deadline.

Popen uses bufsize=0, so stdout is a raw stream: one read(n) can return the currently available chunk without reaching EOF. A real helper that wrote `first`, paused, then wrote `second` produced `(b"first", "eof")` with exit status 0; `second` remained unread. On the Windows branch this accepts a truncated credential and can miss subsequent invalid or oversized output. Please accumulate chunks while enforcing the existing cap/deadline and add a delayed multi-write regression. This was reproduced by exercising the threaded branch with real pipes on macOS, not on native Windows.

packages/unifi-mcp-shared/src/unifi_mcp_shared/bootstrap.py
```diff
@@ -121,6 +138,352 @@ def _read_secret_file(var: str, path_str: str, logger: logging.Logger) -> str:
...
+        try:
+            os.killpg(group, sig)
+        except ProcessLookupError:
+            return _reap(proc, _SECRET_COMMAND_REAP_S)
+        except OSError:
+            # EPERM: a helper that changed uid (sudo, doas, pkexec). Nothing to
+            # signal, but startup must still refuse with the credential exit
+            # code rather than a traceback.
+            return _reap(proc, _SECRET_COMMAND_GRACE_S)
+        if _reap(proc, grace):
```
sirkirby: [P2] Do not equate reaping the direct helper with terminating its process group.

This early return skips SIGKILL when the direct helper exits after SIGTERM, even if a descendant in the same process group ignores SIGTERM. I reproduced `_terminate_process_tree` returning True while that descendant survived and wrote a marker afterward. This is not the documented exception for a descendant that deliberately leaves the session. Please ensure remaining group members are terminated after the grace period and avoid reporting whole-tree termination solely from the direct child's exit. Cover a TERM-resistant descendant and the case where the group leader has already exited.

packages/unifi-mcp-shared/tests/test_bootstrap_command_provider.py
```diff
@@ -0,0 +1,459 @@
...
+
+def test_a_helper_that_leaves_a_detached_descendant_still_refuses(helper, tmp_path):
+    """A descendant outside the helper's session must not hang startup.
+
+    Out of process on purpose: this failure mode was an unbounded block inside
+    ``BufferedReader.close()``, which no in-process timeout can interrupt. The
+    helper here SUCCEEDS and exits 0 -- only its detached child holds stdout --
+    which is the shape docs/credential-providers.md explicitly contemplates.
+    """
+    script = helper("setsid sleep 30 &\nprintf %s the-password", name="detached.sh")
```
sirkirby: [P2] Construct the detached helper without depending on an external setsid executable.

This test fails on the maintainer Mac: `setsid` is absent, no detached process starts, and the helper exits successfully, giving status 0 instead of the expected refusal status 6. The focused suite therefore finishes 24 passed / 1 failed. Please use Python's POSIX session support to create the intended process arrangement, and explicitly clean up the detached test child afterward.

sirkirby (changes requested):
Tyler, I reviewed `1c48b405962e1213c0899c6c9702f115a2fb6f7d` and am requesting changes for the reader and process-cleanup defects below, plus the macOS test failure.

Independent validation on macOS: the focused command-provider suite finished 24 passed / 1 failed. Two additional synthetic-helper probes reproduced truncated output through the threaded reader and a same-process-group descendant surviving termination. The threaded-reader probe used real pipes on macOS; this is not native Windows acceptance evidence. No real credentials or controller mutations were involved.

Please also rebase onto current main while preserving the published startup security fix. My isolated trial rebase hit seven conflicts, including bootstrap.py and testing.py, and I aborted it without changing your branch. The provider documentation and test helpers still describe automatic working-directory .env loading, which current main deliberately removed.

The publication note needs correction too: unchanged runtime imports do not guarantee that the new feature is available. I verified that published Shared 0.6.17 does not resolve a command-only password. Apps/plugins exposing _COMMAND need a dependency floor that guarantees the provider implementation, coordinated with its Shared release. The existing security floor must remain intact.

After these corrections, please rerun make pre-commit and the provider acceptance checks on the rebased head. Full current-main integration and native Windows acceptance remain unverified.

tmowbrey:
Pushed `57530959` on a rebase onto current `main`. All three defects fixed, each reproduced as a failing test before the fix.

## The reader

You were right about the mechanism. `Popen` gets `bufsize=0`, so stdout is an `io.FileIO` and one `read()` is one syscall returning only what has already arrived — the POSIX selector path loops, the threaded path did not. It now reads until EOF or the cap.

One consequence worth stating rather than leaving for someone to find: a helper that exits 0 while a detached process still holds stdout now reaches the deadline instead of returning early with the first burst. That makes the threaded path agree with the POSIX one and with what `docs/credential-providers.md` already promises, but it is a behaviour change on the platform the docs mark unverified, so it should not go into a release silently.

## The process group

Reaping the direct child said nothing about the group. It now reaps the child, asks the kernel whether the group still holds members, and only stops escalating once nothing is left.

**I had left the same mistake on the `EPERM` branch.** An adversarial pass over my own fix found `return reaped or _reap(...)` still reporting a terminated tree from the child's exit alone — at the exact point where a survivor had just been proven and we had lost the ability to signal it. That is the case that matters most, since a uid-changing descendant is why `EPERM` happens at all. Fixed, with its own regression, and I mutation-tested that regression against the pre-correction code to confirm it pins the defect rather than something incidental.

Two limits are now stated in the docstring instead of being papered over:

- An unreaped zombie still answers `killpg`, so where the server is PID 1 with no init reaping orphans — a container without `init: true`, which is what `docker/docker-compose.yml` currently gives — a fully dead tree reads as surviving. A false negative on a path that refuses startup either way, so I did not reach for PID-1 zombie reaping, which would mean `waitpid(-1)` and could swallow someone else's child.
- The group id is the reaped child's pid, so a full pid wrap inside the grace period could aim the second signal at an unrelated group. At default `pid_max` that needs millions of forks in two seconds.

Worst-case latency on the escalation path goes from about 4s to about 6s, inside the 30s budget.

## The macOS failure

`setsid(1)` is util-linux; macOS has no such binary. The descendant never started, the helper succeeded, and the test asserted a refusal that could not happen — which is your 24 passed / 1 failed. It uses `os.setsid` now, asserts via a pidfile that the descendant really detached, and kills it in a `finally`.

Both process tests also had a weaker arrangement than they claimed: a shell's `trap "" TERM` does not protect the `sleep` it forks, so `killpg` reaped that and the shell exited on its own. The descendant now ignores SIGTERM in its own handler, and the test waits for the handler to be installed before signalling — without that gate it was racing the interpreter's startup.

## The floor — my note was wrong

You checked it and I did not. I argued no floor was needed because no `apps/*/src` path is touched, so the runtime import surface from Shared is unchanged. That was the wrong thing to measure. What matters is whether the app's advertised configuration depends on behaviour that only exists in an unpublished release, and all three plugins pass `UNIFI_<SERVER>_PASSWORD_COMMAND` through in both `.mcp.json` and `.claude-plugin/plugin.json`. Against installed Shared 0.6.17 the variable is silently ignored and startup then fails naming a missing credential rather than the version — silent, which is worse than an `ImportError`.

Network, Protect and Access now declare `unifi-mcp-shared>=0.6.18,<0.7`, preserving the 0.6.17 startup-trust floor. **0.6.18 is my assumption for the next Shared release — change it to whichever release actually carries the provider.** `uv lock --check` is clean because `[tool.uv.sources]` resolves Shared from the workspace, so the floor stages before the release exists. `apps/api` takes Shared as a workspace dependency with no constraint; `packages/unifi-mcp-relay` stays at 0.6.17 since it imports nothing from `bootstrap`.

## The rebase, and one thing in the description that had gone stale

Seven conflicts, same ones you hit. The substantive half is that this PR and `468be044` assert opposite outcomes for the same scenario: this branch *refused* a `.env`-supplied indirection with exit 6, and `main` never loads the `.env` at all so there is nothing to refuse. Both probes now assert exit 0 and keep the assertion that carries the weight — `not marker.exists()`, that the helper never *ran*. That survives the change intact and is the stronger claim.

I also struck the "Considered & declined" entry that rejected exactly this change. It declined turning `_FILE` from refuse into ignore on the grounds that it altered a contract you had verified; you then made that change and went further. Marked obsolete in the description rather than quietly deleted.

`docs/credential-providers.md` needed the same treatment. Its "What the boundary does not cover" section existed to disclose the working-directory host redirect — the hole `468be044` closes — so it now describes the limit that remains, which is that the launcher is the trust boundary and an explicitly loaded `--env-file` puts its values inside it. The `_trusted_environ` docstring and the `shutil.which` comment both justified themselves in terms of `load_dotenv` behaviour that no longer exists; reworded.

Two stale strings on `main` itself I did not touch, since they are not mine to change inside this PR: `_origin()` still returns `"(from a .env file)"`, and `resolve_env` still reports `"%s was supplied by a .env file, not by the environment the server was started with"`. Neither is reachable the way it reads now.

## Gates

Suites run separately: Shared 526, Core 2393, Network 1267, Protect 525, Access 235, API 1473. `ruff check` and `ruff format --check` clean, `uv lock --check` clean, `make check-generated` no drift.

Still unverified and I am not claiming otherwise: native Windows. The threaded reader is exercised by real pipes on Linux through the `os_name="nt"` parameter, which is why the platform is a parameter rather than a global, but that is not Windows acceptance.

sirkirby (approved):
Tyler, I re-reviewed the final head and completed the remaining maintainer follow-up on your branch, preserving your commits and attribution. The delayed threaded reads, TERM-resistant descendants, already-exited/reaped group leaders, and portable detached-process fixture are now covered and passing.

I ran the full make pre-commit gate, pin alignment with fresh installs, and a separate installed-wheel synthetic-provider probe. All passed, and all 14 GitHub checks are green on 57dd79e1ee8d7a1de378acced4787cc9f64c43fd, rebased onto current main. I have no remaining blocking findings.

The Shared 0.6.19 floor now matches the intended feature release; publish Shared before downstream apps. Native Windows and Keychain-backed controller acceptance remain explicitly unverified in the contract. Those documented platform limitations are unchanged. Approved.

PR #671, opened 2026-09-07
fix(core): restrict diagnostics to operation metadata

tmowbrey:
Enabling diagnostics logged whole controller payloads at INFO. Secret-key redaction and address regexes cannot reliably remove client names, plain hostnames, arbitrary controller strings, or every address representation.

Diagnostics now log selected operation metadata only: tool name or HTTP method, duration, boolean outcome when available, and exception class. Arguments, results, request paths, controller bodies, and exception messages are omitted. The old argument/result flags remain accepted for configuration compatibility but cannot enable payload logging; all three apps' configuration guides document this. Controller responses returned to callers are unchanged.

No new general-purpose address masker or dependency-logger filter is included. Those would not provide the closed diagnostic boundary this fix requires. Serialization failures also no longer fall back to the original object's string representation.

Validation: a focused probe reproduced client-name, plain-hostname, and unbracketed-IPv6 leakage in the initial masking approach. Twelve Core regressions now verify the complete emitted metadata and exclusion of supplied values even with payload flags enabled, exception handling, disabled diagnostics, and payloads whose string conversion must never be called. Sixteen Shared diagnostics compatibility tests passed in a separate process. Full `make pre-commit` and all 14 GitHub checks passed on ab83bfcc7b89b36a035c5aac4c08f2e9c6f0ad64. The installed-wheel live check preserved the client response while emitting metadata only. String-valued disabled configuration is also parsed correctly; previously `"false"` enabled diagnostics.

Publish Core first and align downstream floors before claiming this behavior for released apps. This change covers diagnostics; it does not claim that unrelated dependency loggers have been comprehensively scrubbed.


sirkirby (approved):
Tyler, I reproduced address/name leakage that the initial regex approach could not cover, so I tightened this to a closed metadata boundary: diagnostics omit arguments, results, request paths, controller bodies, and exception messages. I removed the unused general-purpose logger filter and updated the three app guides to explain the legacy payload flags. I also reproduced and fixed string-valued "false" enabling diagnostics.

Twelve Core regressions and sixteen Shared compatibility tests passed. The rebased Network suite passed 1,196 tests, including the firewall privacy assertions that exposed the diagnostics leak during #647 integration. Independently installed wheels completed a real client read with the MAC retained in the response and only allowed metadata in the diagnostic record.

The full `make pre-commit` passed on ab83bfcc7b89b36a035c5aac4c08f2e9c6f0ad64 and all 14 current GitHub checks are green. I am merging this ahead of the remaining lifecycle integration and Core release.

Issue #678, opened 2026-09-07
Heal-time connection close is not cancellation-safe


tmowbrey:
Part of #676.

## Defect

`ManagerFactory._heal_blocked_connections` (`apps/api/src/unifi_api/services/managers.py`) pops the
blocked connections from the cache, awaits `_stop_domain_managers(dropped)`, and only then closes
them in a plain `for cm in removed:` loop.

Both awaits are cancellation points, and the connections are already out of the cache. A cancellation
between the stop and the close — again, lifespan shutdown is the realistic trigger — leaves an
authenticated controller session that nothing owns and nothing will ever close. `invalidate_controller`
has the same shape.

## Fix

Close the removed connections in a `finally`, so the close runs whether the stop completed, raised or
was cancelled. The close itself needs shielding if it is to survive the cancellation that triggered
it, since an already-cancelled task cannot ordinarily await.

Worth deciding deliberately whether the shield is right here or whether a leaked session at hard
shutdown is acceptable — the process is ending either way, and an unbounded shield can delay exit.
I lean toward shielding only the close and keeping it bounded.

## Evidence

Code-level, network-free regression: cancel between stop and close, assert the connection was closed
and the cancellation still propagated.

## Scope

The heal path, and `invalidate_controller` if the same treatment applies cleanly.

PR #710, opened 2026-09-13
fix(api): stop every dropped manager before a cancellation propagates

### LINKED ISSUE #677: Cancellation skips the remaining listeners in _stop_domain_managers
Part of #676.

## Defect

`ManagerFactory._stop_domain_managers` (`apps/api/src/unifi_api/services/managers.py`) stops each
dropped manager inside `try: ... except Exception`. `asyncio.CancelledError` inherits from
`BaseException`, not `Exception`, so it is not caught: a cancellation delivered while awaiting one
manager's `stop_listening()` propagates out of the `for` loop and **every listener after it in the
list is never stopped**.

The managers are already popped from the cache by the caller at that point, so nothing else will
reach them. Each one left running keeps its own reconnect task alive and re-logs-in on its own timer
— with the credentials that were being rotated away, which is the case the stop exists to prevent.

The realistic trigger is lifespan shutdown, where cancellation is normal rather than exceptional.

## Fix

Let cancellation propagate, but not before the remaining managers have been stopped — stop them all,
then re-raise. Note that `stop_listening()` itself deliberately re-raises a cancellation aimed at its
caller, so the fix has to distinguish "this task is being cancelled" from "this manager's stop was
cancelled" rather than swallowing either.

## Evidence

Code-level and provable network-free: a regression that cancels during the first manager's stop and
asserts the second was still stopped, and that `CancelledError` still reaches the caller. No
controller or hardware needed.

## Scope

`_stop_domain_managers` only. The related close-path issue is filed separately.

wangzhengzhuo05:
## Summary

`ManagerFactory._stop_domain_managers` stopped each dropped manager inside
`try: ... except Exception`. `asyncio.CancelledError` is a `BaseException`, so a cancellation
delivered while awaiting one manager's `stop_listening()` propagated straight out of the `for`
loop and every manager after it in the list was never stopped. By then the managers have already
been popped from the cache by the caller, so nothing else reaches them: each one left running
keeps its own reconnect task and re-logs-in on its own timer with the credentials the stop exists
to retire. Lifespan shutdown is the realistic trigger, where cancellation is normal.

The loop now stops every manager and only then re-raises. To decide whether a `CancelledError` is
aimed at this task or came from the manager's own `stop_listening()`, it checks
`asyncio.current_task().cancelling()` — that counter stays non-zero for the rest of the task once
our own cancellation has been delivered, so it is a reliable "this task is being cancelled"
signal; a `CancelledError` arriving while it is zero can only have come from the manager's stop and
is re-raised immediately rather than swallowed.

The per-manager body moved unchanged into `_stop_domain_manager`, including the existing
`except Exception` around `stop_listening()` that keeps one broken manager from blocking the rest.
This is issue #677 only; the close-path issue #678 is deliberately not touched.

Fixes #677

## Type of change

- [x] `fix:` bug fix
- [ ] `feat:` new feature

## Scope

- [ ] Network MCP server
- [ ] Protect MCP server
- [ ] Access MCP server
- [ ] Relay sidecar
- [ ] Cloudflare Worker / CLI
- [x] API server
- [ ] Shared packages
- [ ] Plugin packaging
- [ ] Documentation

## Checklist

- [ ] I ran `make pre-commit`, or I explained below why it was not run.

`make pre-commit` was not run in full; the focused suite for the changed app was run instead:
`uv run --package unifi-api-server pytest apps/api/tests/test_managers_factory.py -v` — 28 passed.

## Tests

Two tests were added to `apps/api/tests/test_managers_factory.py`:

- `test_cancellation_during_one_stop_still_stops_the_rest` — the regression from the issue. The
  first manager's `stop_listening()` blocks, the task is cancelled while it is in flight, and the
  test asserts both that the second manager was still stopped and that `CancelledError` still
  reached the caller (`pytest.raises` plus `task.cancelled()`). No controller or network needed.
- `test_stop_domain_managers_stops_both_in_order_without_cancellation` — the control, so the
  regression cannot pass because stopping became a no-op.

**Mutation check.** Replacing the cancellation handling with a plain
`await self._stop_domain_manager(manager)` loop (the old shape) turns the regression red with
`AssertionError: assert 'second' in []` while the control stays green — the failure is attributable
to the second manager not being stopped, exactly the defect in the issue. Restoring the fix is
green again.


sirkirby:
Thanks for the focused fix and cancellation regression. #677 was closed by #714, which also addressed #678 and uses a shielded cleanup task so cancellation cannot cancel the in-flight first stop, then closes every removed connection from `finally`. This branch now conflicts with main and is fully superseded, so I am closing it. Your analysis of the lost-listener failure was correct and helped validate the merged behavior.

PR #711, opened 2026-09-13
feat(network): add guarded Internet traffic routes

keithah:
## Summary

Adds a narrowly scoped `unifi_create_traffic_route` workflow for safe Internet traffic routes and hardens every shared mutation path around it.

- Allow only one explicit `CLIENT` target with a valid unicast MAC address and a freshly verified WAN target.
- Enforce the same guard in `TrafficRouteManager`, MCP tools, legacy `FirewallManager` compatibility methods, and REST update/toggle dispatch; generic REST creation remains deliberately unavailable.
- Reject empty target lists instead of widening scope, require a route ID before reporting create success, and return confirmed creates under the standard `data` envelope.
- Invalidate both guarded and legacy route caches after every uncertain create, update, toggle, kill-switch update, or delete request outcome.
- Reuse the cached route manager for legacy compatibility paths; preserve privacy-safe logging and regenerated metadata/docs.

## Type of change

- [x] `fix:` bug fix
- [x] `feat:` new feature
- [x] `test:` adding or updating tests

## Scope

- [x] Network MCP server
- [ ] Protect MCP server
- [ ] Access MCP server
- [ ] Relay sidecar
- [ ] Cloudflare Worker / CLI
- [x] API server
- [x] Shared packages
- [x] Plugin packaging
- [x] Documentation

## Checklist

- [x] I ran focused checks for the changed paths.
- [x] I ran `make pre-commit`.
- [x] I updated generated manifests and checked drift.
- [x] I added or updated regression tests for behavior changes.
- [x] I updated user-visible documentation.

## Testing

- Traffic-route manager suite: 30 passed.
- Network MCP and legacy firewall suites: 115 passed.
- Ruff check and format check passed for the changed source and tests.
- `make pre-commit` passed on `71ee594c…`: 1,563 tests; generated skill-reference, API action-catalog, and support-skill drift checks; Worker typecheck.
- Earlier local acceptance applies to the prior published source revision only. This cache-coherency follow-up has not been redeployed, and no live traffic route was created or changed during this update.

## Notes for reviewers

The Internet-route allowlist is intentionally narrow. Broad client selectors, malformed or multicast MACs, non-WAN targets, stale target state, unsafe enablement, and unsafe retargeting are rejected. Safety-critical decisions use fresh controller reads rather than manager caches.

The manager is the canonical policy boundary: MCP tools translate its validation result rather than reimplementing it, while legacy compatibility and REST mutation paths cannot bypass the guard.


sirkirby (changes requested):
Keith, thanks for following through on the review feedback. I reviewed ed361ef58d5a6bd3559d30b307fbcdb8f7ef4f38 and confirmed that the Copilot findings are addressed. The guarded route workflow fits the project, and I want to merge this once the dependency issue below is resolved.

## Required before merge

- [ ] Coordinate the Core release and update the Core dependency floors in both `apps/network/pyproject.toml` and `apps/api/pyproject.toml` to the version containing these manager constructor changes. Both currently allow `unifi-core>=0.4.55,<0.5`, but that published version does not support the new arguments. Publish the supporting Core changes first, then update the downstream floors and rerun the installed-package check. A floor pointing at an unpublished release will not resolve the installation problem.

The [pin-alignment job](https://github.com/sirkirby/unifi-mcp/actions/runs/35177390852) confirms that Network fails at startup with `TrafficRouteManager.__init__() takes 2 positional arguments but 3 were given`. I also tested the API's firewall manager builder against the published Core 0.4.55 wheel and reproduced `FirewallManager.__init__() got an unexpected keyword argument 'traffic_route_manager'`. The API import check passes because this manager is constructed lazily.

## Validation

I ran the affected suites: 626 tests and 196 subtests passed, and Ruff passed. Six live controller reads passed. Five create-preview checks passed, covering valid WAN targeting and rejection of non-WAN, broad, multicast, and empty targets. Those previews used a synthetic unicast client MAC; I made no controller mutations.

Before final approval, I still need independent live create/update/read-back/delete verification on the final revision. The current read and preview results do not establish mutation acceptance.

keithah:
Core remediation is prepared in #733 (bfc6ffd3565b2aba77de2cbde125c576dc5811aa). Steward passed the exact head; 2,719 unifi-core tests plus Ruff format/lint pass. The downstream floor bump is intentionally waiting for Core v0.4.56 to be merged/published because PyPI currently only has unifi-core 0.4.55 and pin-alignment correctly rejects the new manager APIs against that wheel.

keithah:
## Dependency and release order

PR #711 depends on this Core change in PR #733. The dependency is runtime/API compatibility, not merely ordering: Network and API now construct Core managers with the new TrafficRoute manager contracts and rely on the strict write-boundary behavior added here. The currently published `unifi-core==0.4.55` wheel does not contain those constructor signatures or the new behavior, which is why the pin-alignment job fails when it installs the published package.

Required sequence:

1. Merge PR #733 after its CI/API symmetry issue is fixed.
2. Publish the Core package containing this merged code as `unifi-core 0.4.56`.
3. Update PR #711's Network and API dependency floors to `unifi-core>=0.4.56,<0.5` and refresh its lockfile if required.
4. Rerun the installed-package/pin-alignment check and the affected Network/API suites.
5. Re-review and merge PR #711 only after the published `0.4.56` wheel is available and the downstream checks pass.

The floor must not be raised before `0.4.56` is published: that would make dependency resolution fail rather than solve the mismatch. Conversely, merging #711 against the old floor allows installation with `0.4.55`, which lacks the APIs it calls.

PR #715, opened 2026-09-14
fix(network): validate event buffer configuration

### LINKED ISSUE #681: Event buffer size and TTL are unguarded
Part of #676.

## Defect

`EventManager` builds its buffer straight from config with no validation
(`packages/unifi-core/src/unifi_core/network/managers/event_manager.py`):

```python
max_size=int(cfg.get("buffer_size", 100)),
ttl_seconds=int(cfg.get("buffer_ttl_seconds", 300)),
```

Both values come from the environment through `config.yaml`:

```yaml
buffer_size: ${oc.env:UNIFI_NETWORK_EVENT_BUFFER_SIZE,100}
buffer_ttl_seconds: ${oc.env:UNIFI_NETWORK_EVENT_BUFFER_TTL,300}
```

Two failure shapes, both from an operator typo rather than anything exotic:

- A non-numeric value raises `ValueError` out of `int()` during manager construction. That happens at
  startup, so the server fails to come up, and the traceback points at the event manager rather than
  at the variable the operator got wrong.
- `0` or a negative value is accepted. A zero-capacity buffer silently discards every event, and
  `unifi_recent_events` then reports exactly what a healthy-but-quiet site reports — `listening=true`,
  `attached=true`, `buffer_size=0` — with no indication that the configuration is the reason. That is
  the "empty buffer that reads as a valid result" failure the listener work set out to remove, just
  reintroduced through configuration.

## Fix

Validate both at construction: reject or clamp non-positive values, and fail with a message naming
the variable rather than letting `int()` raise. Whether a bad value should refuse startup or fall
back to the default with a warning is a judgement call — the rest of the bootstrap refuses, so
refusing is probably more consistent, but the listener is optional in a way credentials are not.
Happy to follow whichever you prefer.

## Evidence

Entirely code-level and network-free. No controller involvement.

sirkirby:
A malformed `UNIFI_NETWORK_EVENT_BUFFER_SIZE` or `UNIFI_NETWORK_EVENT_BUFFER_TTL` currently either crashes with an unhelpful `int()` traceback or accepts a nonpositive value that makes the event listener look healthy while silently retaining no events.

This change validates both settings as positive integers when `EventManager` is constructed. Invalid values now refuse startup with an error that names the exact environment variable. The positive-integer constraint is documented in the example environment, Network README, and configuration reference.

Closes #681

Validation:
- `make pre-commit`
- Focused Core event-buffer/streaming suites: 31 passed
- Focused Network runtime/event-manager suites: 73 passed
- Full Core suite: 2,598 passed
- Full Network suite: 1,315 passed
- Full API suite: 1,556 passed
- Invalid cases cover zero, negative integers, numeric and nonnumeric strings, booleans, floats, and null values for both settings.
- Runtime tests prove actual `UNIFI_NETWORK_EVENT_BUFFER_*` overrides produce variable-specific startup errors.

No live controller traffic is required: validation completes before the event manager opens a controller connection. Exact-image startup refusal checks are included in the review evidence.

Exact-image startup checks:
- Network image: `sha256:cde24b98ce105aa013d9a9c1d0d7d8cd52eed11528a952c557cc5d881cada6f6`
- `UNIFI_NETWORK_EVENT_BUFFER_SIZE=0` refused startup and named that variable.
- `UNIFI_NETWORK_EVENT_BUFFER_TTL=invalid` refused startup and named that variable.
- Positive values (`5`, `60`) constructed the runtime with buffer capacity 5.


sirkirby:
Exact-image configuration smoke passed on `sha256:cde24b98ce105aa013d9a9c1d0d7d8cd52eed11528a952c557cc5d881cada6f6`: size `0` and TTL `invalid` each refused runtime startup with the exact environment variable named; positive values `5` and `60` constructed the runtime with capacity 5. No controller connection was opened.

PR #733, opened 2026-09-19
fix(core): harden traffic route write boundary

keithah:
## Summary

- validate and normalize TrafficRoute writes in Core before controller POST/PUT
- block unsafe INTERNET catch-all creates and preserve supported partial updates
- require exactly one canonical created-route response with a nonblank `_id`
- invalidate route caches after every attempted create, including uncertain outcomes
- add cache-generation protection and focused write-boundary regressions

## Verification

- `uv run --python 3.13 --package unifi-core pytest packages/unifi-core/tests -q` — 2719 passed
- `uvx --from ruff ruff format --check packages/unifi-core/src packages/unifi-core/tests` — passed
- `uvx --from ruff ruff check packages/unifi-core/src packages/unifi-core/tests` — passed
- Steward exact-head review passed at `bfc6ffd3565b2aba77de2cbde125c576dc5811aa`

This release is required before downstream Network/API dependency floors can be raised for PR #711.


keithah:
## Downstream dependency: PR #711

PR #711 depends on this Core PR at runtime. It uses the new Core TrafficRoute manager constructor contracts and write-boundary behavior; the published `unifi-core==0.4.55` wheel does not provide those APIs. That is why #711's pin-alignment job fails when it tests against the published package, even though local monorepo tests pass.

Required release sequence:

1. Fix the current API cross-layer symmetry failure in this PR: the Core model now marks `domains`, `ip_addresses`, `ip_ranges`, `regions`, and `target_devices` as mutable, but the API `TrafficRoute` GraphQL type does not expose them. Either add the fields to the API contract and regenerate its artifacts, or keep the Core/API mutability contract aligned without weakening the symmetry gate.
2. Rerun the full affected checks and Steward against the new exact head.
3. Merge this Core PR and publish `unifi-core 0.4.56`.
4. Update #711's Network and API dependency floors to `unifi-core>=0.4.56,<0.5`, refresh lock data if needed, and rerun pin alignment plus affected tests.

The Core release must come first because changing #711's floor to an unpublished version would make installation unsatisfiable; leaving the old floor permits installation of `0.4.55`, which lacks the manager APIs #711 calls.

